PR #121, opened 2026-04-16
Add cloud service query for HTTP SQL execution (stacked on #120)

### LINKED ISSUE #115: rewrite cloud service query to use query endpoints
Query endpoints API can be used to send queries to services uses existing API key / JWT auth. 

This will be a much better approach for querying cloud services:
1. It does not need a local clickhouse client, so we can remove all of that. It just uses HTTP
2. We dont need anything special for authing the user to a service, so that can all go too

sdairs:
## Summary

Closes #115. Stacked on #120 (`issue-48-debug-auth-source`) — rebase/merge that first.

New `clickhousectl cloud service query` subcommand that executes SQL against a cloud service over HTTP via `queries.clickhouse.cloud`, reusing whatever credentials are already configured. No local `clickhouse` binary and no service password required.

- `--query` / `--queries-file` / stdin for SQL input
- `--database` selects the target database
- `--format` overrides output format (defaults to `PrettyCompact` on a TTY, `TabSeparated` when piped)
- Response body is streamed straight to stdout

OAuth JWT works here the same as an API key (matches the sample curl in the issue), so the command is classified read-only for the OAuth write-gate — the SQL's own semantics are enforced on the ClickHouse side.

### Library change

Adds `Client::authenticated_request` in `clickhouse-cloud-api` so CloudClient can reuse the same auth/http for adjacent hosts. No generated-code changes.

### Not in this PR

Leaving `cloud service client` in place for now — issue 115 calls this out as a "much better approach" that eventually replaces it, but removal is a follow-up.

## Test plan

- [x] `cargo test --workspace` passes (209 tests, +2 new for `RunQueryRequest` serialization)
- [x] `cargo build` clean, `cargo clippy` clean on changed crates
- [x] `cloud service query --help` renders and matches the `feedback_agent_context_help` memory (CONTEXT FOR AGENTS adds non-obvious context only)
- [ ] Manual: `cloud service query --name <svc> --query "SELECT 1"` hits queries.clickhouse.cloud with OAuth JWT
- [ ] Manual: same command with API key auth works
- [ ] Manual: piping SQL on stdin works; redirecting stdout picks TabSeparated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,6 +77,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.102"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+
+[[package]]
 name = "assert-json-diff"
 version = "2.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


sdairs:
Closing to restart the implementation cleanly \u2014 will reopen once rewritten.

PR #172, opened 2026-05-15
Cover query endpoint GET + DELETE in service suite

### LINKED ISSUE #158: Close service-suite query endpoint coverage — GET + DELETE
## Summary

The service integration test calls `instance_query_endpoint_upsert` and re-upserts, but never calls `instance_query_endpoint_get` or `instance_query_endpoint_delete`. The current test leaves the endpoint binding hanging after deleting the API key, so the GET/DELETE paths are unexercised.

## Scope

Extend the existing `Query API Endpoint` phase in `tests/integration_test.rs`:

- After the upsert + assertions, add a phase that calls `instance_query_endpoint_get` and asserts the response matches the upsert (roles, openApiKeys, allowedOrigins).
- Before the API key delete, add a phase that calls `instance_query_endpoint_delete`. Verify via `instance_query_endpoint_get` returning 404 (or the appropriate "no binding" response).

## Where this lands

- `crates/clickhouse-cloud-api/tests/integration_test.rs` — extend the `Query API Endpoint` phase.

## Notes

- Order matters: delete the endpoint binding **before** deleting the API key, so we can distinguish "binding cleanup works" from "API key was already gone."

Parent: #151

sdairs:
## Summary

- Adds an `instance_query_endpoint_get` step after the upsert that asserts the binding stored matches what we sent (id, roles, openApiKeys, allowedOrigins).
- Adds an `instance_query_endpoint_delete` step before the API key delete, followed by a GET that must return 4xx — so the test can distinguish binding cleanup working from the API key being gone.
- Extends `CleanupRegistry` with a query endpoint slot that is cleaned up before API keys on the failure path, preserving the same ordering invariant on panic/early-exit.

Closes #158.

## Test plan
- [x] `cargo test -p clickhouse-cloud-api` (live tests remain `#[ignore]`; non-live suite passes)
- [ ] Live `cloud_service_crud_lifecycle` run in CI exercises the new GET + DELETE steps end-to-end

### DIFF
--- a/crates/clickhouse-cloud-api/tests/common/support.rs
+++ b/crates/clickhouse-cloud-api/tests/common/support.rs
@@ -372,6 +372,10 @@ pub struct CleanupRegistry {
     /// being deleted, table drops are redundant but harmless.
     tables: Vec<String>,
     api_key_ids: Vec<String>,
+    // Query endpoint bindings reference an API key; they must be deleted
+    // before the key they point to, otherwise we can't distinguish "binding
+    // cleanup works" from "API key was already gone."
+    query_endpoint_service_ids: Vec<String>,
 }
 
 impl CleanupRegistry {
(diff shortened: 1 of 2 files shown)


iskakaushik (dismissed):
LGTM!

iskakaushik (approved):

PR #174, opened 2026-05-15
Cover instance_password_update + instance_scaling_update

### LINKED ISSUE #159: Cover instance_password_update + instance_scaling_update
## Summary

`instance_password_update` and `instance_scaling_update` (the vertical-only scaling endpoint, distinct from `instance_replica_scaling_update`) have no live coverage.

## Scope

Extend `tests/integration_test.rs` with two new phases:

### Password update
- `instance_password_update` rotates the service password. Assert the response surfaces a new password.
- The service-level password isn't used elsewhere in the test (the query flow uses an openapi key) — a successful response is the pass condition.

### Vertical scaling
- Add a phase to the existing `Scaling` section using `instance_scaling_update` instead of `instance_replica_scaling_update`. Round-trip memory the same way (e.g. 8 → 16 → 8 GB), polling for the change to land.
- Confirm post-condition equals pre-condition so the rest of the scaling phase observes the expected base state.

## Where this lands

Parent: #151

sdairs:
Closes #159.

## Summary

Adds two phases to the live service-CRUD integration test:

- **Vertical scaling (deprecated endpoint)** — exercises `instance_scaling_update` (PATCH `/scaling`), distinct from `instance_replica_scaling_update` (PATCH `/replicaScaling`) already covered. Captures `minTotalMemoryGb`/`maxTotalMemoryGb` from the service at the tail of the existing scaling phase (1 replica @ 8 GB, so totals map cleanly to per-replica), round-trips `8 → 16 → 8 GB`, polling for each change. Post-condition equals pre-condition so downstream phases see the expected base state. Non-blocking — the endpoint is deprecated and nothing downstream depends on it.
- **Password rotation** — new "Password" phase exercises `instance_password_update` with an empty body (server generates the password). Asserts the response surfaces a non-empty password. The query path used elsewhere in the suite is openapi-key based, so the rotated password is not consumed downstream; no re-rotation needed since the service is deleted immediately after. Non-blocking.

Also rewrites four pre-existing nested `if let` blocks in `spec_coverage_test.rs` as let-chains so `cargo clippy --tests -- -D warnings` passes — CI's clippy invocation omits `--tests`, so these had drifted past the lint floor. Semantics unchanged.

## Test plan

- [x] `cargo clippy -p clickhouse-cloud-api --tests -- -D warnings`
- [x] `cargo test -p clickhouse-cloud-api` (mock + spec-coverage tests pass; live integration test remains `#[ignore]` and is unchanged structurally)
- [ ] Live integration run against the test org (cloud-integration CI on PR)

Do NOT merge — verify the live run first.

### DIFF
--- a/crates/clickhouse-cloud-api/tests/integration_test.rs
+++ b/crates/clickhouse-cloud-api/tests/integration_test.rs
@@ -1213,7 +1213,142 @@ async fn cloud_service_crud_lifecycle() -> TestResult<()> {
             })
             .await?;
 
-        // ── 7. Delete ────────────────────────────────────────────────
+        // Vertical scaling round-trip via the deprecated
+        // `instance_scaling_update` endpoint (PATCH /scaling). This is
+        // distinct from `instance_replica_scaling_update` (PATCH
+        // /replicaScaling) exercised above: the deprecated endpoint takes
+        // `minTotalMemoryGb` / `maxTotalMemoryGb` and only the vertical
+        // axis. We enter this phase at 1 replica @ base_memory_gb so the
+        // total-memory body maps directly to per-replica memory; round-trip
+        // back to the entry state so subsequent phases see the expected
+        // base.
+        let pre_vertical = failures
+            .run(
+                &ctx,
+                StepKind::Blocking,
+                "capture pre-state for deprecated vertical scaling",
+                || {
+                    let client = client.clone();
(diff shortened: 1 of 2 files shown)


iskakaushik (dismissed):
LGTM!

iskakaushik (approved):

PR #175, opened 2026-05-15
Cover instance_private_endpoint_create + config_get

### LINKED ISSUE #160: Cover instance_private_endpoint_create + instance_private_endpoint_config_get
## Summary

`instance_private_endpoint_create` and `instance_private_endpoint_config_get` have no live coverage. Both are scoped to a single service.

## Scope

Extend `tests/integration_test.rs` with a `Private Endpoints` phase:

- `instance_private_endpoint_create` with a synthetic endpoint id appropriate for `ctx.provider` (AWS/GCP/Azure each have a different format — pick a format the API will accept without a real cloud resource backing it, or skip if all formats require real provider state).
- `instance_private_endpoint_config_get` — assert the config response deserializes and references the service.
- Tear down any binding created so the test is idempotent.

## Notes

- If the live API rejects synthetic endpoint ids (some clouds validate against the provider), this issue may need to switch to assertion-on-error: confirm the rejection is a 4xx with a sensible body, rather than provisioning a working endpoint.
- The org-level `organization_private_endpoint_config_get_list` lives in a separate issue — keep these distinct.

## Where this lands

Parent: #151

sdairs:
## Summary

Adds a Private Endpoints phase to `tests/integration_test.rs` covering the two service-scoped private endpoint methods that previously had no live coverage:

- `instance_private_endpoint_config_get` — runs twice (before and after the create attempt) and asserts the response contains non-empty `endpointServiceId` and `privateDnsHostname`.
- `instance_private_endpoint_create` — called with a provider-shaped but synthetic id derived from `ctx.run_id`.

### Assertion-on-error fallback (used)

Issue #160 explicitly flagged the risk that synthetic endpoint ids would be rejected by the control plane because no real cloud resource exists behind them. This PR takes that path:

The create step asserts the call either:

- succeeds (in which case the test registers inline cleanup via `instance_update` with `privateEndpointIds.remove`, then re-reads `config_get`), **or**
- fails with an unambiguous 4xx and a non-empty error message body.

Any other outcome — 5xx, network error, or 200 with a malformed payload — is treated as a real failure recorded via `FailureRecorder`. The synthetic id format is provider-aware (`vpce-…` for AWS, 19-digit decimal for GCP, GUID for Azure) and embeds `ctx.run_id` so leaked ids in API logs trace back to a specific run. The rationale is documented inline in the test.

All new steps are `NonBlocking` so the rest of the service CRUD lifecycle (IP access, scaling, delete) still runs even if the API contract for private endpoints shifts.

Closes #160

## Test plan

- [x] `cargo test -p clickhouse-cloud-api` — 86 + 6 tests pass; integration test is `#[ignore]` and unchanged on the non-ignored path.
- [x] `cargo clippy -p clickhouse-cloud-api -- -D warnings` (the invocation CI runs) — clean.
- [x] `cargo clippy -p clickhouse-cloud-api --tests` — the only warning under `integration_test.rs` is the pre-existing `scale_service_and_wait` (9/7) function; my code introduces no new lints.
- [ ] Cloud integration job (CI) — must run live to exercise the new phase.

### DIFF
--- a/crates/clickhouse-cloud-api/tests/integration_test.rs
+++ b/crates/clickhouse-cloud-api/tests/integration_test.rs
@@ -909,7 +909,219 @@ async fn cloud_service_crud_lifecycle() -> TestResult<()> {
             })
             .await?;
 
-        // ── 5. IP Access ─────────────────────────────────────────────
+        // ── 5. Private Endpoints ─────────────────────────────────────
+        //
+        // Cover `instance_private_endpoint_create` and
+        // `instance_private_endpoint_config_get` against the live service.
+        //
+        // `instance_private_endpoint_config_get` should always succeed: it
+        // returns the service-side endpoint service id + private DNS hostname
+        // and does not need a real provider-side endpoint to exist.
+        //
+        // `instance_private_endpoint_create` requires a `ServicPrivateEndpointePostRequest`
+        // whose `id` is a provider-specific identifier (AWS `vpce-…`, GCP
+        // numeric PSC id, Azure GUID). The control plane validates that id
+        // against the underlying provider, so a synthetic value with no real
+        // cloud resource behind it is expected to be rejected with a 4xx.
+        //
+        // The assertion is therefore: the call must either succeed (in which
+        // case we register inline cleanup, re-read config, and remove the
+        // binding) OR fail with an unambiguous 4xx structured error. Any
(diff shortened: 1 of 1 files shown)


iskakaushik (dismissed):
LGTM!

iskakaushik (approved):

PR #177, opened 2026-05-15
Cover postgres_instance_create_read_replica

### LINKED ISSUE #162: Cover postgres_instance_create_read_replica
## Summary

`postgres_instance_create_read_replica` has no live coverage.

## Scope

Extend `tests/integration_postgres_test.rs` with a `Read Replica` phase, inserted after the primary is in `running` state:

- `postgres_instance_create_read_replica` against the primary service id. Register the replica id with the cleanup registry.
- Poll for the replica to reach `running`.
- Verify the replica appears in `postgres_service_get_list` and `postgres_service_get` returns the expected primary reference.
- Tear down the replica explicitly before the primary delete phase (and confirm via cleanup that no orphaned replicas remain if the test fails mid-way).

## Where this lands

- `crates/clickhouse-cloud-api/tests/integration_postgres_test.rs`
- `crates/clickhouse-cloud-api/tests/integration/support.rs` — extend cleanup to handle replicas (likely just another postgres id, but the parent ref may matter for delete order).

Parent: #151

sdairs:
## Summary

- Extends `integration_postgres_test.rs` with a Read Replica phase that exercises `postgres_instance_create_read_replica` end-to-end (create, poll-to-running, list+get verification including primary reference, explicit delete) before the primary Delete phase.
- Adds replica-id tracking to `CleanupRegistry` (`register_postgres_replica` / `unregister_postgres_replica`); the cleanup pass deletes registered replicas before registered primaries, since the live API refuses to delete a primary while a replica still references it.
- Replica is registered in the cleanup registry immediately on create so a panic between create and the in-body teardown cannot leak the replica and break the primary delete.

## Cleanup ordering

Two complementary mechanisms keep replicas torn down before primaries:

1. **`CleanupRegistry`** — the cleanup pass iterates `postgres_replica_ids` to deletion before iterating `postgres_ids`. Registration happens immediately after the create call returns the id.
2. **Explicit in-body teardown** — the Read Replica phase deletes its replica and waits for 404 before the primary Delete phase runs. `unregister_postgres_replica` keeps the registry tidy on the happy path.

Both paths are documented in code comments (`CleanupRegistry` doc-comment and the Read Replica phase header in the test).

## Test plan

- [x] `cargo clippy -p clickhouse-cloud-api --tests -- -D warnings` (only pre-existing errors in `spec_coverage_test.rs` and `integration_test.rs` remain; nothing in files touched by this PR)
- [x] `cargo test -p clickhouse-cloud-api` (86 unit + 6 spec-coverage tests pass; integration_postgres test is `#[ignore]` per existing suite policy)
- [ ] Live cloud-integration run via PR CI to exercise the new phase end-to-end against production.

Closes #162

### DIFF
--- a/crates/clickhouse-cloud-api/tests/common/support.rs
+++ b/crates/clickhouse-cloud-api/tests/common/support.rs
@@ -132,6 +132,10 @@ impl TestContext {
         format!("clickhousectl-it-pg-{}", self.run_id)
     }
 
+    pub fn postgres_replica_name(&self) -> String {
+        format!("clickhousectl-it-pgrr-{}", self.run_id)
+    }
+
     pub fn postgres_run_tags(&self) -> Vec<ResourceTagsV1> {
         vec![
             ResourceTagsV1 {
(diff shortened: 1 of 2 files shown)


iskakaushik (dismissed):
LGTM!

iskakaushik (approved):

Issue #151, opened 2026-05-13
Close live API integration test coverage gap


sdairs:
## Summary

The Rust client at `crates/clickhouse-cloud-api/` exposes 100 async methods. The live integration tests in `tests/integration_test.rs` (service CRUD) and `tests/integration_postgres_test.rs` (postgres CRUD) exercise 25 of them. The other 75 are mock-tested in `tests/client_test.rs` only, which catches request-shape regressions but not live API drift.

The mock tests + `scripts/check-openapi-drift.py` will tell us when the spec moves. They won't tell us when a method that compiles locally returns 4xx in production. As `clickhousectl` ships more cloud commands, this gap turns into runtime errors for end users.

This issue groups the work to close that gap.

## Strategy

- **Three test binaries**, one per resource shape:
  - `integration_org_test.rs` (new) — org-scoped endpoints, no service provisioning.
  - `integration_test.rs` (extend) — endpoints needing one ClickHouse service.
  - `integration_postgres_test.rs` (extend) — endpoints needing one Postgres service.
- **Each binary = one `#[tokio::test]` lifecycle** with `FailureRecorder` accumulating non-blocking failures so a single run reports every broken endpoint, not just the first.
- **Always run all three on every PR.** Path-based selection deferred.
- **Failure policy**: NonBlocking by default; Blocking only when downstream phases depend on the step.

### Bootstrap
- [ ] #152 — `integration_org_test.rs` skeleton + secondary-user fixture

### Org suite (depends on #152)
- [ ] #153 — members + invitations
- [ ] #154 — custom roles CRUD
- [ ] #155 — activity log
- [ ] #156 — openapi key full lifecycle
- [ ] #157 — org prometheus + private endpoint config listing

### Service suite (extends `integration_test.rs`)
- [ ] #158 — query endpoint GET + DELETE
- [ ] #159 — password update + vertical scaling
- [ ] #160 — private endpoint create + config get
- [ ] #161 — ClickHouse settings round-trip
- [ ] #169 — scaling_schedule get / upsert / replace

### Postgres suite (extends `integration_postgres_test.rs`)
- [ ] #162 — read replica
- [ ] #163 — instance config POST/PATCH (blocked on models fix)

### Separate
- [ ] #165 — CLI↔client contract layer (scoping)

## Out of scope this round

- **ClickPipes, ClickStack** — WIP PRs cover them (#130, #144).
- **BYOC** — no plan to support in consuming clients.
[...]

PR #178, opened 2026-05-15
Add scaling_schedule round-trip coverage to service suite

### LINKED ISSUE #169: Cover scaling_schedule_get / replace / upsert in service integration suite
## Summary

`scaling_schedule_get`, `scaling_schedule_upsert`, and `scaling_schedule_replace` (all Beta-tagged in the OpenAPI spec) have no live coverage. They were missed in the original #151 scope.

## Scope

Extend `tests/integration_test.rs` with a `Scaling Schedule` phase, after the existing `Scaling` phase (so the service is in a known steady state):

- `scaling_schedule_get` — capture pre-test state (may legitimately be empty / unset).
- `scaling_schedule_upsert` — set a synthetic schedule (e.g. a single window with low min/max replicas that won't actually trigger during the test run).
- `scaling_schedule_get` — confirm the upsert is visible and round-trips.
- `scaling_schedule_replace` — replace with a different schedule. Verify via subsequent `get`.
- Restore the pre-test state (or delete the schedule if it didn't exist before). Idempotent post-state.

## Where this lands

## Notes

- Pick schedule windows that won't actually scale the test service during the test window — these endpoints are Beta and we want shape coverage, not real scaling behaviour entangled with the rest of the suite.
- All three methods are Beta in `crates/clickhouse-cloud-api/src/meta.rs` — confirm the affordance still renders correctly if the CLI surfaces these.

Parent: #151

sdairs:
## Summary

- Adds a Beta `Scaling Schedule` phase to `integration_test.rs` covering `scaling_schedule_get`, `scaling_schedule_upsert`, and `scaling_schedule_replace` (parent #151).
- Pre-state is captured via GET, an inert window is upserted then PATCH-replaced with another inert window, and every change is GET-verified.
- Pre-state restore is registered with `CleanupRegistry` so it runs in cleanup, not in the test body — a mid-phase failure still rolls back the synthetic schedule.

## Why the schedules are inert

The synthetic entries use `minReplicas=maxReplicas=1` and `min/maxReplicaMemoryGb=8`, identical to the current service state (the previous "scale in to 1 replica" step). Even if the window happens to be active when the test runs, the entry produces no scaling action — it's effectively a no-op. As a second line of defence, each window is one hour wide (1-2 UTC for upsert, 2-3 UTC for replace) and applies on Sunday only.

## Test plan

- [x] `cargo clippy -p clickhouse-cloud-api --tests` (no new warnings introduced by these changes)
- [x] `cargo test -p clickhouse-cloud-api` — 86+6+1 tests pass, ignored live tests still skipped
- [ ] Cloud integration test exercises the new phase end-to-end on PR CI

Closes #169

### DIFF
--- a/crates/clickhouse-cloud-api/tests/common/support.rs
+++ b/crates/clickhouse-cloud-api/tests/common/support.rs
@@ -372,6 +372,12 @@ pub struct CleanupRegistry {
     /// being deleted, table drops are redundant but harmless.
     tables: Vec<String>,
     api_key_ids: Vec<String>,
+    scaling_schedule_restores: Vec<ScalingScheduleRestore>,
+}
+
+pub struct ScalingScheduleRestore {
+    pub service_id: String,
+    pub pre_state: ScalingSchedule,
 }
 
 impl CleanupRegistry {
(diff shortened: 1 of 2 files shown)


iskakaushik (approved):
LGTM!

PR #173, opened 2026-05-15
Add integration_org_test.rs skeleton for org-only live coverage

sdairs:
## Summary

Bootstraps the third live integration test binary so the per-area org coverage issues (#153–#157) have somewhere to plug in without coupling to the 15-minute ClickHouse service provisioning lifecycle. No real test phases yet — those land in their own issues.

- New `tests/integration_org_test.rs` with a single `#[tokio::test] #[ignore]` lifecycle that verifies org access and follows the existing `FailureRecorder` + `CleanupRegistry` shape.
- `TestContext` gains `secondary_user_id` (populated from `CLICKHOUSE_CLOUD_TEST_SECONDARY_USER_ID`). Field is `Option<String>` so the existing two suites stay green; a `secondary_user_id()` accessor errors with a clear message when the org suite needs it.
- `CleanupRegistry` extended with idempotent `register_role` / `register_invitation` (404 swallowed on teardown, mirrors api-key cleanup).
- `org_run_tags` / `org_run_tag_filters` helpers added for any future org-scoped resource that supports tagging.
- Workflow `cloud-integration.yml` wires the new binary as a third step and adds `CLICKHOUSE_CLOUD_TEST_SECONDARY_USER_ID` to the env block (via `secrets.CLICKHOUSE_CLOUD_TEST_SECONDARY_USER_ID`).
- README + CLAUDE.md updated to mention the new suite and env var.

Closes #152
Parent: #151

## Test plan

- [x] `cargo build -p clickhouse-cloud-api` clean
- [x] `cargo clippy -p clickhouse-cloud-api --test integration_org_test --test integration_postgres_test --lib -- -D warnings` clean (pre-existing clippy errors in `integration_test.rs` and `spec_coverage_test.rs` are unrelated and present on `main`)
- [x] `cargo test -p clickhouse-cloud-api` — all suites green; `integration_org_test` registers one ignored lifecycle test as expected
- [ ] Live `--ignored` run via cloud-integration workflow on this PR (relies on `CLICKHOUSE_CLOUD_TEST_SECONDARY_USER_ID` being added to repo secrets before the third step can succeed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #154, opened 2026-05-13
Cover custom org roles CRUD in org integration suite


sdairs:
## Summary

`organization_roles_get_list`, `organization_role_post`, `organization_role_get`, `organization_role_patch`, `organization_role_delete` have no live coverage.

## Scope

- `organization_roles_get_list` — capture pre-test state for sanity assertions.
- `organization_role_post` — create a synthetic role named `clickhousectl-it-role-{run_id}`. Register for cleanup.
- `organization_role_get` — confirm the new role is returned and fields match the post body.
- `organization_role_patch` — modify a property of the role (description, permissions). Verify via subsequent `get`.
- `organization_role_delete` — delete. Verify via 404 on `get`.

## Where this lands

- `crates/clickhouse-cloud-api/tests/integration_org_test.rs` (new phase)

## Notes

- The role name must include `ctx.run_id` so concurrent CI runs don't collide.
- Custom roles don't need to be assigned to anyone to exercise CRUD — the secondary user is uninvolved.

Parent: #151

Issue #152, opened 2026-05-13
Add integration_org_test.rs skeleton for org-only live coverage


sdairs:
## Summary

Org-scoped endpoints (members, invitations, roles, etc.) have no live coverage today because adding them to the service lifecycle test would couple them to a 15-minute provisioning step they don't need.

Create a third test binary that mirrors the FailureRecorder/CleanupRegistry pattern in the existing two and unblocks the per-area org coverage issues.

## Scope

- Add `crates/clickhouse-cloud-api/tests/integration_org_test.rs` with one `#[tokio::test] #[ignore]` lifecycle that exercises basic org access (e.g. `organization_get` for `ctx.org_id`) and finishes cleanly.
- Extend `tests/integration/support.rs`:
  - Add `org_run_tags` / `org_run_tag_filters` helpers if any resource type needs tagging.
  - Add `CLICKHOUSE_CLOUD_TEST_SECONDARY_USER_ID` to `TestContext::from_env()`.
  - Extend `CleanupRegistry` with `register_role` and `register_invitation` so area issues have somewhere to plug in.
- Update `.github/workflows/cloud-integration.yml` to invoke `cargo test --test integration_org_test -- --ignored` and pass the new env var.
- Update `README.md` / `CLAUDE.md` documenting the new env var and suite.

## Out of scope

- Any actual endpoint exercises beyond the org sanity check — those land in their own issues.

## Notes

- Don't move the existing "Org Checks" phase out of the service test; the suites stand alone.

Parent: #151

PR #180, opened 2026-05-15
Cover activity log endpoints in org integration suite

### LINKED ISSUE #155: Cover activity log endpoints in org integration suite
## Summary

`activity_get_list` and `activity_get` have no live coverage.

## Scope

- Place this phase **after** the custom-roles phase so the activity log has a known recent event to assert on (role create + delete).
- `activity_get_list` — fetch recent activity, assert non-empty.
- Find an activity entry attributable to this test run (matching the synthetic role name in the entry's metadata if available, or just an entry within the last few minutes for the current org).
- `activity_get` for that entry — assert fields populated.

## Where this lands

- `crates/clickhouse-cloud-api/tests/integration_org_test.rs` (new phase, after roles)

## Notes

- Activity log writes are eventually consistent — poll for the expected entry rather than asserting on the first read.
- If activity entries don't carry enough metadata to attribute precisely, fall back to "an entry with timestamp within the test window" as the assertion.

Parent: #151

sdairs:
## Summary

Adds an Activity Log phase to `integration_org_test.rs` covering the two previously-uncovered org-scoped methods.

- `activity_get_list` — polled for up to ~30s (3s interval) for an entry whose `createdAt` falls within the current test window; falls back gracefully on a quiet org by recording a NonBlocking failure rather than panicking.
- `activity_get` — fetched for the entry found above; asserts `id` round-trips, `organizationId` is non-empty, `createdAt` is non-zero.
- Both steps are `NonBlocking` — eventual-consistency misses do not fail the run, they surface in the `FailureRecorder` summary.

## Phase ordering

Placed after the Org Access phase. Once the Custom Roles phase (#154) lands directly above this one, the role create + delete events become the known-recent activity the polling check correlates against; until then the phase degrades to "any entry within the test window". Documented inline so a future reviewer adding #154 knows where to slot it.

## Polling strategy

`poll_until` from `support.rs` with a 30s budget and 3s interval. On timeout the inner error is wrapped into a clean NonBlocking failure message ("no activity entry observed within 30s budget: ...") for readable `FailureRecorder` output.

## Stacked PR note

This PR is stacked on [#173](https://github.com/ClickHouse/clickhousectl/pull/173) (#152 bootstrap) and is best-reviewed after the Custom Roles phase (#154) lands. GitHub will auto-retarget `--base` to `main` when #173 merges.

Closes #155
Parent: #151

## Test plan

- [x] `cargo build -p clickhouse-cloud-api` clean
- [x] `cargo clippy -p clickhouse-cloud-api --test integration_org_test -- -D warnings` clean (pre-existing clippy errors in `spec_coverage_test.rs` and `integration_test.rs` are present on the base branch and unrelated)
- [x] `cargo test -p clickhouse-cloud-api` — all suites green; `integration_org_test::cloud_org_lifecycle` registers as one ignored lifecycle test as expected
- [ ] Live `--ignored` run via cloud-integration workflow (requires the existing `CLICKHOUSE_CLOUD_*` secrets already wired up by #152)


iskakaushik (dismissed):
LGTM!

iskakaushik (approved):

PR #198, opened 2026-05-19
Fix self-update: download tarball and extract binary

### LINKED ISSUE #196: darwin/arm64 chctl update failing
on darwin/arm64 with tar.gz install:

```
$ chctl --version
clickhousectl 0.1.18

$ chctl update
Error: Download failed: No compatible binary found for this platform (expected clickhousectl-aarch64-apple-darwin)
```

graphaelli:
The release workflow uploads `clickhousectl-{target}-v{version}.tar.gz` archives, but `update.rs` looked for a raw binary asset named `clickhousectl-{target}` and wrote the response bytes directly to disk. This meant `chctl update` always failed with "No compatible binary found for this platform", and would have clobbered the running binary with a gzipped tarball if the name had matched.

Build the asset name from `release.tag_name` to match the workflow, then stream the download through `GzDecoder` + `tar::Archive` and pull out the `clickhousectl` entry before the atomic rename.

closes #196

### DIFF
--- a/crates/clickhousectl/src/update.rs
+++ b/crates/clickhousectl/src/update.rs
@@ -1,9 +1,12 @@
 use crate::error::{Error, Result};
 use crate::paths;
+use flate2::read::GzDecoder;
 use serde::Deserialize;
 use std::fs;
+use std::io::{self, Cursor};
 use std::path::PathBuf;
 use std::time::{SystemTime, UNIX_EPOCH};
+use tar::Archive;
 
 const GITHUB_REPO: &str = "ClickHouse/clickhousectl";
 const CHECK_INTERVAL_SECS: u64 = 24 * 60 * 60; // 24 hours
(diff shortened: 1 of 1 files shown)


sdairs:
Nice fix — diagnosis and approach match what I had in #199 (which I've now closed in favour of this), and your per-step error messages + negative test are better than what I wrote.

One small suggestion: the four `Error::Download(...)` wraps inside `extract_binary_from_archive` are semantically about extraction, not download (the download has already succeeded by the time we get here). `Error::Extract` already exists in `error.rs` for exactly this case — swapping to it would give users a clearer *"Extraction failed: …"* prefix when an archive is malformed. Pure polish, doesn't block.

For context: I'll open a separate small PR on top of this with the version bumps (`clickhousectl`, `clickhouse-cloud-api`, the dep ref between them, and `npm/package.json` to 0.2.2 in lockstep) and a CLAUDE.md note that releases need all three bumped together now. Keeping it separate so this fix can land cleanly.

sdairs (approved):

graphaelli:
good call, i can change those real quick

sdairs (approved):

PR #199, opened 2026-05-19
Fix self-update broken by archive packaging; bump to 0.2.2

sdairs:
## Summary

- `clickhousectl update` broke in 0.2.1: the release workflow now ships `clickhousectl-{target}-v{version}.tar.gz` archives, but `update.rs` was still looking for a raw asset named `clickhousectl-{target}` (no version, no extension), so the asset lookup always failed with *"No compatible binary found for this platform"*. The npm installer was updated for the archive layout in the dist rework; this codepath was missed.
- Fix: download the `.tar.gz`, extract the `clickhousectl` entry via the already-vendored `flate2` + `tar` crates, then feed the bytes into the existing temp-write + chmod + atomic-rename path. New unit test round-trips a release-layout tarball through the extractor.
- Bump `clickhousectl`, `clickhouse-cloud-api`, and `npm/package.json` to **0.2.2** in lockstep (plus the dep reference inside `clickhousectl`).
- `CLAUDE.md`: require lockstep version bumps on every release, replacing the prior "only when the API client changed" rule for `clickhouse-cloud-api`.

> Note: existing 0.2.1 installs still need a one-time manual reinstall (`npm i -g clickhousectl` / `cargo binstall` / direct binary). The fix only reaches users via 0.2.2.

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo test -p clickhousectl --bin clickhousectl update::` — 5/5 pass, including the new tarball round-trip test
- [x] `cargo clippy -p clickhousectl --all-targets` — clean
- [x] Verified the real v0.2.1 release archive layout matches what the new extractor expects (`clickhousectl-{target}-v{ver}/clickhousectl`)
- [ ] After merge + tag `v0.2.2`: install 0.2.1 from a fresh source, then run `chctl update` and confirm it successfully upgrades to 0.2.2

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -137,7 +137,7 @@ cargo add -p clickhouse-cloud-api url
 - If the user references a GitHub issue (e.g. "work on issue 3"), use `gh issue view 3` to get the details, then create a branch like `issue-3-short-description`.
 - Update `README.md` and any relevant documentation as part of the change — PRs should include doc updates for new or changed functionality.
 - Commit to the branch, push, and create a PR with `gh pr create`.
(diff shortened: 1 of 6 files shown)


iskakaushik (approved):

sdairs:
Closing in favour of #198 — same fix, opened first, and the per-step error messages and negative test there are nicer than mine. Will open a small follow-up PR with just the lockstep version bumps + CLAUDE.md release-process update on top of #198.

PR #207, opened 2026-05-21
Fix misalignment with OpenAPI spec for Kafka pipes

markdawson:
## Summary

`chctl cloud clickpipe list` crashes on services that contain a Kafka
pipe whose `reversePrivateEndpointIds` field is `null` (43 of 78 Kafka
pipes in the test org). One bad pipe takes down the whole response.

The OpenAPI spec declares the field non-nullable `array` but the live
API emits `null`. `#[serde(default)]` rescues missing keys, not
explicit `null`, so deserialization fails with `invalid type: null,
expected a sequence`.

## Fix

Add a `null_to_empty` serde helper that coerces JSON `null` to an
empty `Vec`. Apply it to the three Kafka source structs that carry
the offending field. Narrow scope: this is a bridge until the
upstream OpenAPI emitter is fixed 
(`control-plane/apps/openapi/src/service/OpenapiDocRegistry.ts:151-157`
drops `nullable: true` for `type: 'array'` fields). Once that lands,
re-running the codegen flips these fields to `Option<Vec<String>>`
and the shim comes out.

## Test plan

- [x] `cargo test -p clickhouse-cloud-api` — 119/119 pass. New
      `list_click_pipes_with_null_array_fields` test mirrors the
      real-world response shape and is the regression guard.
- [x] E2E pipe creations now round-trip through
      `click_pipe_get_list` so future server-shape regressions
      surface in tests.
- [x] `chctl cloud clickpipe list` against all 108 services in the
      test org — 0 failures (was 15+ before).

## Files

| File | Change |
|---|---|
| `src/serde_helpers.rs` | New — `null_to_empty` deserializer |
| `src/lib.rs` | `pub mod serde_helpers;` |
| `src/models.rs` | Annotate 3 `reversePrivateEndpointIds` fields |
| `tests/client_test.rs` | Regression test |
| `tests/clickpipes/stages/mod.rs` | E2E list round-trip |

### DIFF
--- a/crates/clickhouse-cloud-api/src/lib.rs
+++ b/crates/clickhouse-cloud-api/src/lib.rs
@@ -21,6 +21,7 @@ pub mod error;
 pub mod meta;
 #[allow(non_camel_case_types)]
 pub mod models;
+pub mod serde_helpers;
 
 pub use client::Client;
 pub use error::Error;
--- a/crates/clickhouse-cloud-api/src/models.rs
+++ b/crates/clickhouse-cloud-api/src/models.rs
@@ -7657,7 +7657,7 @@ pub struct ClickPipeKafkaSource {
     pub iam_role: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none", default)]
     pub offset: Option<ClickPipeKafkaOffset>,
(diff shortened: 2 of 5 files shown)


sdairs (approved):
lgtm

PR #211, opened 2026-05-23
Use REST API in drift checker to fix 401 on issue creation

sdairs:
## Summary

- `gh issue create` and `gh issue list` go through GraphQL and do a `viewer { id }` lookup as part of every call. The Actions `GITHUB_TOKEN` is an installation token, not a user token, so that introspection returns `HTTP 401: Bad credentials` — which is what just killed the drift workflow the first time it actually had drift to file ([run 26328411342](https://github.com/ClickHouse/clickhousectl/actions/runs/26328411342)). The bug went unnoticed since the workflow shipped because every prior run found zero drift and never reached `create_issue`.
- Switch the three gh call sites (`ensure_label_exists`, `open_drift_issues`, `create_issue`) to `gh api` against the REST endpoints. REST accepts the installation token directly given the existing `issues: write` permission on the workflow. A small `_repo_slug()` helper resolves `owner/repo` from `GITHUB_REPOSITORY` in Actions and falls back to `gh repo view` locally.

## Test plan

- [x] `python3 scripts/check-openapi-drift.py --dry-run` reproduces the same 4-gap report on the current spec.
- [x] `gh api repos/ClickHouse/clickhousectl/issues --method GET -f labels=openapi-drift -f state=open` exercises the exact list call the script will make in CI and returns `[]` cleanly.
- [x] `gh api repos/ClickHouse/clickhousectl/labels` confirms the `openapi-drift` label already exists, so `ensure_label_exists` will hit a 422 in CI — silently absorbed (no `check=True`), keeping the call idempotent.
- [ ] After merge, re-trigger the workflow (`gh workflow run openapi-drift.yml`) so it files the drift issue it failed to create today.

### DIFF
--- a/scripts/check-openapi-drift.py
+++ b/scripts/check-openapi-drift.py
@@ -446,14 +446,41 @@ def check_missing_fields(
 # ---------------------------------------------------------------------------
 
 
+def _repo_slug() -> str:
+    """Return `owner/repo` for the current repo.
+
+    Prefers `GITHUB_REPOSITORY` (set by GitHub Actions); falls back to
+    `gh repo view` for local runs.
+    """
+    slug = os.environ.get("GITHUB_REPOSITORY")
+    if slug:
+        return slug
+    result = subprocess.run(
+        ["gh", "repo", "view", "--json", "nameWithOwner", "-q", ".nameWithOwner"],
+        capture_output=True, text=True, check=True,
+    )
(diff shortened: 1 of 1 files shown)


sdairs:
Closing — the underlying failure was transient (the next scheduled run succeeded without code changes). Misdiagnosed the cause; no fix needed.

PR #215, opened 2026-05-23
Unblock postgres_instance_config PATCH (#163)

### LINKED ISSUE #163: Cover postgres_instance_config_post + postgres_instance_config_patch
## Summary

`postgres_instance_config_post` and `postgres_instance_config_patch` have no live coverage. The deferred "Runtime Config" phase in `tests/integration_postgres_test.rs` documents the failure: `PgConfig`'s required non-Option fields serialize as `null`/`0`/first-variant, and the live API rejects with `Validation failed for following fields: pg_config.*`.

## Current state (verified against latest spec)

- `pgConfig`: no `required` array, 23 properties, none described as "Optional" → resolver marks all required → `PgConfig` in `models.rs:11919` has only bare-typed fields.
- `postgresInstanceConfig`: `required: ["pgConfig", "pgBouncerConfig"]` — wrapper forces both nested objects on every request.
- `pgBouncerConfig`: free-form map with `minProperties: 1`, generated as empty struct → serializes to `{}` and fails its own constraint.
- No `OPTIONALITY_EXEMPTIONS` entries for any of these structs.

Net effect: no shape constructible from the current typed models is accepted by the live API.

## Minimum exemption set to unblock

- All 23 `PgConfig.*` fields → `Option<T>` + `#[serde(skip_serializing_if = "Option::is_none")]`
- `PostgresInstanceConfig.pgConfig` → `Option<T>`
- `PostgresInstanceConfig.pgBouncerConfig` → `Option<T>`

(Delta vs the earlier draft list: `ssl_min_protocol_version` needs adding — it's in the model now but wasn't in the original suggestion.)

These entries go in `OPTIONALITY_EXEMPTIONS` in `tests/spec_coverage_test.rs`. Rationale: spec is wrong; see follow-up comment for upstream validation plan.

## Tradeoff

`PostgresInstanceConfig` is also the GET response shape. Making all `PgConfig.*` fields `Option<T>` means consumers of the GET response unwrap each field even though the API always returns them populated. Acceptable cost for the unblock.

## Scope (once unblocked)

Extend `tests/integration_postgres_test.rs` `Runtime Config` phase:

- `postgres_instance_config_post` — set an initial config (e.g. `max_connections`).
- `postgres_instance_config_get` — confirm the change is visible.
- `postgres_instance_config_patch` — change `max_connections` to another valid value.
- `postgres_instance_config_get` — confirm the patched value.
- Optionally restore the original.

## Where this lands

- `crates/clickhouse-cloud-api/tests/integration_postgres_test.rs`

Parent: #151

sdairs:
## Summary

- Flip every `PgConfig` field (23) plus both `PostgresInstanceConfig` nested fields (`pgConfig`, `pgBouncerConfig`) to `Option<T>` with `skip_serializing_if = "Option::is_none"`, so `Client::postgres_instance_config_patch` can express a real partial update. Pair the change with 25 `OPTIONALITY_EXEMPTIONS` entries in `spec_coverage_test.rs`; the staleness detector will tell us automatically when upstream fixes the spec.
- Re-enable the deferred PATCH round-trip in `integration_postgres_test.rs` (GET baseline → PATCH `max_connections` → poll-until-applied → PATCH back).
- Add a gated `CLICKHOUSE_CLOUD_POSTGRES_CONFIG_PROBE=1` probe phase that fires the 6 × 2 behaviour-matrix scenarios from the issue's follow-up comment via raw `reqwest` and prints a markdown table on stderr. Default off so it doesn't run on every integration test.

Closes #163.

## Follow-ups

- Capture the behaviour matrix and file the upstream spec issue (draft text already lives in #163). To capture:
  ```
  CLICKHOUSE_CLOUD_POSTGRES_CONFIG_PROBE=1 \
    cargo test -p clickhouse-cloud-api --test integration_postgres_test cloud_postgres_crud_lifecycle -- --ignored --nocapture
  ```
- Once the spec is fixed upstream, the new `OPTIONALITY_EXEMPTIONS` entries become stale and `field_optionality_matches_spec` will flag them — remove the block.

## Test plan

- [x] `cargo build --tests` (whole workspace)
- [x] `cargo clippy --tests`
- [x] `cargo test` (all non-ignored tests pass)
- [x] `cargo test -p clickhouse-cloud-api --test spec_coverage_test field_optionality_matches_live_spec -- --ignored` — 25 new exemptions all register against the live spec, no stale entries
- [ ] `cargo test -p clickhouse-cloud-api --test integration_postgres_test -- --ignored --nocapture` (live cloud creds required)
- [ ] `CLICKHOUSE_CLOUD_POSTGRES_CONFIG_PROBE=1 cargo test -p clickhouse-cloud-api --test integration_postgres_test -- --ignored --nocapture` — capture behaviour matrix for upstream

### DIFF
--- a/crates/clickhouse-cloud-api/src/models.rs
+++ b/crates/clickhouse-cloud-api/src/models.rs
@@ -12201,10 +12201,10 @@ pub struct PgConfig {
 /// `postgresInstanceConfig` from the ClickHouse Cloud API.
 #[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
 pub struct PostgresInstanceConfig {
-    #[serde(rename = "pgBouncerConfig", skip_serializing_if = "Option::is_none", default)]
-    pub pg_bouncer_config: Option<PgBouncerConfig>,
-    #[serde(rename = "pgConfig", skip_serializing_if = "Option::is_none", default)]
-    pub pg_config: Option<PgConfig>,
+    #[serde(rename = "pgBouncerConfig", default)]
+    pub pg_bouncer_config: PgBouncerConfig,
+    #[serde(rename = "pgConfig", default)]
+    pub pg_config: PgConfig,
 }
 
 /// `postgresInstanceUpdateConfigResponse` from the ClickHouse Cloud API.
--- a/crates/clickhouse-cloud-api/src/models.rs
+++ b/crates/clickhouse-cloud-api/src/models.rs
@@ -11968,10 +11968,10 @@ pub struct PgConfig {
 /// `postgresInstanceConfig` from the ClickHouse Cloud API.
 #[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
 pub struct PostgresInstanceConfig {
-    #[serde(rename = "pgBouncerConfig", skip_serializing_if = "Option::is_none", default)]
-    pub pg_bouncer_config: Option<PgBouncerConfig>,
-    #[serde(rename = "pgConfig", skip_serializing_if = "Option::is_none", default)]
-    pub pg_config: Option<PgConfig>,
+    #[serde(rename = "pgBouncerConfig", default)]
+    pub pg_bouncer_config: PgBouncerConfig,
+    #[serde(rename = "pgConfig", default)]
+    pub pg_config: PgConfig,
 }
 
 /// `postgresInstanceUpdateConfigResponse` from the ClickHouse Cloud API.
--- a/crates/clickhouse-cloud-api/src/models.rs
+++ b/crates/clickhouse-cloud-api/src/models.rs
@@ -12155,7 +12155,7 @@ pub struct PgConfig {
     #[serde(skip_serializing_if = "Option::is_none", default)]
     pub effective_cache_size: Option<serde_json::Value>,
     #[serde(skip_serializing_if = "Option::is_none", default)]
-    pub effective_io_concurrency: Option<i64>,
+    pub effective_io_concurrency: Option<serde_json::Value>,
     #[serde(skip_serializing_if = "Option::is_none", default)]
     pub idle_in_transaction_session_timeout: Option<serde_json::Value>,
     #[serde(skip_serializing_if = "Option::is_none", default)]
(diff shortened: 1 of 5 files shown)


sdairs:
## Blocked: response shape returns string-encoded numbers

Marking this PR as draft and blocked. The request-shape fix in this branch works (PATCH/POST now succeed against the live API per the captured matrix), but the **response** shape can't be deserialised into the spec-declared types.

### What we see

For `pgConfig` fields the spec declares as `integer` / `number`, the live API returns the value wrapped in a JSON string. Example response from a successful PATCH (matrix body 2):

```json
{"result":{"pgConfig":{"max_connections":"200"},"pgBouncerConfig":{},"message":"…"}}
```

`max_connections` is `"200"` (string), not `200` (number). Deserialising into `Option<i64>` fails with:

```
JSON error: invalid type: string "107", expected i64 at line 1 column 46
```

This affects every strictly-numeric `pgConfig` field — `max_connections`, `effective_io_concurrency`, `max_worker_processes`, `max_parallel_workers`, `max_parallel_workers_per_gather`, `max_parallel_maintenance_workers`, `random_page_cost`. The mixed-type fields (`type: ["string","integer"]` like `work_mem`, `*_timeout`, etc.) are already typed as `serde_json::Value`, so they're unaffected.

The same asymmetry applies in both directions: requests must send JSON numbers (string-encoded values are rejected), but responses come back string-encoded. So we can't just flip the field types — the model would then serialise wrong on the request side.

### Why this PR is paused

A client-side workaround (custom `deserialize_with` that accepts both number and string) was considered and rejected — we don't want to mangle types to paper over an API inconsistency. This needs to be fixed on the API side: response numeric fields should match the spec's declared `integer`/`number` types.

### Next steps

- [ ] File / link upstream API issue for response-side type mismatch on `pgConfig` numeric fields
- [ ] Unblock once the API returns numbers as numbers
- [ ] Re-enable the gated integration round-trip (`patch pgConfig.max_connections` in `integration_postgres_test.rs`)

sdairs:
## Follow-up: `run_pg_config_probe` looks like a one-off diagnostic

During code review, `run_pg_config_probe` (in `tests/integration_postgres_test.rs`) reads as a one-off evidence-gathering tool rather than a permanent test:

- It's gated behind `CLICKHOUSE_CLOUD_POSTGRES_CONFIG_PROBE` so it never runs in normal CI.
- It's `StepKind::NonBlocking` so it can't fail the test run.
- Its output is a markdown table printed to stderr — not an assertion, not a regression guard. The table is meant to be pasted into the upstream spec issue (#163).

Once the spec issue is resolved and the matrix is captured, this function has no ongoing purpose and becomes dead weight in the integration test binary. It also duplicates the library's auth/base-URL handling (raw `reqwest::Client::new()` + manual env-var reads instead of going through `Client`), which will silently drift if the library's env-var names or defaults change.

**Suggested follow-up:**
1. **Drop it from this PR** — run the probe as a local scratch file, paste the results into the spec issue, and keep only the permanent test coverage (the PATCH round-trip + serialization tests). Or,
2. **If it should stay in-repo**, move it to a standalone `#[ignore]` test or an `examples/` binary so it's clearly separated from the integration suite, and document that it's for spec-issue evidence gathering only.

This is non-blocking for merging the PR — the `PgConfig` optionality fix itself is correct and well-tested — but worth resolving before the probe code calcifies.

iskakaushik (dismissed):
LGTM

iskakaushik (approved):

PR #218, opened 2026-05-24
Try local before remote when resolving version specs (#217)

### LINKED ISSUE #217: Version resolution should try local first, then fall back to remote
## Problem

When resolving a version spec for `local server start -v <spec>` (or `local install`), we currently make a remote lookup (GitHub API / builds.clickhouse.com) before checking whether the requested version is already installed locally. This means a fully-local request can still fail or hang on a remote call.

## Desired behaviour

Always try to resolve locally first. Only fall back to remote lookup if no local match exists.

- `-v 26.4` → look for any installed `26.4.x.y`. If found, use it. Otherwise, look up remotely.
- `-v 26.4.1.1` → look for an exact local match. If found, use it. Otherwise, look up remotely.
- `-v 26.4.1` → same pattern: look for any installed `26.4.1.y`. If found, use it. Otherwise, look up remotely.
- `-v stable` / `-v lts` / `-v latest` → these are channel/floating specs and should continue to consult the remote, since "what's stable today" is a remote concept.

## Why

- Faster: no network round-trip when the bits are already on disk.
- Works offline: starting a server with an already-installed version shouldn't require internet.
- Predictable: if the user has the version locally, that's what they get.

sdairs:
Closes #217.

## Summary
- Concrete specs (`25`, `25.12`, `25.12.5.44`) now resolve from already-installed versions before any network call, so `install`/`use`/`server start -v` work offline when the bits are on disk and skip the GitHub/builds round-trip otherwise.
- Floating specs (`latest`, `stable`, `lts`) continue to consult the remote — "what's stable today" is a remote concept by definition.
- 3-part specs (e.g. `25.12.9`) remain rejected at the parser, matching prior behaviour. Confirmed in plan as out-of-scope.

## Shape of the change
- `version_manager::resolve` gets a pure `find_local_match(spec, &[String])` plus a `try_resolve_local(spec)` I/O wrapper.
- `version_manager::install` gains two wrappers — `install_local_first` (for `install`) and `ensure_installed_local_first` (for `use` and `server start -v`) — that try local first and only call `resolve()` on miss. UX strings are lifted verbatim from the pre-existing post-resolve check so messaging stays consistent.
- The three call sites in `local/mod.rs` (`install`, `use_version`, `start_server`) collapse to a single wrapper call each.
- The post-resolve local check in `install_resolved`/`ensure_installed` is kept as defence-in-depth — it still fires for `install stable`/`install lts`, which start floating and may then match a local minor.

## Test plan
- [x] `cargo build`
- [x] `cargo clippy --all-targets`
- [x] `cargo test -p clickhousectl` (286 unit + 37 request-shape + 1 new integration test, all green)
- [x] 10 new unit tests for `find_local_match` covering: floating-spec → None, major prefix match, major component-boundary guard (rejects `250.x.y.z` for `Major(25)`), minor component-boundary guard (rejects `25.120.x.y` for `Minor(25,12)`), exact match (no partial), newest-wins ordering.
- [x] New integration test (`tests/local_install_local_first_test.rs`) pre-seeds a fake `$HOME/.clickhouse/versions/25.12.9.61/clickhouse`, runs `local install 25.12 --json`, asserts exit 0, stderr contains "already installed", and stderr does NOT contain "Resolving". This is the regression guard for "no network call when local match exists" — the whole point of the issue.
- [ ] Manual offline smoke: `local install 25.12`, `local use 25.12`, `local server start -v 25.12 --foreground` with `25.12.9.61` pre-installed (covered by integration test above for `install`; `use` and `server start` follow the same code path).

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to version resolution order and offline UX; remote fallback and floating specs are unchanged, with tests guarding local matching.
> 
> **Overview**
> Concrete ClickHouse version specs (`25`, `25.12`, exact `x.y.z.w`) are now satisfied from **already-installed** binaries under `~/.clickhouse/versions` before `resolve()` hits GitHub or builds.clickhouse.com. That applies to **`local install`**, **`local use`**, and **`local server start -v`**, so those flows can work offline and skip the resolve round-trip when a matching build is on disk.
> 
> **`try_resolve_local`** / **`find_local_match`** implement the disk lookup (newest-first, with prefix boundaries so e.g. `Major(25)` does not match `250.x`). **`latest`**, **`stable`**, and **`lts`** still always go remote. **`install_local_first`** and **`ensure_installed_local_first`** wrap the old resolve/install path on miss; install keeps the same “already installed” / `--force` messaging, while use/server start reuse **`ensure_installed`** semantics silently when local matches.
> 
> Call sites in **`local/mod.rs`** no longer inline resolve logic. Post-resolve local checks in **`install_resolved`** / **`ensure_installed`** remain as defense in depth. Coverage adds unit tests for **`find_local_match`** and an integration test that **`local install 25.12`** must not print **“Resolving”** when `25.12.9.61` is pre-seeded.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0f3329bb3f30670a3a9750f575c2e52570358ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,8 @@ clickhousectl local remove 25.12.5.44
 
 `local use` also creates a symlink at `~/.local/bin/clickhouse` pointing to the selected version's binary, so the plain `clickhouse` command (e.g. `clickhouse local`, `clickhouse client`) is on PATH. Pass `--no-global` to skip. If a regular file already exists at that path it is left alone with a warning. `local remove` of the active default version also clears the symlink.
 
+Concrete version specs (`25`, `25.12`, `25.12.5.44`) are resolved from your local installs first, so they work offline and don't pay a network round-trip when the bits are already on disk. Floating channels (`stable`, `lts`, `latest`) always consult the remote, since "what's stable today" can change.
+
 #### ClickHouse binary storage
 
 ClickHouse binaries are stored in a global repository, so they can be used by multiple projects without duplicating storage. Binaries are stored in `~/.clickhouse/`:
--- a/README.md
+++ b/README.md
@@ -91,8 +91,6 @@ clickhousectl local remove 25.12.5.44
 
 `local use` also creates a symlink at `~/.local/bin/clickhouse` pointing to the selected version's binary, so the plain `clickhouse` command (e.g. `clickhouse local`, `clickhouse client`) is on PATH. Pass `--no-global` to skip. If a regular file already exists at that path it is left alone with a warning. `local remove` of the active default version also clears the symlink.
 
-Concrete version specs (`25`, `25.12`, `25.12.5.44`) are resolved from your local installs first, so they work offline and don't pay a network round-trip when the bits are already on disk. Floating channels (`stable`, `lts`, `latest`) always consult the remote, since "what's stable today" can change.
-
 #### ClickHouse binary storage
 
 ClickHouse binaries are stored in a global repository, so they can be used by multiple projects without duplicating storage. Binaries are stored in `~/.clickhouse/`:
--- a/crates/clickhousectl/src/local/mod.rs
+++ b/crates/clickhousectl/src/local/mod.rs
@@ -87,10 +87,7 @@ async fn install(version_spec: &str, force: bool, json: bool) -> Result<()> {
     let spec = version_manager::parse_version_spec(version_spec)?;
     let platform = version_manager::platform::Platform::detect()?;
 
-    eprintln!("Resolving {}...", spec);
-    let resolved = version_manager::resolve::resolve(&spec, &platform).await?;
-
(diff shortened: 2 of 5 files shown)


sdairs:
bugbot run

iskakaushik (approved):
Local-first version resolution: concrete specs resolve from installed builds before any network call, floating specs (`latest`/`stable`/`lts`) still go remote. `find_local_match` is a pure matcher with solid component-boundary guards (rejects `250.x` for Major(25), `25.120.x` for Minor(25,12)), backed by 10 unit tests + a regression integration test asserting no "Resolving" on a local hit. LGTM.

Issue #235, opened 2026-06-03
Remove storageSize from Postgres create/update requests


sdairs:
The ClickHouse Cloud OpenAPI spec no longer includes `storageSize` on the Postgres request bodies (`PostgresServicePostRequest`, `PostgresServicePatchRequest`, `BasePostgresService`, `PostgresServiceListItem`). It remains only on the `PostgresService` response.

`clickhouse-cloud-api/models.rs` still carries the field on those request structs (stale vs the vendored snapshot, which is already correct), and the CLI exposes a required `--storage-gb` flag on `postgres create` / `postgres update`.

This makes `postgres create` require a flag the API no longer accepts. Remove `storage_size` from the request models and drop the `--storage-gb` CLI flag, keeping `storageSize` on the `PostgresService` response (still displayed in `get`).

PR #240, opened 2026-06-05
Detect struct fields removed from the spec but left in models.rs (#237)

### LINKED ISSUE #237: OpenAPI drift detection misses fields removed from the spec but still present in models.rs
## Problem

Our OpenAPI drift detection is **one-directional for fields** (spec → code). It catches fields *added* to the spec that are missing from `models.rs`, and optionality flips (`T` ↔ `Option<T>`) on fields present in *both*. It does **not** catch the reverse: a field that has been *removed* from the live spec but still lingers as a struct field in `models.rs`.

This is exactly how #235 / #236 slipped through — `storageSize` was dropped from the Postgres request schemas (`PostgresServicePostRequest`, `PostgresServicePatchRequest`, `BasePostgresService`, `PostgresServiceListItem`) upstream, the vendored snapshot was even refreshed to match, but `models.rs` kept the field and every check stayed green. A model that is a *superset* of the spec passes all current field checks.

## Root cause

Every field check iterates over **spec** properties only:

- `scripts/check-openapi-drift.py`
  - `check_missing_fields` (~L415): `for prop_name in props: if prop_name not in fields → missing`
  - `check_field_optionality` (~L331): same loop; a field in the model but not the spec hits `if prop_name not in fields: continue` (silently skipped)
- `crates/clickhouse-cloud-api/tests/spec_coverage_test.rs`
  - `struct_fields_cover_every_spec_property` and `field_optionality_matches_spec` — same spec → struct direction

### Telling asymmetry

For **operations** the check is already bidirectional: `assert_client_operation_coverage` (`spec_coverage_test.rs:80`) reports both `missing` *and* `extras` ("Extra client methods (not in spec)"). The same "extras" concept was simply never built for fields.

## Proposed fix

Add an "extra struct fields" check mirroring the existing "extra client methods" one:

1. In `check-openapi-drift.py`: a reverse pass that, for each struct mapped to a spec schema, flags model fields with no corresponding spec property. Surface it in the drift report (new "Extra struct fields (not in spec)" section + summary row).
2. In `spec_coverage_test.rs`: a `struct_fields_have_no_extras_vs_spec` test asserting no model field is absent from its spec schema.

### Wrinkle to handle: allowlist

Some struct fields legitimately have no spec property, e.g.:
- Response-only / computed fields we add intentionally.
- Structs in `models.rs` that don't map 1:1 to a spec schema (the checks already `continue` when the schema isn't found — keep that).

Introduce an allowlist constant analogous to `OPTIONALITY_EXEMPTIONS` (e.g. `EXTRA_FIELD_EXEMPTIONS` of `("RustStructName", "fieldName")`) so intentional additions don't false-positive. The drift script should parse the same list (it already does this for exemptions via `parse_optionality_exemptions`) so the report and the test stay in sync.

## Acceptance criteria

- Removing a field from a spec schema while leaving it in `models.rs` fails `spec_coverage_test.rs` and shows up in the drift report.
- Intentional code-only fields are covered by a documented allowlist that fails when it goes stale.
- Reusing #236 as a regression reference: had this existed, the stale `storageSize` fields would have been flagged.

sdairs:
Closes #237.

## Problem

OpenAPI drift detection was **one-directional for fields** (spec → code). It caught fields *added* to the spec that are missing from `models.rs` and optionality flips, but not the reverse: a field *removed* from the live spec yet still present as a struct field. A model that is a superset of the spec passed every field check — exactly how the stale `storageSize` fields in #235/#236 slipped through.

## Change

Adds the code→spec mirror of the existing missing-field check, matching the missing/extra split already used for client methods (`assert_client_operation_coverage`), **and remediates the findings it surfaces** so the model matches the vendored snapshot.

### Detector

- **`spec_coverage_test.rs`**: `assert_no_extra_struct_fields` + `struct_fields_have_no_extras_vs_spec` (snapshot) and `struct_fields_have_no_extras_vs_live_spec` (live, `#[ignore]`). Adds an `EXTRA_FIELD_EXEMPTIONS` allowlist — analogous to `OPTIONALITY_EXEMPTIONS` / `NON_OPENAPI_CLIENT_METHODS` — with stale-entry detection. Schemas with no/empty `properties` are skipped so composition/marker schemas don't flag every field.
- **`check-openapi-drift.py`**: `check_extra_fields` reverse pass + `parse_extra_field_exemptions` (reads the same Rust constant), surfaced as a new "Extra Struct Fields" report section and summary row.
- **`AGENTS.md`**: documents the now-bidirectional field coverage and the `EXTRA_FIELD_EXEMPTIONS` allowlist.

### Remediation

The detector surfaced 5 stale fields against the snapshot; all are **fixed, not exempted**, so `EXTRA_FIELD_EXEMPTIONS` stays empty.

| Schema | Field(s) removed | Spec allows |
|--------|------------------|-------------|
| `PostgresServicePatchRequest` | `name`, `provider`, `region`, `postgresVersion` | `size`, `haType`, `tags` |
| `ScimEnterpriseManager` | `$ref` | `displayName`, `value` |

- `PostgresServicePatchRequest` is live CLI surface — same class of stale field as the `storageSize` cleanup in #235/#236. The `postgres update` command loses the matching `--name`/`--region`/`--provider`/`--pg-version` flags (keeps `--size`/`--ha-type`/`--add-tag`/`--remove-tag`); `PostgresUpdateOptions`, the `postgres_update` handler, the `main.rs` match arm, the parse/write-command tests, and the README update example are updated accordingly.
- `ScimEnterpriseManager.$ref` was a code-only SCIM attribute absent from the ClickHouse spec, with no usages.
- Library `update_postgres_service` mock test rewritten to patch `size` instead of `name`.

The live-spec-only findings (`seekSnapshot` on the two PubSub sources) are **out of scope** — they aren't in the vendored snapshot this PR aligns to; they'll be picked up by the live drift run / snapshot refresh.

## Verification

- `struct_fields_have_no_extras_vs_spec` now **passes** — all 9 snapshot spec-coverage tests green; the 8 live-spec tests stay `#[ignore]`.
- Allowlist proven both ways: a real entry is suppressed (`NOTE: … exempted`), a bogus entry fails as stale.
- Drift script renders the new section/summary row; script and test agree on the same findings.
- `cargo build`, `cargo clippy --all-targets`, and the full `clickhousectl` + `clickhouse-cloud-api` test suites pass; all integration targets compile.

---

> [!NOTE]
> **Medium Risk**
> Removes `postgres update` flags and patch request fields users may still rely on; changes are spec-aligned but are a breaking CLI/API-model surface for Postgres metadata updates.
> 
> **Overview**
> Adds **bidirectional OpenAPI field coverage** so drift detection catches Rust struct fields that no longer exist in the spec, not only missing spec properties. `spec_coverage_test.rs` gains `assert_no_extra_struct_fields` with snapshot and live-spec tests plus an empty `EXTRA_FIELD_EXEMPTIONS` allowlist (stale-entry checks). `check-openapi-drift.py` mirrors this with `check_extra_fields`, exemption parsing, and an **Extra Struct Fields** report section.
> 
> **Remediation** removes five stale fields surfaced by the new check: `PostgresServicePatchRequest` drops `name`, `provider`, `region`, and `postgresVersion`; `ScimEnterpriseManager` drops `$ref`. The **`cloud postgres update`** CLI is narrowed to match the patch schema—`--name`, `--region`, `--provider`, and `--pg-version` are removed; updates use `--size`, `--ha-type`, and tag flags. README, library mock tests, and CLI parse/write tests follow the same shape.
> 
> `AGENTS.md` documents bidirectional field coverage and extra-field exemptions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 676d5b98c960e0eb2e107d71b7be023ed91f901f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -82,14 +82,23 @@ In `models.rs`, required non-nullable fields use bare types (`T`), optional/null
 **Tooling:**
 
 - `scripts/resolve-field-requirements.py` — resolves required/optional for every schema field, outputs a JSON manifest. Handles both conventions + PATCH + nullable.
-- `scripts/check-openapi-drift.py` — daily CI drift check; reports missing/extra methods, missing schemas/fields, and field-level optionality mismatches against the live spec.
+- `scripts/check-openapi-drift.py` — daily CI drift check; reports missing/extra methods, missing/extra struct fields, missing schemas, and field-level optionality mismatches against the live spec.
 - `spec_coverage_test.rs::field_optionality_matches_spec` — asserts every field's `Option<T>` vs `T` matches the snapshot.
+
+Field coverage is **bidirectional**, mirroring the missing/extra split used for client methods:
+
+- `struct_fields_cover_every_spec_property` (spec → code) — every spec property has a matching struct field; catches fields *added* to the spec.
+- `struct_fields_have_no_extras_vs_spec` (code → spec) — every struct field maps to a spec property; catches fields *removed* from the spec but left behind in `models.rs` (a superset model would otherwise pass every other field check). Schemas with no/empty `properties` are skipped, so composition/marker schemas don't flag every field. The drift script's "Extra Struct Fields" section reports the same finding.
 
 Field optionality is maintained by hand. When the drift check or test flags a mismatch, edit `models.rs` directly to flip the field (`T` ↔ `Option<T>`) and adjust the `#[serde(skip_serializing_if = "Option::is_none")]` attribute to match.
 
 **Optionality exemptions:**
 
 Sometimes the spec marks a field as required but the API rejects empty/default values, meaning the field is effectively optional. These fields are kept as `Option<T>` in `models.rs` and listed in the `OPTIONALITY_EXEMPTIONS` constant in `spec_coverage_test.rs`. The test logs each exemption and fails if any become stale (spec was fixed upstream). When adding a new exemption, add a `("RustStructName", "specFieldName")` entry with a comment explaining the API behavior.
+
+**Extra-field exemptions:**
+
+A struct field that intentionally has no spec property (a code-only/computed field, or a standard attribute the upstream spec omits) goes in the `EXTRA_FIELD_EXEMPTIONS` constant in `spec_coverage_test.rs`, analogous to `OPTIONALITY_EXEMPTIONS` and to `NON_OPENAPI_CLIENT_METHODS` for methods. `struct_fields_have_no_extras_vs_spec` fails on a stale entry (one that no longer corresponds to an actual extra field), and `check-openapi-drift.py` parses the same list so the report and test stay in sync. The list is empty by default — only add an entry for a *deliberate* addition, not to silence a field that should be removed.
 
 ##### Deprecated field hiding
 
--- a/README.md
+++ b/README.md
@@ -499,7 +499,6 @@ clickhousectl cloud postgres create \
 
 # Update metadata (all flags optional)
 clickhousectl cloud postgres update <pg-id> \
-  --name renamed \
   --size m7i.4xlarge \
   --add-tag env=prod --remove-tag legacy
 
--- a/crates/clickhouse-cloud-api/src/models.rs
+++ b/crates/clickhouse-cloud-api/src/models.rs
@@ -10521,14 +10521,6 @@ pub struct PostgresServicePatchRequest {
     #[serde(rename = "haType", skip_serializing_if = "Option::is_none", default)]
     pub ha_type: Option<PgHaType>,
     #[serde(skip_serializing_if = "Option::is_none", default)]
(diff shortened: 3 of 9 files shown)


iskakaushik (approved):
LGTM. Nice symmetric completion of the field-drift detector (code→spec mirror of the missing-field check), with stale-entry detection on the empty `EXTRA_FIELD_EXEMPTIONS` allowlist and the drift script kept in sync. The `postgres update` flag removals are correct — those fields aren't in the patch schema, so they were no-ops on the wire. README, lib mock test, and CLI shape tests all updated in lockstep. Full CI incl. cloud integration is green.

PR #272, opened 2026-06-25
Make update notice prominent on every non-JSON command (#269)

### LINKED ISSUE #269: Make update notice more prominent: show on every non-JSON output, drop 24h check gate
## Problem

Users aren't updating clickhousectl. The update notice isn't surfaced enough.

Today the notice is shown in very few places:

- It only prints on `--help` and `--version` (the clap error branch in `crates/clickhousectl/src/main.rs`), via `update::print_cached_update_notice()`.
- It does **not** print during normal command execution at all — see the comment at `main.rs`: *"The notice itself is only shown on --help (above), not during normal execution."*
- A 24-hour cache gate (`CHECK_INTERVAL_SECS = 24 * 60 * 60` + `refresh_update_cache()` in `crates/clickhousectl/src/update.rs`) means we only refresh the cached "latest version" once per day.

## What we want

1. **Kill the once-every-24-hour check gate.** Remove the `CHECK_INTERVAL_SECS` throttle so we surface the available-update state on every invocation.

2. **Show the notice on every command that does not pass `--json`.** Every single non-`--json` output should display the update notice when a newer version is available — not just `--help`/`--version`.

3. **`--help` must always show it** (when an update is available).

4. **`clickhousectl --version` must always show it** (when an update is available).

5. **Always put the notice at the very end** of the output.

6. **New message text** (single line, replacing the current two-line message):

   > There is a new version of clickhousectl. Update with `clickhousectl update`.

## Notes / open questions

- `--json` output must remain clean (no notice), so machine/agent consumers are unaffected.
- Need to make sure the notice goes to **stderr** (as it does today) so it never corrupts piped stdout, and that it always lands after the command's own output.
- Consider keeping the background cache refresh (so we don't make a blocking network call on every command) but drop the staleness throttle, or refresh more aggressively. Implementation detail to settle in the PR.

## Acceptance criteria

- [ ] 24h check interval removed.
- [ ] Update notice shown at the end of every non-`--json` command when an update is available.
- [ ] Update notice shown after `--help` and `--version`.
- [ ] Notice uses the new single-line message.
- [ ] `--json` output unaffected.
- [ ] Tests updated/added.

sdairs:
Closes #269

## Problem

Users weren't updating clickhousectl because the update notice was barely surfaced — it only printed on `--help`/`--version`, never during normal command execution.

## What changed

- **Show the notice on every non-`--json` command.** The cached update notice now prints to **stderr at the very end** of every command that produces human-readable output. `--json` and detected coding agents continue to suppress it, so machine/agent consumers are unaffected.
- **New single-line message:**
  > There is a new version of clickhousectl. Update with `clickhousectl update`.
- **`--version` always checks the network**, refreshes the cache, and resets the staleness timer, then prints the notice from the fresh cache.
- **`update --check`** now also writes the cache and resets the timer on every run.
- **Successful `update`** clears the cache so the notice disappears immediately.
- **Kept the 24h background-refresh gate** for normal commands (one network call per day). On network failure we now preserve any previously-cached "update available" version instead of clobbering it with the current version, so a known update is never hidden.

## Tests

- `cache_is_stale` boundary cases (`update.rs`).
- `command_json_flag` mapping + `--json`/`update` suppression (`main.rs`).
- All 353 tests pass; clippy clean.

## Manual verification (live network, current latest `0.3.1`)

| Scenario | Result |
|---|---|
| `--version`, seeded stale cache | Hit network, cache rewritten to `0.3.1`, timer reset |
| `--version` on a temp `v0.0.1` build | Printed version, then the notice |
| `update --check`, seeded wrong cache | `Update available: v0.0.1 → v0.3.1`; cache rewritten, timer reset |
| Normal human command, newer cache | Notice prints at the very end after output |
| Same command with `--json` / under agent detection | Notice suppressed |

### DIFF
--- a/README.md
+++ b/README.md
@@ -833,7 +833,7 @@ clickhousectl update
 clickhousectl update --check
 ```
 
-The CLI also checks for updates in the background (at most once per 24 hours) and displays a notice when a newer version is available.
+The CLI checks for updates in the background (at most once per 24 hours) and caches the result. When a newer version is available, a one-line notice is printed to stderr at the end of every command that produces human-readable output. JSON output (`--json` or a detected coding agent) is never affected, so machine consumers stay clean. Running `clickhousectl update` clears the cached notice.
 
 ## Cloud integration testing
 
--- a/crates/clickhousectl/src/main.rs
+++ b/crates/clickhousectl/src/main.rs
@@ -34,24 +34,39 @@ async fn main() {
     let cli = match Cli::try_parse() {
         Ok(cli) => cli,
         Err(e) => {
-            // For --help and --version, show the update notice after the output
-            if e.kind() == ErrorKind::DisplayHelp || e.kind() == ErrorKind::DisplayVersion {
-                e.print().expect("failed to print output");
-                update::print_cached_update_notice();
-                std::process::exit(0);
+            match e.kind() {
+                // --version always hits the network to refresh the cache + timer,
+                // then prints the notice from the freshly-updated cache.
+                ErrorKind::DisplayVersion => {
+                    e.print().expect("failed to print output");
+                    update::force_refresh_update_cache().await;
+                    update::print_cached_update_notice();
+                    std::process::exit(0);
+                }
+                // --help shows the notice from cache (no blocking network call).
+                ErrorKind::DisplayHelp => {
+                    e.print().expect("failed to print output");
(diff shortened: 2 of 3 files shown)


sdairs:
bugbot review

iskakaushik (approved):
LGTM

PR #274, opened 2026-06-25
Bare `local server start` bootstraps `latest` when nothing is set up

### LINKED ISSUE #264: [enhancement]: bare `local server start` should auto-install `latest` when nothing is installed
## Summary

Running `clickhousectl local server start` with no installed version, no default set, and no `--version` flag should bootstrap automatically: install `latest` and start the server. Today it fails instead.

## Current behaviour

`start_server` (`crates/clickhousectl/src/local/mod.rs:294-300`) only auto-installs when `--version` is passed:

```rust
let version = if let Some(spec_str) = &version_spec {
    // server start --version <spec> → resolves + installs if missing
    let spec = version_manager::parse_version_spec(spec_str)?;
    let platform = version_manager::platform::Platform::detect()?;
    version_manager::install::ensure_installed_local_first(&spec, &platform).await?
} else {
    // bare `server start` → just reads the default, no install
    version_manager::get_default_version()?
};
```

With nothing installed, `get_default_version()` returns `Error::NoDefaultVersion`, which prints:

```
No default version set. Run: clickhousectl local use <version>
```

and exits with code 1. No download happens. The only zero-to-running path right now is `local server start --version latest`, or the documented three-step `local install stable && local use stable && local server start`.

## Proposed behaviour

When no `--version` is given **and** no default version is set (`get_default_version()` → `NoDefaultVersion`), fall back to installing `latest` and starting with it, rather than erroring. Effectively, the bare-start `else` branch should treat a missing default as "install `latest`" instead of a hard error.

Suggested shape:
- If a default exists → use it (unchanged).
- If no default is set → resolve `VersionSpec::Latest`, `ensure_installed_local_first`, set it as the default (matching `install`'s first-install-sets-default behaviour), then start. Emit a note to stderr that `latest` was installed so the auto-download isn't silent.
- A `VersionNotFound` default (file points at a removed binary) should remain an error — only the *no default at all* case bootstraps.

## Notes 
- Do not set the default version to the version that was installed. This is deliberate; it means the user will keep getting the latest version if they dont pin.
- Doc update to `README.md` to reflect that a bare `server start` now bootstraps.
- Update all docs that show using a tag like "stable" to use latest

### LINKED ISSUE #280: local remove does not clear master sidecar record
## Summary

PR #274 introduced an etag-based master sidecar record (in `crates/clickhousectl/src/version_manager/master.rs`) that tracks which `master` build is installed so it can be reused if the upstream HEAD hasn't changed.

When a user runs `local remove <version>` on the version that the master sidecar record points at, the version directory is deleted but the sidecar record is **not** cleared. This leaves a stale record: the binary no longer exists, so `reuse_if_unchanged`'s `binary_exists` check (`master.rs:166-168`) returns `false` and reuse is correctly refused (a re-download is forced) — but the sidecar silently drifts and is never cleaned up.

## Location

- `crates/clickhousectl/src/version_manager/install.rs:183` — `clear_record_for_version` is called when a non-master install *overwrites* a version dir, but `local remove` (`crates/clickhousectl/src/local/mod.rs` `remove()` handler, ~line 220) deletes a version dir without clearing the master record.

## Expected behavior

`local remove <version>` should clear the master sidecar record when the removed version matches the record's tracked version, for symmetry with the `install.rs` overwrite path.

## Impact

Low. This is not a correctness bug (reuse is safely refused when the binary is absent), but it leaves stale state in `.clickhouse/` and means a subsequent `local server start` on master will always re-download instead of potentially reusing.

## Discovered during

Code review of PR #274 (Bare `local server start` bootstraps `latest` when nothing is set up).

sdairs:
Closes #264.

## Summary

Running `clickhousectl local server start` with no installed version, no default set, and no `--version` flag now bootstraps automatically: it installs `latest` and starts with it, instead of failing with `NoDefaultVersion` and exit code 1.

- If a default exists → use it (unchanged).
- If no default is set (`NoDefaultVersion`) → resolve `latest`, install if missing, and start. We deliberately **do not** set it as the default, so unpinned users keep tracking `latest` on later starts. A note is printed to stderr so the auto-download isn't silent.
- A `VersionNotFound` default (file points at a removed binary) stays an error — only the *no default at all* case bootstraps.

## Changes

- **`start_server` (`src/local/mod.rs`)** — the bare-start branch matches on `get_default_version()` and handles `NoDefaultVersion` by bootstrapping `latest`. Message: `No version specified and no default set; installing latest...`.
- **`ensure_installed` (`src/version_manager/install.rs`)** — bug fix required for the feature. For `latest`/master builds the exact version is only known after download, so `install_resolved` returned `VersionAlreadyInstalled`, breaking `ensure_installed`'s "return existing, don't error" contract. Because bare start never sets a default, *every* repeat start hits this path and would fail. Now maps `VersionAlreadyInstalled` → `Ok(version)`. This also fixes the pre-existing `server start --version latest`-on-repeat failure.
- **Docs / help text** — README server section and `server start` help document the bootstrap; the "typical workflow" help text drops the three-step `install stable && use stable && start` dance in favour of a bare `server start`; `stable`-as-tag examples switched to `latest` (the install keyword catalog keeps `stable` documented).

## Skip re-downloading master when unchanged (etag change detection)

Because bare start never sets a default, every repeat re-entered the `latest` path and re-downloaded the full ~153MB master binary (then re-ran `clickhouse --version`), even when master hadn't moved. This PR now short-circuits that.

Master exposes no readable version (the `--version` string is shared across many commits, sibling metadata files 403), but `builds.clickhouse.com` returns a stable, content-keyed `etag` on a HEAD request (~50ms, no body). We use it as a change-detection key:

- **New `src/version_manager/master.rs`** — a per-platform sidecar (`~/.clickhouse/versions/.master-builds.json`) recording each installed master build's `etag` + version; a best-effort `head_info()` HEAD helper; a pure `should_reuse`/`reuse_if_unchanged` decision; and a `record()` writer.
- **`install_resolved`** — for the master source, HEAD-and-compare before downloading. Etag match + binary present → reuse the installed build, skipping the download *and* version detection. Otherwise download and re-record. Handles "etag recorded but binary missing" by falling through to download. Applies to both `install latest` and `server start --version latest` / bare start (shared path).

This also fixes a **latent bug**: because a master build's detected version string is shared across commits, the old code would re-download a moved master, detect the same version, hit `VersionAlreadyInstalled`, and discard the fresh binary — paying the full download cost yet keeping the stale build. Reuse/overwrite now key on the etag, so a changed master correctly overwrites in place.

## Verification

Manual end-to-end in an isolated `HOME` (no installs, no default):

- Bare `server start` → message prints, `latest` downloads, server starts, exit 0, **no `default` file created**. ✅
- Second bare `server start` → reuses the installed build, starts, exit 0 (previously errored). ✅
- Default pointing at a removed binary → `Error: Version <x> not found`, exit 1. ✅

Etag change detection, against the live CDN with the built binary:

- First `install latest` → downloads master, writes the sidecar with the build's `etag`. **~12.4s**. ✅
- Second `install latest` → `latest is up to date (master build unchanged); using <v>`, single HEAD, no download. **~0.14s**. ✅
- 7 new unit tests cover the reuse decision (etag match/mismatch, missing binary, no record, failed HEAD) and sidecar round-trip.

- `cargo build`, `cargo clippy` (clean), `cargo test` (all pass). ✅

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ clickhousectl local client --host remote-host --port 9000  # Connect to a specif
 
 Start and manage ClickHouse server instances. Each server gets its own isolated data directory at `.clickhouse/servers/<name>/data/`.
 
-A bare `clickhousectl local server start` bootstraps from zero: if no version is installed and no default is set, it installs `latest` and starts with it (it does not set a default, so you keep tracking `latest` on subsequent starts). Pin a version with `--version`, or set a default with `local use`, to opt out.
+A bare `clickhousectl local server start` bootstraps from zero: if no version is installed and no default is set, it installs `latest` and starts with it (it does not set a default, so you keep tracking `latest` on subsequent starts). Pin a version with `--version`, or set a default with `local use`, to opt out. Because `latest` tracks the rolling master build, repeat `latest` installs/starts do a cheap `HEAD` against `builds.clickhouse.com` and skip the ~150 MB re-download when master hasn't changed (the build's `etag` is cached in `~/.clickhouse/versions/.master-builds.json`).
 
 ```bash
 # Start a server (runs in background by default)
--- a/README.md
+++ b/README.md
@@ -92,11 +92,11 @@ clickhousectl local list                    # Installed versions
 clickhousectl local list --remote           # Available for download
 
 # Manage default version
-clickhousectl local use stable              # Latest stable (installs if needed)
+clickhousectl local use latest              # Latest master build (installs if needed)
 clickhousectl local use lts                 # Latest LTS (installs if needed)
 clickhousectl local use 25.12               # Latest 25.12.x.x (installs if needed)
 clickhousectl local use 25.12.5.44          # Exact version
-clickhousectl local use stable --no-global  # Set default but don't touch ~/.local/bin/clickhouse
+clickhousectl local use latest --no-global  # Set default but don't touch ~/.local/bin/clickhouse
 clickhousectl local which                   # Show current default
 
 # Remove a version
@@ -159,11 +159,13 @@ clickhousectl local client --host remote-host --port 9000  # Connect to a specif
 
 Start and manage ClickHouse server instances. Each server gets its own isolated data directory at `.clickhouse/servers/<name>/data/`.
 
+A bare `clickhousectl local server start` bootstraps from zero: if no version is installed and no default is set, it installs `latest` and starts with it (it does not set a default, so you keep tracking `latest` on subsequent starts). Pin a version with `--version`, or set a default with `local use`, to opt out.
+
 ```bash
 # Start a server (runs in background by default)
-clickhousectl local server start                          # Named "default"
+clickhousectl local server start                          # Named "default" (installs latest if nothing is set up yet)
 clickhousectl local server start --name dev               # Named "dev"
-clickhousectl local server start --version stable         # Use a specific version (installs if needed, doesn't change default)
+clickhousectl local server start --version latest         # Use a specific version (installs if needed, doesn't change default)
 clickhousectl local server start --foreground             # Run in foreground (-F / --fg)
 clickhousectl local server start --http-port 8124 --tcp-port 9001  # Explicit ports
 clickhousectl local server start --config-file analytics  # Apply a custom config (see "Custom config files" below)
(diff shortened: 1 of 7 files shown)


iskakaushik (approved):
LGTM

PR #281, opened 2026-07-05
Move OAuth tokens to global ~/.clickhouse/tokens.json (fixes #277)

### LINKED ISSUE #277: cloud auth looks for .clickhouse/tokens.json in cwd
### What happened?

chctl writes and reads `.clickhouse/tokens.json` from whatever directory you happen to be in when you `clickhousectl cloud auth login` or `clickhousectl cloud auth status`

ideally, this would be stored in `$XDG_CONFIG_HOME`

if it matters, I downloaded and unpacked the clickhouse binary from builds.clickhouse.com

### Steps to reproduce

```shell
clickhousectl cloud auth login
cd /
clickhousectl cloud auth status
```

### clickhousectl version

0.3.1

### Installation method

Other (describe in "What happened?")

### Operating system

macOS (Apple Silicon)

### OS version

_No response_

### Relevant logs or output

```shell

```

sdairs:
OAuth `cloud auth login` / `auth status` read and wrote `<cwd>/.clickhouse/tokens.json`, so logging in from one directory left you logged out everywhere else (#277). Tokens now live at `~/.clickhouse/tokens.json` alongside versions/configs via the existing `paths::base_dir()` helper. `load_tokens()` does a one-time best-effort migration from a legacy cwd-based file (copy global, delete legacy); `clear_tokens()` removes both so logout in an old project dir doesn't leave a file that re-migrates. `credentials.json` (API keys) stays project-local — org-scoped per-project — so the asymmetry is deliberate.

Behavioral note: with a single global store, logging into staging now overwrites a prod login (the token records its `api_url`, so it's self-consistent, but juggling environments via separate dirs no longer works). A per-host keyed store could be a follow-up.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,7 +16,7 @@ The user-facing CLI surface. Contains all logic for local commands, wraps `click
 
 Use `--help` to learn the current command surface.
 
-Project-local data lives in `.clickhouse/`. Globally installed ClickHouse binaries live in `~/.clickhouse/`.
+Project-local data lives in `.clickhouse/`. Globally installed ClickHouse binaries live in `~/.clickhouse/`. OAuth tokens (`~/.clickhouse/tokens.json`) are the exception — they're global user identity, not project-scoped.
 
 The CLI does not need to have 100% coverage of endpoints exposed by the API library: be intentional about what is exposed to users.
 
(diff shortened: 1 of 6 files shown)


iskakaushik (approved):
LGTM

PR #288, opened 2026-07-07
Resolve OpenAPI drift: object storage ingestion controls, pg autovacuum config, horizontal autoscaling fields (#287)

sdairs:
Resolves #287 (library-only, per the drift workflow in CLAUDE.md — CLI flag exposure tracked separately).

## What

The daily drift check found 21 spec properties with no matching struct field, so API response data for them was silently dropped. This adds all 21 and refreshes the vendored snapshot:

| Struct | New fields | Modelled as |
|--------|-----------|-------------|
| `ClickPipe{,Post,Patch}ObjectStorageSource` | `skipInitialLoad`, `startAfter` | `Option<bool>` / `Option<String>` (nullable in spec) |
| `ScalingScheduleEntry{,Request}` | `numReplicas`, `replicaMemoryGb` | `Option<i64>` / `Option<f64>` (optional via `required[]`) |
| `ServicePostRequest` | `minReplicas`, `maxReplicas`, `replicaMemoryGb` | `Option<f64>` matching the `numReplicas` sibling |
| `PgConfig` | 8 `autovacuum_*` settings | `Option<serde_json::Value>` (string-or-number union like every other pgConfig field) |

## Optionality exemptions

- `ServicePostRequest.{minReplicas,maxReplicas,replicaMemoryGb}`: the description heuristic infers required, but the horizontal-autoscaling trio is mutually exclusive with `numReplicas`/`min/maxReplicaMemoryGb`/`min/maxTotalMemoryGb`, so it must be omittable — same rationale as the existing `numReplicas` entry.
- `PgConfig.autovacuum_*`: covered by the existing partial-update rationale from #163.

## Tests

- Serde coverage in `models_test.rs` for all new fields (camelCase wire names, omitted-when-`None`, string-or-int pgConfig values).
- The Postgres config integration round-trip now patches/verifies/resets `autovacuum_max_workers` alongside `max_connections`.
- No integration coverage for the scaling fields (`numReplicas`/`replicaMemoryGb`/horizontal autoscaling require org-level feature enablement) or `skipInitialLoad`/`startAfter` (they change ClickPipe ingestion behaviour); those have serde coverage only.
- Exhaustive struct literals in `commands.rs`, `support.rs`, and `integration_test.rs` updated with behaviour-preserving defaults.

## Verification

- `python3 scripts/check-openapi-drift.py --dry-run` → **"No drift. Library fully covers the live spec."**
- `cargo test -p clickhouse-cloud-api` / `cargo test -p clickhousectl` — all green
[...]

PR #273, opened 2026-06-25
Resolve OpenAPI drift: slow query patterns, Postgres metrics, RPE update (#224)

sdairs:
Resolves #224.

Issue #224 was auto-filed by the daily drift check against an older live spec. Since then the spec drifted further (and two of #224's items — the `CustomPrivateDnsMapping` type and the `customPrivateDnsMappings` fields — were already fixed in a prior PR). This PR clears the **full current drift** so the daily check goes green and #224 closes cleanly. Library-only, per CLAUDE.md; the only CLI changes are construction sites that wouldn't otherwise compile.

## clickhouse-cloud-api

**5 client methods** (`client.rs`)
- `slow_query_patterns_get_list`, `slow_query_pattern_get` — Postgres slow query patterns (Beta)
- `postgres_instance_metrics_get` — Postgres metrics (Beta)
- `service_clickhouse_setting_delete` — delete a ClickHouse setting (Beta)
- `click_pipe_reverse_private_endpoint_update` — PATCH a reverse private endpoint

**9 model types** (`models.rs`)
- `PostgresSlowQueryPattern`, `PostgresQueryExecution`, `PostgresSlowQueryPatternDetail`
- `PostgresMetrics`, `PostgresMetric`, `PostgresMetricSeries`, `PostgresMetricDataPoint`
- `License`, `UpdateReversePrivateEndpoint`

**12 added struct fields** — `exactlyOnce`, `skipCertVerification`, `disableTls` across ClickPipe source structs; `name` on `PostgresServicePatchRequest`. Optionality resolved per spec (`required` array + `Optional`-description heuristic).

**2 removed fields** — `seekSnapshot` dropped upstream from `ClickPipePostPubSubSource` / `ClickPipePubSubSource`.

**1 optionality fix** — `ClickPipeMutateDestination.roles` → `Option<Vec<String>>`.

**Metadata/tests** — 4 newly-Beta ops added to `BETA_OPERATIONS`; 3 now-stale `OPTIONALITY_EXEMPTIONS` removed (spec now agrees with the model); vendored OpenAPI snapshot refreshed.

## clickhousectl
Request builders updated for the changed struct fields with behaviour-preserving defaults. Exposing the new endpoints/fields as CLI commands/flags is intentionally a separate follow-up.

## Verification
- `cargo build` (workspace), `cargo build -p clickhouse-cloud-api --features deprecated-fields`
- `cargo test` — full suite passes, including all 8 **live-spec** coverage tests (`--ignored`)
[...]

PR #290, opened 2026-07-07
Add bidirectional enum-value drift detection (#275)

### LINKED ISSUE #275: Drift tooling doesn't compare enum variant values against the spec
## Problem

Neither `scripts/check-openapi-drift.py` nor `spec_coverage_test.rs` compares the **values** of inline enums (the spec's `enum: [...]` arrays) against the Rust enum variants. They track:

- enum/struct/type **names** (existence)
- struct **fields**
- field **optionality** (`Option<T>` vs `T`)

…but a value added to or removed from a spec enum is invisible to both the daily drift check (reports 0) and the coverage test.

## How it surfaced

In #273 (PR resolving #224), the spec dropped `seekSnapshot` and removed `snapshot` from the PubSub `seekType` enum (now `["latest","earliest","timestamp"]`). The companion `seekSnapshot` *field* removal was caught, but the stale `Snapshot` variant on `ClickPipePostPubSubSourceSeektype` / `ClickPipePubSubSourceSeektype` was left behind and serialized to `"snapshot"` — a value the API now rejects. The drift check still reported **0**. Caught only by Cursor Bugbot, fixed in #273.

## Proposed fix

Extend drift detection to enum variant values, bidirectionally (mirroring the missing/extra split already used for methods and struct fields):

- **spec → code**: every spec `enum` value has a matching Rust variant (catches *added* values)
- **code → spec**: every Rust variant (excluding the `Unknown(String)` catch-all and `#[serde(untagged)]` arms) maps to a spec value (catches *removed* values, which is the case above)

Add the check to both `check-openapi-drift.py` (report section + the daily CI run) and a new `spec_coverage_test.rs` assertion against the vendored snapshot, keeping them in lockstep like the existing field checks. Compare against the `#[serde(rename = "...")]` values, not the variant identifiers.

sdairs:
> [!IMPORTANT]
> ### Reviewer guidance: review the model fixes, skip the drift-checker methodology
>
> This PR is still intended to merge because it contains valid ClickHouse Cloud model corrections described below. However, the Rust/Python drift-checker implementation introduced here is temporary and will be **completely replaced by #297** with one shared `syn`-based analyzer.
>
> **Please review in this PR:** the enum/model corrections, regenerated values, compatibility impact, and regression coverage.
>
> **Please do not spend time reviewing in depth:** the line-oriented Rust parsing, Python regex parsing, spec-to-Rust mapping architecture, or duplicated comparison logic in `spec_coverage_test.rs` and `check-openapi-drift.py`. Review that methodology in #297 instead.
>
> Merge order: merge #290 first, then rebase/retarget #297 onto `main`.

Closes #275.

> **Stacked on #288** — this branch is based on `issue-287-openapi-drift` and reuses its refreshed spec snapshot. Merge #288 first; this PR's diff will then shrink to its own commit.

## Problem

Neither `scripts/check-openapi-drift.py` nor `spec_coverage_test.rs` compared the **values** of spec `enum: [...]` arrays against Rust enum variants, so a value added to or removed from a spec enum was invisible to both. Real incident (#273): the spec removed `"snapshot"` from the PubSub `seekType` enum, the stale `Snapshot` variant kept serializing a value the API rejects, and the drift check reported 0.

## Changes reviewers should review

The new checks surfaced real model drift, fixed in this PR:

- **29 missing values**: `europe-west2` in 8 region enums, `GCP_PSC_SERVICE_ATTACHMENT` in both reverse-private-endpoint type enums, 6 new threshold types in the 3 ClickStack alert enums, and `duration` in `ClickStackNumberFormatOutput`. These were added with their `Display` arms.
- **66 extra `PgSize` variants**: instance sizes the spec no longer lists were removed by regenerating the enum from the spec.
- **Breaking library impact**: the removed `PgSize` variants are no longer public, and the `Default` moves from the removed `c6gd.medium` to `c6gd.large`. The CLI is unaffected because `parse_pg_size` uses Serde strings.
- Regression coverage demonstrates both directions of enum-value drift, including the #273 `seekType`/`snapshot` failure mode.

## Temporary drift implementation — superseded by #297

This PR adds bidirectional enum-value coverage to both the Rust snapshot test and Python daily workflow:

- **spec → code**: catches spec values with no matching Rust variant.
- **code → spec**: catches Rust variants that serialize values no longer accepted by the spec.

Those protections and characterization cases are useful, but the implementation extends the existing duplicated parsing approach. #297 replaces the methodology with one private Rust analyzer that:

- parses Rust structurally with `syn`;
- owns mapping, exemptions, and all comparison semantics once;
- emits one typed `DriftReport` consumed by both the snapshot test and daily workflow; and
- inventories nested, parameter, composition, and unsupported enum constraints explicitly.

<details>
<summary>Historical implementation details in this PR (not the design to review)</summary>

- Inline property enums resolve through the struct field's declared type, stripped of `Option`/`Vec`/`Box`; named enum schemas resolve through `pascalize_identifier`.
- Wire values come from `#[serde(rename = "...")]`, falling back to the variant identifier.
- Untagged catch-alls are excluded by attribute rather than name.
- Non-untagged data-carrying enums, numeric enums, and `items`-level enums are excluded.
- `EXTRA_ENUM_VALUE_EXEMPTIONS` mirrors the existing extra-field exemption mechanism.

These mechanics are replaced, not evolved, by #297.

</details>

## Verification

- `cargo test --workspace`
- `cargo clippy --workspace --all-targets` with zero warnings
- `cargo build --features deprecated-fields`
- Drift-script dry run against the snapshot: zero findings, matching the Rust tests
- Regression drills for added values, removed values, non-enum mappings, and stale exemptions


iskakaushik (approved):
LGTM

iskakaushik (approved):
LGTM

iskakaushik (approved):

rndD (approved):
LGTM

rndD (approved):

Issue #296, opened 2026-07-11
Model acknowledged OpenAPI enum constraints as typed Rust values


sdairs:
## Context

Follow-up to #297, which implemented #295 on top of #290.

The shared OpenAPI drift analyzer inventories enum constraints that are not backed by concrete Rust value types. This issue tracks the 11 scalar-backed locations acknowledged when the analyzer landed.

Reviewed against `main` and the live OpenAPI spec on 2026-08-13:

- All original 11 locations remain scalar-backed and acknowledged.
- Their OpenAPI constraints and values have not changed since this issue was filed.
- The live analyzer reports 13 acknowledged unsupported constraints: these 11 plus two later UDF attachment error-response constraints. The UDF constraints have a separate inline justification and are outside this issue.
- A report with zero actionable drift does not mean these locations are resolved; the acknowledgements suppress them pending this work or a reviewed decision to retain a scalar type.

### Deprecated API-key roles

- `ApiKey.roles`
- `ApiKeyPostRequest.roles`
- `ApiKeyPatchRequest.roles`

These fields are deprecated in favor of `assignedRoles` / `assignedRoleIds` and are hidden unless the `deprecated-fields` feature is enabled. Review whether changing their exported types is worthwhile. Retaining scalar types is acceptable if the compatibility decision is documented beside their acknowledgement entries.

### Query API endpoint roles

- `ServiceQueryAPIEndpoint.roles`
- `InstanceServiceQueryApiEndpointsPostRequest.roles`

These remain `Vec<String>`. The CLI also exposes query endpoint roles through repeatable `--role` strings and currently passes them through without validating the two spec values. Request and response representations should follow the current direction-aware model policy; response values must remain tolerant of unknown future values.

### BYOC availability-zone suffixes

- `ByocInfrastructurePostRequest.availabilityZoneSuffixes`

This remains `Vec<String>` although the spec constrains values to `a` through `f`.

### Upgrade-window numeric values

- `UpgradeWindow.startHourUtc`
- `UpgradeWindow.duration`
- `UpgradeWindowPutRequest.startHourUtc`

These remain integers. Use typed numeric representations or constrained newtypes with Serde behavior matching the OpenAPI wire values. Keep request and response types distinct where required by the current model policy, and update the `UpgradeWindow` to `UpgradeWindowPutRequest` conversion.

### Slow-query list parameters

- `slow_query_patterns_get_list.sort_by`
- `slow_query_patterns_get_list.sort_order`

[...]

PR #297, opened 2026-07-11
Refit OpenAPI drift checks around shared syn analyzer

### LINKED ISSUE #295: Refit OpenAPI drift comparison flow around a shared syn-based analyzer
## Context

Follow-up to #290.

#290 adds valuable immediate protection for bidirectional string-enum value drift, but it extends the existing approach of parsing Rust source independently in two places:

- `crates/clickhouse-cloud-api/tests/spec_coverage_test.rs`
- `scripts/check-openapi-drift.py`

The comparison logic is now important infrastructure. Keeping parallel line/regex-based parsers makes Rust syntax support, Serde attribute handling, exemptions, and drift semantics harder to evolve safely. The snapshot tests and the daily live-spec workflow should consume one canonical analysis result.

## Goal

Create a shared, internal Rust analyzer that:

1. parses the library source structurally with `syn`;
2. compares the Rust API model with an OpenAPI document through pure, testable functions; and
3. returns a typed, serializable `DriftReport` used by both the snapshot coverage tests and the daily drift workflow.

This should be internal development tooling. It must not add `syn` to the public runtime dependency surface of `clickhouse-cloud-api` or `clickhousectl`.

### 1. Introduce a canonical analyzer

Add a small internal tooling crate or module with separate layers for:

- Rust source inventory
- OpenAPI inventory
- mapping/resolution
- comparison
- report serialization

Parse `client.rs`, `models.rs`, and relevant metadata using `syn`, including:

- public client methods;
- public structs, enums, and type aliases;
- struct field names and declared types;
- raw identifiers;
- `Option`/`Vec`/`Box` and nested generic types;
- Serde `rename` and container attributes;
- `cfg(feature = "deprecated-fields")` markers;
- unit variants, data-carrying variants, and variant-level Serde attributes; and
- attribute-based unknown/catch-all handling.

### 2. Return data, not assertions

The core comparison API should return a typed `DriftReport` rather than printing, panicking, or creating issues directly. It should preserve the current finding categories, including:

- missing and extra client methods;
- missing model types;
- missing and extra struct fields;
- field optionality mismatches;
- beta-status changes;
- deprecated-field changes and marker mismatches;
- missing and extra enum values;
- stale snapshot changes; and
- stale exemptions.

Each finding should carry a stable location, preferably an OpenAPI JSON pointer plus the mapped Rust item/field where applicable.

### 3. Make both consumers thin

- `spec_coverage_test.rs` should call the shared analyzer against the vendored snapshot and render assertion failures from `DriftReport`.
- The daily workflow should call the same analyzer against the live spec and use its serialized JSON to render/create the GitHub issue.
- Python may remain as the fetch/issue-orchestration layer initially, but it should no longer reimplement Rust parsing or comparison semantics.

### 4. Preserve behavior before broadening coverage

First cut over with characterization tests that reproduce the behavior landed in #290 and all existing drift checks. Once parity is established, extend OpenAPI enum discovery to inventory locations currently outside #290's scope, including:

- enums nested under `items`;
- operation-parameter enums;
- enums nested through composition where they map to a concrete Rust type; and
- numeric enum constraints, either as supported typed checks or explicit unsupported findings rather than silent omissions.

## Test plan

Add compact table-driven fixtures covering at least:

- Serde-renamed and identifier-fallback enum variants;
- `Unknown(String)` and differently named catch-all variants such as `Other(String)`;
- a real spec value named `Unknown`;
- untagged union enums that must not be treated as value enums;
- nested `Option<Vec<Box<T>>>` field types;
- raw field identifiers;
- deliberately added and removed enum values, including the #273 `seekType`/`snapshot` regression;
- missing fields and fields mapped to a non-enum Rust type;
- stale exemption detection;
- nested `items` enum locations; and
- equivalent `DriftReport` output for the vendored snapshot through the test and workflow entry points.

Keep the full vendored-snapshot coverage test as integration coverage in addition to the focused fixtures.

## Acceptance criteria

- [ ] One Rust implementation owns Rust-source parsing and code-vs-spec comparison semantics.
- [ ] Rust parsing uses `syn`, not line-oriented parsing or regexes.
- [ ] `syn` remains an internal tooling/dev dependency.
- [ ] Existing snapshot checks and daily live-spec reporting consume the same typed `DriftReport`.
- [ ] Characterization fixtures prove parity with the behavior present after #290.
- [ ] Existing exemption lists retain stale-entry detection.
- [ ] Enum locations outside #290's direct-property scope are inventoried and either checked or reported explicitly as unsupported.
- [ ] Old duplicated Rust/Python source-parsing and comparison code is removed after cutover.
- [ ] `cargo test --workspace` and `cargo clippy --workspace --all-targets` pass.
- [ ] Development documentation is updated to describe the shared analyzer and how to add a new drift check.

## Non-goals

- Remediating newly discovered OpenAPI drift in the same PR as the analyzer refit.
- Adding or changing CLI commands.
- Changing the public runtime API solely to accommodate the tooling refactor.

Model/API remediation surfaced by the new analyzer should be reviewed in separate follow-up PRs.

sdairs:
Closes #295.

> **Stacked on #290** — this PR targets `issue-275-enum-value-drift` so its diff contains only the analyzer refit. After #290 merges, rebase this branch onto `main` and retarget the PR.

Follow-up model/API remediation for the 11 acknowledged scalar-backed enum constraints is tracked in #296.

## Problem

The snapshot coverage test and daily Python workflow independently parsed Rust source and implemented the same code-vs-spec comparisons. That duplicated syntax handling, Serde interpretation, exemptions, and drift semantics in two fragile line/regex-based implementations.

## What changed

- Added a private, non-published `clickhouse-openapi-analyzer` workspace crate.
- Parse `client.rs`, `models.rs`, and `meta.rs` structurally with `syn`.
- Return a deterministic, versioned, serializable `DriftReport` with stable OpenAPI JSON pointers and Rust item locations.
- Preserve operation, model, field, optionality, beta, deprecation, enum-value, snapshot, and stale-exemption findings.
- Inventory enums under array items, operation parameters, and composition nodes.
- Report numeric or scalar-backed enum constraints explicitly; acknowledge the 11 current snapshot locations while making new or stale acknowledgements actionable.
- Reduce `spec_coverage_test.rs` and `check-openapi-drift.py` to thin consumers of the same report.
- Keep Python responsible only for fetching, issue rendering, and GitHub orchestration.
- Extend CI, add Rust/Python fixtures and entry-point parity coverage, and update development documentation.

## Impact

There are no CLI behavior changes or public runtime API changes. `syn` remains isolated to the private analyzer/dev dependency path. The vendored snapshot produces zero actionable findings and 11 visible acknowledged diagnostics.

## Verification

- `cargo build --workspace`
- `cargo test --workspace`
- Confirmed the analyzer crate is absent from both published crates' normal dependency graphs


sdairs:
## Code review findings

Reviewed the shared-analyzer refit. Verified locally: analyzer tests + entrypoint parity, `spec_coverage_test` (0 actionable drift), clippy `-D warnings`, and the Python unittest suite all pass. Three non-blocking findings:

**1. `MissingReferencedModelType` is redundant with `MissingModelType` and double-counts in the issue summary** — `compare.rs:67-113`
Loop 1 emits `MissingModelType` for every defined schema lacking a Rust model. Loop 2 emits `MissingReferencedModelType` for any *referenced* schema that is defined but lacks a model. Since loop 1 already iterates **all** defined schemas, every `MissingReferencedModelType` is necessarily also a `MissingModelType` — the referenced branch's only unique value is its `MissingSchemaDefinition` case (referenced-but-undefined). Result: two findings per missing type, and the Python summary inflates the count via `total('missing_model_type', 'missing_referenced_model_type')` (`check-openapi-drift.py:190`). Cosmetic on the clean snapshot, but the "Missing model types" count will be wrong during real drift. Suggest dropping the `MissingReferencedModelType` branch (keep only `MissingSchemaDefinition` in loop 2), or skipping it when the schema is also in `spec.schemas`.

**2. Nested-object property enums are attributed to the top-level property** — `openapi.rs:321-328`
`enum_context` sets `property = path[4]` for anything under `.../properties/<x>/...`. For a nested shape like `Widget.properties.foo.properties.bar` with an enum on `bar`, the constraint is attributed to field `foo`, and `map_enum` resolves `foo`'s type — likely producing a spurious `UnsupportedEnumConstraint` ("not an enum") or a wrong mapping if `foo` happens to be an enum. Not exercised by the current snapshot, but since this is now the canonical checker, a comment noting the "first `properties` segment wins" limitation (or handling the deepest segment) would prevent a confusing future false-positive.

**3. Daily `openapi-drift.yml` has no Rust build cache (perf only)** — `.github/workflows/openapi-drift.yml`
The workflow now cold-compiles the analyzer via `cargo run` every day with no `Swatinem/rust-cache` step (contrast `test-cloud-api.yml`). Correctness is fine; it's just wasted CI minutes. Optional.

Only #1 has a behavioral consequence (a double-count during actual drift), and it's a small localized fix. Nice single-sourcing overall — the executable-vs-library parity test plus Python-only-consumes-JSON contract match what AGENTS.md promises.

sdairs:
Review findings for head `aa8b14f`:

1. **[P2] Mirror Serde's distinct field and variant rename rules.** In [`rust_inventory.rs`](https://github.com/ClickHouse/clickhousectl/blob/aa8b14f31d50b7481bf26cddbb3d948c6f00e76e/crates/clickhouse-openapi-analyzer/src/rust_inventory.rs#L386-L403), `apply_rename_rule` is used for both struct fields and enum variants, but Serde applies `rename_all` differently to each. It also handles acronym-heavy variants differently: under `snake_case`, Serde serializes `HTTPServer` as `h_t_t_p_server`, while the analyzer records `httpserver`. This can produce false missing/extra enum findings. Please split field and variant transformations and mirror `serde_derive`'s rules.

2. **[P2] Restrict enum discovery to schema-bearing OpenAPI nodes.** In [`openapi.rs`](https://github.com/ClickHouse/clickhousectl/blob/aa8b14f31d50b7481bf26cddbb3d948c6f00e76e/crates/clickhouse-openapi-analyzer/src/openapi.rs#L270-L306), the recursive walk treats every object containing an array-valued `enum` key as an enum constraint. That includes arbitrary example/default payloads and extensions. A valid example such as `{"enum": ["sample"]}` would become an unsupported or incorrectly mapped constraint and fail the drift workflow. Traversal should be limited to actual schema locations and their `items`/composition children.

Verification completed successfully:

- `cargo test --workspace`
- Python drift-rendering tests
- Confirmed the analyzer is absent from `clickhouse-cloud-api`'s normal dependency graph

The existing Cursor finding about the live test including snapshot drift appears intentional: the live consumer is expected to surface stale snapshot operation/schema findings.

iskakaushik (approved):
LGTM

PR #307, opened 2026-07-22
Stop depending on Postgres get echoing credentials

### LINKED ISSUE #306: Managed Postgres API stops returning credentials outside create/password-reset on July 31, 2026 — fix before 0.4.0
## Background

The Postgres team announced that on **July 31, 2026** the Managed Postgres API will stop returning the superuser `password` and `connectionString` in its responses. After the change, credentials appear **only** in the create API and password-reset API responses. (Read replicas and restores inherit credentials from their source.)

We currently depend on the get endpoint echoing credentials in several places. We are fixing this **now**, before the 0.4.0 release, rather than waiting for the change to land and the OpenAPI drift workflow to flag it.

### Cloud integration tests (will start failing on July 31)

- `crates/clickhouse-cloud-api/tests/clickpipes/postgres_cdc_test.rs` — runs in CI. Polls `postgres_service_get` until running, then uses the **get** response's `connection_string` to connect to Postgres (line ~233), parses port/database out of it (lines ~240, ~258), and feeds `pg_ready.password` into the ClickPipe config (line ~256). All of these become empty strings after the change.
- `crates/clickhouse-cloud-api/tests/integration_postgres_test.rs:132-135` — asserts the get response has a non-empty `connection_string`; this assertion inverts on July 31.

**Fix:** capture credentials from the **create** response (both tests already call `postgres_service_create` and discard its body in favor of the polled get) and build the connection URI from `hostname` + port + `username` + the saved password. Note `PostgresService` has no `port` field today — the CDC test currently parses port from `connection_string`; default to 5432 unless the updated spec adds a `port` field.

### CLI (`crates/clickhousectl/src/cloud/postgres.rs`)

- `postgres get` prints `connection_string` (line ~541). Guarded by an is-empty check so it degrades silently, but it should stop relying on the echoed value — either drop it or construct the URI from `hostname`/`username` without the password.
- `postgres get --json` / `patch --json` will emit misleading `"password": ""` / `"connectionString": ""` once the API stops returning them (fields are `String` + `#[serde(default)]`).
- `postgres create` (lines ~601-606) keeps working — create still returns credentials — and its "password shown only once" wording becomes literally true. May need mechanical updates if the response model changes shape.
- `reset-password` is unaffected (already returns the locally-set password rather than trusting the echo).

### API library models (constrained by the spec)

`PostgresService` (`crates/clickhouse-cloud-api/src/models.rs:11648`) is the shared response model for create/get/patch and carries `password` and `connection_string`. The analyzer compares `models.rs` against the vendored snapshot and the live spec, so removing the fields **before** the upstream spec changes produces missing-field drift findings. Model changes land when the updated spec is published (snapshot refresh + model update + `client_test.rs` mock updates per the standard remediation flow in CLAUDE.md); everything else in this issue removes our dependence on the credential echo ahead of that.

### Not affected

- The Terraform provider half of the announcement (we are not a Terraform consumer).
- Local postgres commands.
- ClickHouse-service credentials (create/reset already the only source).

## Plan

1. Update both integration tests to source credentials from the create response and stop depending on get echoing them — safe against both the old and new API behavior.
2. Update the CLI `postgres get` handler to stop printing the echoed connection string (or derive a password-less URI from `hostname`/`username`).
3. When the upstream spec updates: standard drift remediation — refresh the vendored snapshot, update `PostgresService` (and any new create-response schema), fix `client_test.rs:2149`, and finish any CLI compile-site cleanups.

Docs: https://clickhouse.com/docs/cloud/managed-postgres/openapi

sdairs:
## Summary

The Managed Postgres API stops returning the superuser `password` and `connectionString` outside the create and password-reset responses on **July 31, 2026**. This PR removes our dependence on the get endpoint echoing credentials, ahead of the 0.4.0 release — covering steps 1 and 2 of #306. Step 3 (model/snapshot updates) waits for the updated upstream spec and follows the standard drift-remediation flow.

## Changes

- **`postgres_cdc_test.rs`**: capture `username`/`password` from the create response (the only surface guaranteed to keep returning them), parse port/database from the create response's connection string (default 5432/`postgres`), and connect to Postgres from discrete parts (`hostname` from the polled get + saved credentials) instead of the get-echoed connection string. Dropped the get-response connection-string assertion; added create-response credential assertions.
- **`integration_postgres_test.rs`**: assert `username`/`password`/`connectionString` are non-empty on the **create** response, and drop the connection-string assertion on the polled get (which inverts on July 31).
- **CLI `postgres get`**: no longer prints the echoed connection string (it was already guarded by an is-empty check, so it would have silently disappeared; hostname and username remain in the rendered output). Credentials are available from `postgres create` and `postgres reset-password`.

Both test changes are safe against the old **and** new API behavior.

## Testing

- `cargo test --workspace` — all pass (live-cloud integration tests remain `#[ignore]`d but compile).
- `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo fmt --all --check` — clean.

Part of #306

### DIFF
--- a/crates/clickhouse-cloud-api/tests/clickpipes/postgres_cdc_test.rs
+++ b/crates/clickhouse-cloud-api/tests/clickpipes/postgres_cdc_test.rs
@@ -158,6 +158,10 @@ async fn cloud_clickpipe_postgres_cdc() -> TestResult<()> {
         // credential-derived here rather than from the polled get below.
         let pg_username = pg_created.username.clone();
         let pg_password = pg_created.password.clone();
+        assert!(
+            !pg_created.connection_string.is_empty(),
+            "postgres create returned empty connection string"
+        );
         let pg_port = parse_pg_port(&pg_created.connection_string).unwrap_or(5432);
         let pg_database = parse_pg_database(&pg_created.connection_string)
             .unwrap_or_else(|| "postgres".to_string());
--- a/crates/clickhouse-cloud-api/tests/clickpipes/postgres_cdc_test.rs
+++ b/crates/clickhouse-cloud-api/tests/clickpipes/postgres_cdc_test.rs
@@ -152,6 +152,23 @@ async fn cloud_clickpipe_postgres_cdc() -> TestResult<()> {
             .result
             .ok_or("postgres create returned no result")?;
         let postgres_id = pg_created.id.to_string();
+        // The create response is the only API surface guaranteed to return
+        // credentials: from July 31, 2026 the get endpoint stops echoing
+        // `password` and `connectionString`, so capture everything
+        // credential-derived here rather than from the polled get below.
+        let pg_username = pg_created.username.clone();
+        let pg_password = pg_created.password.clone();
+        let pg_port = parse_pg_port(&pg_created.connection_string).unwrap_or(5432);
+        let pg_database = parse_pg_database(&pg_created.connection_string)
+            .unwrap_or_else(|| "postgres".to_string());
+        assert!(
+            !pg_username.is_empty(),
+            "postgres create returned empty username"
+        );
+        assert!(
+            !pg_password.is_empty(),
+            "postgres create returned empty password"
+        );
         cleanup.register_postgres(postgres_id.clone());
         eprintln!("  provisioned postgres id <redacted>");
 
@@ -206,10 +223,6 @@ async fn cloud_clickpipe_postgres_cdc() -> TestResult<()> {
         );
         let (pg_ready, ch_ready) = tokio::try_join!(pg_ready_fut, ch_ready_fut)?;
 
-        assert!(
-            !pg_ready.connection_string.is_empty(),
-            "empty pg connection string"
-        );
         assert!(!pg_ready.hostname.is_empty(), "empty pg hostname");
         let ch_endpoint = ch_ready
             .endpoints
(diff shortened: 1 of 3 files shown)


sdairs:
Addressed the Bugbot finding in 414367f: the CDC test now asserts the create response's connection string is non-empty before parsing port/database from it, so a missing value fails at provision time instead of silently falling back to `5432`/`postgres`.

The **Cloud integration tests** failure is not caused by this PR: ClickPipe creation returned `Api { status: 500, message: "Internal error" }`. The scheduled runs on main failed at the exact same step with the identical error on July 17 ([29553927900](https://github.com/ClickHouse/clickhousectl/actions/runs/29553927900)) and July 20 ([29716642475](https://github.com/ClickHouse/clickhousectl/actions/runs/29716642475)), while July 21–22 runs passed — it's an intermittent upstream API failure. Notably, the direct Postgres connection in the same test run succeeded using the credentials captured from the create response, which is exactly the code path this PR changed.

sdairs:
The re-run on 414367f reproduced the same failure: ClickPipe create → `Api { status: 500, message: "Internal error" }` (~3 min after the request), with every prior step passing — including the direct Postgres connection built from the create-response credentials. Four occurrences now (nightlies July 17 & 20, both PR runs today), all identical, while the July 21–22 morning runs passed. This looks like an upstream ClickPipes API issue rather than anything in this PR; worth flagging to the ClickPipes team if it persists.

iskakaushik (approved):

PR #310, opened 2026-07-23
Bring help, version, and bare invocations into the telemetry lifecycle

### LINKED ISSUE #309: Telemetry: help, version, and bare invocations bypass the telemetry lifecycle entirely (no notice, no marker, no event)
## Summary

The first-run telemetry notice, the `~/.clickhouse/telemetry.json` marker, and telemetry event capture only happen for invocations that parse into a real command and reach dispatch. `clickhousectl --help`, `-h`, subcommand help, `--version`, bare `clickhousectl`, and incomplete commands (e.g. `clickhousectl local` with no subcommand) all exit before the telemetry module is ever reached.

For a new user whose first invocations are all help-shaped — a very plausible onboarding path — we currently have zero visibility that they exist, and the first-run notice is deferred until their first "real" command.

**Desired behavior:** help/version/bare/incomplete invocations should participate in the normal telemetry lifecycle — show the first-run notice, write the marker file, and capture an event (subject to the same consent rules as any other command) — so we can see when help is someone's first (or only) interaction with the CLI.

## Why it happens today

In `crates/clickhousectl/src/main.rs`, the CLI parses via `Cli::command().try_get_matches_from_mut(...)` (`main.rs:41`). clap models help and version as parse *errors*, so all of these surface in the `Err` match arm and exit immediately:

- `--help` / `-h` / subcommand help → `ErrorKind::DisplayHelp` arm (`main.rs:54`): prints help + cached update notice, `std::process::exit(0)`.
- `--version` → `ErrorKind::DisplayVersion` arm (`main.rs:47`): prints version, force-refreshes the update cache, prints notice, exits 0.
- Bare `clickhousectl` / incomplete subcommand → clap's `DisplayHelpOnMissingArgumentOrSubcommand` kind, which falls into the `_ => e.exit()` catch-all (`main.rs:61`), same as genuine parse errors. The comment there is explicit and deliberate: "Parse errors exit here, before telemetry capture: a mistyped invocation never produces an event."

All of these exits happen **before** `telemetry::capture` (`main.rs:67`) and **before** `telemetry::finalize` (`main.rs:125`). `finalize` is the only place that writes `telemetry.json` and prints the first-run notice — the file doubles as the "notice has been shown" marker (`telemetry.rs` module doc). So on these paths there is no capture, no event, no notice, and no file, ever.

Verified empirically on a fresh `$HOME`: repeated `--help` / bare invocations never create `~/.clickhouse/` or `telemetry.json`.

## Related inconsistency worth fixing at the same time

Bare `clickhousectl` and `clickhousectl --help` behave differently in ways users can observe:

| Invocation | Exit code | Update notice shown? |
|---|---|---|
| `clickhousectl --help` | 0 | yes (from cache) |
| `clickhousectl` (no subcommand) | 2 | no |

The bare invocation prints the same help text but goes through the `_ => e.exit()` catch-all, so it skips the update notice and exits 2. Exit 2 for a bare invocation is defensible (git-style usage error), but the update-notice asymmetry looks accidental, and whatever telemetry treatment we choose should be consistent across both.

## Invariants any remediation must preserve

- **Consent model (Homebrew-style, issue #283):** nothing is ever *sent* before the first-run notice has been shown. Today even the notice-showing invocation sends nothing (`Action::Notice` writes the marker and prints; the first event goes out on the next consented invocation). Showing the notice on help must not weaken this.
- **`DO_NOT_TRACK`** overrides everything: no notice, no file write, no send.
- **Unwritable config dir fails open to silent:** notice is only printed after the marker write succeeds.
- **Value privacy:** `telemetry::capture` walks `ArgMatches` ids and clap `Arg` metadata only — command path and flag *names*, never values or positionals. This is structurally guaranteed today and must stay that way.
- **The hidden `telemetry send` child** (`main.rs:74-83`) must never produce an event of its own.

## Design considerations for the remediation (for whoever picks this up)

1. **No `ArgMatches` exists on these paths.** `try_get_matches_from_mut` returns `Err`, so `telemetry::capture(&cmd, &matches)` can't run as-is. The event's command path for a help invocation has to come from somewhere else (e.g. clap error context, or re-walking raw args against the clap definitions with the same names-only discipline).
2. **What should the event look like?** Presumably the deepest recognized command path plus a marker that it was help/version/bare (e.g. `--help` flag name in `flags`, or a dedicated field). Needs a deliberate decision.
3. **Should genuinely mistyped invocations (typos, invalid flags) produce events too**, or only "well-formed" help/version/bare/incomplete? The current no-event-on-parse-error behavior was intentional; the ask here is specifically help/version/bare/incomplete-subcommand. Extending to arbitrary parse errors needs care: captured names come from clap definitions so value leakage stays structurally impossible, but the mistyped token itself must never be captured.
4. **Notice ordering:** the first-run notice should presumably print after the help text, alongside/near the update notice.
5. **Ingest side:** confirm the Cloudflare worker / downstream schema tolerates whatever new event shape is chosen (`flags` cap of 64 etc.).
6. Fold the bare-invocation update-notice inconsistency (table above) into the same restructuring, since both stem from the early-exit match in `main.rs:43-63`.

## Acceptance sketch

- First-ever invocation being `--help`, `-h`, `--version`, bare `clickhousectl`, or an incomplete subcommand shows the first-run notice and creates `telemetry.json`.
- Subsequent help/version/bare invocations produce telemetry events under the same consent rules as normal commands.
- `DO_NOT_TRACK`, disabled state, and unwritable-dir behavior unchanged.
- Update notice behavior is consistent between `--help` and bare invocation.
- Existing telemetry tests still pass; new tests cover the help/bare paths.

sdairs:
Closes #309

## What

`--help`/`-h`/subcommand help, `--version`, bare `clickhousectl`, and incomplete command groups previously exited inside the clap `Err` match before the telemetry module ran — no first-run notice, no `telemetry.json` marker, no event, ever. They now join the normal telemetry lifecycle under the same consent rules, and the update notice is printed consistently across `--help` and the bare invocation (which keeps clap's usage-error exit code 2).

## How

- **`telemetry::capture_from_args`** — no `ArgMatches` exists on the clap error path, so the invocation is re-derived from raw argv against the built clap definitions. The names-only discipline holds structurally: every recorded string is cloned from clap metadata (`Command::get_name` / `Arg::get_long`), never from a user-supplied token. Unresolvable tokens, values following value-taking flags, and everything after `--` are skipped.
- **`main.rs` Err arm** — `DisplayHelp`/`DisplayVersion` (exit 0) and `DisplayHelpOnMissingArgumentOrSubcommand`/`MissingSubcommand` (exit 2, clap 4 splits flag-less vs. flag-bearing incomplete groups into these two kinds) all fall through to the update notice + `telemetry::finalize`. Genuinely mistyped invocations (typos, unknown flags) keep the intentional no-event `_ => e.exit()` behavior.
- **Payload schema unchanged**: help/version surface as flag names (`["help"]`/`["version"]`), a bare invocation as `command:""`, incomplete groups as a group path no successful run can produce, all with the real exit code.

## Invariants preserved

- Homebrew-style consent: the notice-showing run sends nothing; first event on the next consented run (pinned by test on the help path).
- `DO_NOT_TRACK` overrides everything; unwritable config dir fails open to silent; the hidden `telemetry send` child never produces its own event.

## Tests

- 10 new unit tests for `capture_from_args` (path depth, value/positional non-leakage incl. `--flag=value`, value tokens matching subcommand names, `--` terminator, unknown tokens, non-UTF-8 tokens). The test helper attempts the same failing parse before `build()` so partially-built clap state matches production exactly.
- 7 new integration tests in `telemetry_test.rs` covering first-run notice/marker on `--help`, event shapes for help/version/bare/incomplete, update-notice consistency on the bare invocation, typo remaining event- and marker-free, and DNT on the help path.

## Note for deployment

The ingest worker (separate repo) should be confirmed to tolerate `command: ""` (bare invocation events) — the schema is otherwise unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -891,12 +891,14 @@ Each event contains exactly:
 
 - the command path (e.g. `local start`)
 - the **names** of the flags passed (e.g. `json`, `org-id`) — never flag values, never positional arguments
-- the exit code (`gh`-style: 0 success, 1 error, 2 cancelled, 4 auth required)
+- the exit code (`gh`-style: 0 success, 1 error, 2 cancelled/usage error, 4 auth required)
 - the CLI version, OS, and architecture
 - whether it ran in CI (`CI` env var)
 - whether it ran under a detected coding agent, and if so which one (e.g. `claude-code`)
 
-There is no install ID, no device ID, and no fingerprinting of any kind. The payload is built from the clap command definitions rather than the raw command line, so leaking an argument value is structurally impossible — the code that builds the event has no access to values at all.
+There is no install ID, no device ID, and no fingerprinting of any kind. Every string in the event is cloned from the clap command definitions — subcommand and flag names as clap defines them — never from what you actually typed, so leaking an argument value is structurally impossible.
+
+Help, version, bare, and incomplete invocations (`--help`, `--version`, `clickhousectl` with no subcommand, or a command group without one) count as usage like any other command and produce the same names-only events under the same consent rules. Genuinely mistyped invocations (unknown commands or flags) never produce an event.
 
 Nothing is ever sent before you have seen the notice or explicitly enabled telemetry with `clickhousectl telemetry enable`: the first run prints a one-time notice to stderr, records that it was shown in `~/.clickhouse/telemetry.json`, and sends nothing. Sending starts from the following run — or immediately if you opt in by running `telemetry enable`, which is explicit consent and skips the notice. The send happens in a short-lived detached process, so command latency is unaffected even when the endpoint is unreachable.
 
--- a/crates/clickhousectl/src/main.rs
+++ b/crates/clickhousectl/src/main.rs
@@ -41,25 +41,45 @@ async fn main() {
     let matches = match cmd.try_get_matches_from_mut(std::env::args_os()) {
         Ok(matches) => matches,
         Err(e) => {
-            match e.kind() {
-                // --version always hits the network to refresh the cache + timer,
-                // then prints the notice from the freshly-updated cache.
+            let exit_code = match e.kind() {
+                // --version always hits the network to refresh the cache +
+                // timer; the notice below then prints from the fresh cache.
                 ErrorKind::DisplayVersion => {
                     e.print().expect("failed to print output");
                     update::force_refresh_update_cache().await;
-                    update::print_cached_update_notice();
-                    std::process::exit(0);
+                    e.exit_code() // 0
                 }
                 // --help shows the notice from cache (no blocking network call).
                 ErrorKind::DisplayHelp => {
                     e.print().expect("failed to print output");
-                    update::print_cached_update_notice();
-                    std::process::exit(0);
+                    e.exit_code() // 0
                 }
-                // Parse errors exit here, before telemetry capture: a mistyped
-                // invocation never produces an event.
+                // Structurally valid but incomplete: bare `clickhousectl`
+                // (full help), or a command group given flags but no
+                // subcommand (usage error). Both print to stderr and keep
+                // clap's usage-error exit code, but participate in the
+                // update notice and telemetry below like --help does.
+                ErrorKind::DisplayHelpOnMissingArgumentOrSubcommand
+                | ErrorKind::MissingSubcommand => {
+                    let _ = e.print();
+                    e.exit_code() // 2
+                }
+                // Genuine parse errors (typos, unknown flags) exit here,
+                // before telemetry capture: a mistyped invocation never
+                // produces an event.
                 _ => e.exit(),
+            };
+            update::print_cached_update_notice();
+            // No `ArgMatches` exists on this path, so the invocation is
+            // re-derived from raw argv against the built clap definitions —
+            // same names-only discipline, same consent rules as any command.
+            #[cfg(feature = "telemetry")]
+            {
+                cmd.build();
+                let args: Vec<std::ffi::OsString> = std::env::args_os().skip(1).collect();
+                telemetry::finalize(telemetry::capture_from_args(&cmd, &args), exit_code);
             }
+            std::process::exit(exit_code);
         }
     };
 
--- a/crates/clickhousectl/src/telemetry.rs
+++ b/crates/clickhousectl/src/telemetry.rs
@@ -12,9 +12,11 @@
 //! everything: no notice, no file write, no send.
 //!
 //! The payload carries the command path and flag *names* only — never flag
-//! values, never positional arguments. It is built from the clap definitions
-//! ([`capture`] walks `ArgMatches` ids and `Arg` metadata, never touching
-//! `get_one`/`get_raw`), so leaking a value is structurally impossible.
+//! values, never positional arguments. Every recorded string is cloned from
+//! the clap definitions ([`capture`] walks `ArgMatches` ids and `Arg`
+//! metadata; [`capture_from_args`] resolves raw tokens against `Command` and
+//! `Arg` metadata and drops whatever doesn't resolve), so a user-supplied
+//! token can never appear in an event.
 //!
 //! Transport is a detached child process (`clickhousectl telemetry send`,
 //! hidden): the parent spawns it with all stdio nulled and never waits, so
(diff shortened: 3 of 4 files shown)


sdairs:
## Review: structural assessment

Reviewed the issue (#309) and the full diff. Overall this is a sound approach, not a hack — the scope is justified and the privacy invariants hold. A few items below.

### The approach is correct

The fundamental constraint is that clap gives no `ArgMatches` on the error path (`try_get_matches_from_mut` returns `Err`), so `telemetry::capture()` — which walks `ArgMatches` ids — can't be reused. The alternatives are all worse: re-parsing without `--help`/`--version` is fragile and doesn't cover bare/incomplete invocations; clap's error context doesn't expose enough to reconstruct the flag set. A parallel `capture_from_args` that walks raw argv against clap's `Command`/`Arg` metadata is the right answer.

The `main.rs` restructuring is clean — returning an exit code from the match and running shared post-match code (update notice + telemetry + exit) correctly unifies the update-notice behavior across `--help` and bare invocation (the asymmetry the issue flagged). `--version` still force-refreshes the cache before the shared notice print, preserving ordering.

The names-only privacy invariant is maintained by construction in both implementations: every recorded string is cloned from clap definitions (`Command::get_name` / `Arg::get_long`), never from a user token. Test coverage is thorough (10 unit + 7 integration).

### Should fix before merge

**Non-UTF-8 + `skip_next_value` ordering bug** (`telemetry.rs:292-300`). When a value-taking flag sets `skip_next_value` and the next token is non-UTF-8, `token.to_str()` hits `continue` without clearing the flag. The following token is then wrongly consumed as the value, so later flags (e.g. `--help`) can be omitted from the event. Fix is a one-line swap — clear `skip_next_value` *before* the UTF-8 check:

```rust
if skip_next_value {
    skip_next_value = false;
    continue;
}
// Non-UTF-8 tokens can only be values; skip them.
let Some(token) = token.to_str() else {
    continue;
};
```

This is exactly the kind of edge case a second parser implementation invites — clap's battle-tested parser handles this, the hand-rolled one doesn't.

### Deployment item

The PR notes the ingest worker should be confirmed to tolerate `command: ""` (bare invocation events). Worth verifying before this ships.

### Minor / optional

- The two parallel implementations (`capture` via `ArgMatches`, `capture_from_args` via raw argv) must stay in sync on the names-only contract. A brief coupling note in `capture_from_args` pointing back to `capture` would make the "keep in sync" obligation explicit for future maintainers.
- `cmd.build()` on a partially-built command (`main.rs:78`) works today and is well-documented, but is a mild fragility point dependent on clap's tolerance for double-building. No better option given the constraints.

sdairs:
## Structural review

I reviewed this change specifically for whether it's a sound foundation or a hacky fix, given how wide-ranging it looks. **Verdict: the structure is sound.** The "wide" feel is mostly tests — of the 434 added lines, roughly 300 are unit/integration tests. The production change is ~130 lines of new capture logic plus a `main.rs` restructure that made the error path *simpler*, not more special-cased. All 28 telemetry unit tests pass locally.

### Why some new mechanism was unavoidable

The root cause is a clap design decision we can't route around: help, version, and incomplete-subcommand invocations are modeled as parse *errors*, so no `ArgMatches` ever exists on those paths and the existing `telemetry::capture` has nothing to walk. The alternatives to `capture_from_args` are all worse:

- **Disabling clap's auto help/version** and modeling them as ordinary flags would touch every command enum in `cli.rs` and forfeit clap's help propagation — that would have been the genuinely invasive change.
- **Stripping `--help` tokens and re-parsing** still fails for bare and incomplete-group invocations (the same error class), so error-path handling is needed regardless.
- **Scraping clap's rendered error/usage text** is fragile string parsing.
- **`Command::ignore_errors`** is documented as best-effort, unsuitable as a foundation.

### Why the shadow parser is acceptable here

`capture_from_args` is a partial re-implementation of clap's argv walk, which would normally be a red flag. What makes it defensible is the asymmetry in its failure modes: every recorded string is cloned from clap metadata (`Command::get_name` / `Arg::get_long`), never from a user token, and anything unresolvable is dropped. When it diverges from real clap semantics, the worst outcome is an under-recorded or distorted *path* in an analytics event — never a leaked value, never a behavior change. A shadow parser feeding telemetry only needs to be approximately right, and the doc comments show this trade was made deliberately.

The `main.rs` change is a net improvement: three arms that each did their own `exit()` converged into "compute exit code, then shared tail" (update notice → finalize → exit). That convergence fixes the bare-invocation update-notice inconsistency as a natural consequence rather than another special case, and typos still hit `_ => e.exit()`, so the intentional no-event-on-mistype behavior survives (pinned by test). Everything downstream reuses the existing pipeline — same `Invocation` type, same `decide` state machine, unchanged payload schema.

### Requests

1. **Add a consistency test tying the two capture implementations together.** `capture` (from `ArgMatches`) and `capture_from_args` (from argv) are parallel implementations of the same names-only discipline, and nothing currently catches drift between them. For a list of *successfully parseable* invocations, assert `capture_from_args` produces the same `Invocation` as `capture`. Cheap, and it's the only thing that will catch divergence.
2. **Document the clap-settings assumptions the shadow parser depends on.** It doesn't understand `infer_long_args` / `infer_subcommands` — neither is used today, but enabling one would make `capture_from_args` silently under-record with no signal. Similarly, `skip_next_value` consumes exactly one token, so a future `num_args(2..)` flag could let a trailing value that matches a subcommand name distort the path (benign, but path-quality erosion). A short comment in `cli.rs` or a guard test naming these assumptions would future-proof it. (The one `value_delimiter = ','` in the CLI today is inline-only and safe.)
3. **Close the ingest-worker question before merge.** The PR body says `command: ""` tolerance "should be confirmed" — bare-invocation events start flowing immediately on release, so this should actually be verified, not just flagged.

### Minor warts, fine to leave

- Exit code 2 now means both "cancelled" and "usage error" in analytics — distinguishable, since incomplete-group events carry a path no successful run produces, and the README was updated.
- `help cloud` records `"help"` while `cloud --help` records `"cloud"` + help flag — tested and documented; resolving the help topic would add complexity for little gain.
- The new arm uses `let _ = e.print()` where the older arms use `.expect(...)` — cosmetic inconsistency.

PR #311, opened 2026-07-25
Fix OpenAPI drift: ClickStack expansion, quotas, and model updates

sdairs:
Closes #308

Remediates the OpenAPI drift between the live ClickHouse Cloud spec and `clickhouse-cloud-api`. The drift had grown since the issue was filed; this PR remediates today's live document (findings at remediation time: 25 missing client methods, 29 missing model types, 27 missing struct fields, 5 optionality mismatches, 5 missing enum values, 25 newly-beta operations).

## What's added

- **ClickStack sources**: `ClickStackPromqlSource` (new union member), get/create/update/delete endpoints, and restructured source schemas — new `section`/`disabled`/`knownColumnsListExpression`/`useTextIndexForImplicitColumn` fields (the last is a string enum `auto`/`enabled`/`disabled` per the live spec, not a boolean).
- **ClickStack connections, roles, saved searches**: full models + CRUD endpoints (roles carry free-form CASL permissions with `serde_json::Value` conditions).
- **ClickStack webhooks**: `ClickStackWebhookInput` + create/update/delete (spec has no get-by-id); responses reuse the existing `ClickStackWebhook` union.
- **Dashboard validation**: `click_stack_validate_dashboard` + response models.
- **Organization quotas**: `OrganizationQuota` + list/get endpoints.
- **Chart models**: shared 13-value `ClickStackChartColor` palette enum, background charts, number-tile color conditions (numeric/between/equality union), categorical-bar and event-patterns tile configs wired into `ClickStackTileConfig`, `ClickStackOnClickExternal` wired into `ClickStackOnClick`.
- **Existing models**: alert `numConsecutiveWindows`, filter `appliesToSourceIds`, line `fitYAxisToData`, pie `limit`/`orderBy`, ClickPipes `partitionByExpr` (required on `ClickPipePostgresPipeTableMapping`, nullable on the patch-remove mapping), 4 new `ActivityType` values, alert state `PENDING`.
- Snapshot refreshed to the live document; `BETA_OPERATIONS` regenerated (+25 entries).

## Breaking changes (library consumers, pre-1.0)

- Source structs' `id` is now `Option<String>` (spec: server-generated, no longer required).
- `ClickStackTraceSource::default_table_select_expression` is now required `String`.
- `ClickStackSource`, `ClickStackTileConfig`, and `ClickStackOnClick` untagged unions gained variants.
[...]

PR #313, opened 2026-07-26
Adopt tolerant-response deserialization: sweep serde(default) and codify the policy

### LINKED ISSUE #312: Adopt tolerant-response deserialization: sweep #[serde(default)] across model fields and codify the policy in CLAUDE.md
## Problem

The library is inconsistent about `#[serde(default)]` on required (non-`Option`) model fields, and PR #311's review remediation pulled it in the wrong direction. Current state of `crates/clickhouse-cloud-api/src/models.rs`:

- **668** required fields carry `#[serde(default)]` (tolerant: missing field on deserialize → `Default::default()`)
- **382** required fields do not (strict: missing field fails the whole response deserialization)
- **12** `Option` fields lack `#[serde(default)]` (harmless — serde treats missing `Option` as `None` — but inconsistent)
- ~50 of the tolerant fields are in ClickStack models; commit a768db6 (PR #311 remediation, prompted by a Bugbot finding) stripped `serde(default)` from `ClickStackTraceSource.default_table_select_expression` and added strictness regression tests, establishing a "strict" convention that contradicts the majority pattern and the CLAUDE.md line "Retain `#[serde(default)]` on model fields". (The #311 branch has since reverted that commit in dcdb3aa, inverting the trace-source test into a missing-field round-trip; the pre-existing strict fields, including `ClickStackLogSource`'s, remain.)

Multiple teams evolve the Cloud API independently and can remove response fields without notice. With strict deserialization, a dropped response field breaks the CLI for users even though the API call itself succeeds — and via `discriminated_union!` (which propagates payload errors for known discriminators rather than falling back to `Unknown`), one missing field on one trace source fails an entire `list sources` response. The 382 strict fields being "fine" so far is survivorship, not safety: each is a latent outage point whose field simply hasn't been dropped yet.

## Why tolerant-response is the right policy

`#[serde(default)]` only affects **deserialization** — it never changes what we serialize. On a required `T` field it yields exactly "strict in what we send, liberal in what we accept" (Postel's law) with a single shared model:

- **Requests stay strict via the type system**: a required field is `T`, so it must be populated to construct the request and is always serialized. No serde attribute weakens this.
- **Responses degrade gracefully**: a dropped field becomes `""`/`0`/`false` instead of a hard error. For a CLI, an empty cell beats a dead command.
- **Detection is CI's job, not runtime's**: the analyzer derives requiredness purely from `T` vs `Option<T>` (`rust_inventory.rs` reads only `rename`/`untagged`/`other`), so tolerance costs zero drift-detection fidelity. The daily drift workflow turns spec drift into a GitHub issue instead of a user-facing outage.
- **Internal consistency**: the library already ignores unknown fields (no `deny_unknown_fields`), gives every string enum an `Unknown(String)` catch-all, and (as of #311) lands unknown union payloads in `Unknown(serde_json::Value)`. Strict struct fields are the one place the library still hard-fails on server evolution.
- **Ecosystem precedent**: proto3/gRPC (missing scalar → default value, clients must tolerate), AWS Smithy SDKs (lenient outputs, input requiredness enforced at build time), Kubernetes API conventions / kube-rs (`Option` + `default` pervasively), Fowler's Tolerant Reader. openapi-generator's strict Rust output is the cautionary tale, not the idiom.

### Alternative considered and rejected: blanket `Option<T>` on response fields

171 model types are transitively shared between request and response positions (computed from `client.rs` signatures), so blanket-`Option` would (a) lose type-level enforcement of required request fields, (b) destroy the `T` vs `Option<T>` code↔spec requiredness mapping the analyzer verifies bidirectionally, and (c) push `unwrap_or_default()` sentinels to every call site instead of declaring the tolerance once in the model.

## Proposed policy (to codify in CLAUDE.md)

> **Optionality and deserialization tolerance.** Requiredness is expressed only in the type: required non-nullable fields are `T`, optional/nullable fields are `Option<T>` + `skip_serializing_if`. We are strict in what we send (the type system enforces required request fields) and liberal in what we accept: **every model field carries `#[serde(default)]`**, unknown fields are ignored (never `deny_unknown_fields`), and enums/unions carry `Unknown` catch-alls. Deserialization of a response must never fail because the API added, removed, or changed a field — spec conformance is enforced by the analyzer and the daily drift job, not by runtime errors. Exception: a field that structurally discriminates an untagged union must stay strict (prefer converting the union to `discriminated_union!` instead). If defaulting a specific field would be misleading, make it `Option<T>` with an `optionality_exemptions` entry, not strict.

This replaces the current unexplained "Retain `#[serde(default)]` on model fields" line, whose lack of rationale is exactly what let a review bot ratchet PR #311 toward strictness.

## Scope of work

This lands as **one PR** delivering the complete end state: union conversions, an enforcement test, the CLAUDE.md policy, and the full sweep. There is no interim policy/code gap to manage — internal sequencing is left to the implementer, with one exception called out below. Baseline is main after PR #311 merges: `ClickStackSource`, `ClickStackTileConfig`, `ClickStackOnClickTarget`, and the webhook/alert-channel-adjacent unions are already `discriminated_union!`-dispatched there.

### Union conversions — the one hard ordering constraint

Convert the remaining structurally-matched untagged sub-unions to discriminator dispatch **before (or together with) defaulting their distinguishing fields**: the Builder-vs-RawSql chart configs (Line/Bar/CategoricalBar/Table/Number/Pie) and `ClickStackAlertChannel`. These are the one place required-field strictness is load-bearing: untagged matching picks the first variant whose required fields are present, so defaulting the distinguishing fields (`configType`, `sql`, …) while the unions are still untagged would make Builder swallow RawSql payloads.

RawSql payloads carry `configType: "rawSql"`; Builder payloads lack the key, so this needs a **key-absence arm** in `discriminated_union!` (`None => Variant`) — the current macro dispatches only on a key's string value. Two semantics to pin down with explicit tests, not accidents:

- `None` conflates "key absent" and "key present but not a string" — decide deliberately (treating both as Builder is probably right) and test it.
- Once the sweep lands, Builder variants become total (every field defaultable), so the absence arm always succeeds and the sub-union's `Unknown` is reachable only via unrecognized `configType` values. This matches spec semantics (payloads reaching the sub-union already passed the parent's `displayType` dispatch, and "no `configType`" *is* the builder discriminator), but it changes what `Unknown` means for these unions — assert the new behavior in tests.

### Enforcement test

Nothing currently enforces the policy: the analyzer ignores `serde(default)`, so a future agent (or bot-prompted remediation) could strip it again silently. Make the policy self-enforcing:

- Add `default: bool` to `SerdeOptions` in `rust_inventory.rs` (small; the parser already handles serde attrs).
- Add a plain test in `clickhouse-cloud-api` (reusing the analyzer inventory as the existing dev-dependency — no new `FindingKind`, no report `schema_version` bump, no Python rendering; this is an internal convention, not spec drift) asserting every `T` field carries `#[serde(default)]`.
- Since the sweep lands in the same PR, the test ships with an **empty exception list** — or, if any structural-discriminator field genuinely cannot be converted (ideally none), a permanent, commented entry.

### CLAUDE.md policy

Land the policy text above, replacing "Retain `#[serde(default)]` on model fields". Include the rationale so future agents can cite it when review bots re-file "serde(default) masks required field" findings.

### The sweep

Add `#[serde(default)]` to every remaining strict `T` field — ClickStack and all legacy domains (Activity, ApiKey, Backup, ClickPipes, Postgres, SCIM, …) — adding `Default` impls where a field type lacks one, plus the 12 bare `Option` fields for uniformity. Replace any remaining strictness regression tests with missing-field-degrades-to-default round-trip tests (the trace-source one was already inverted in dcdb3aa on the #311 branch).

### Documentation

Document the round-trip caveat in the crate docs: a consumer that GETs an object and writes it back could persist a defaulted `""` for a field the server stopped sending. The CLI doesn't round-trip (update commands build requests from flags; PATCH types are separate and all-optional), and exposure is bounded by the daily drift job.

The sweep is invisible to the analyzer's drift comparison, so drift reports are unaffected.

## Acknowledged tradeoff

Sentinel masking: a dropped required response field renders as `""`/`0`/`false` rather than erroring. This is the intended failure mode for a CLI; per-field exceptions where a sentinel would mislead (e.g. cost figures) use `Option<T>` + `optionality_exemptions`.

sdairs:
Closes #312

**Stacked on #311** — base is `issue-308-clickstack-drift`; retarget to `main` after #311 merges.

Adopts "strict in what we send, liberal in what we accept" across `clickhouse-cloud-api`: every model field now carries `#[serde(default)]`, so a server-dropped response field degrades to a default instead of failing the whole response. Requests stay strict via the type system (`T` vs `Option<T>` is unchanged, so analyzer drift detection is unaffected).

## What's in here (one commit per step)

- **`ClickStackAlertChannel` → `discriminated_union!`** dispatch on `type` (`email`/`webhook`), with its variants' strict fields defaulted in the same commit (defaulting a distinguishing field while the union is still untagged would misroute payloads — that ordering constraint holds throughout this PR).
- **Six Builder-vs-RawSql chart-config sub-unions → `discriminated_union!`** via a new key-absence arm: RawSql payloads carry `configType: "sql"` (note: the issue body said `"rawSql"`; the spec and existing tests say `"sql"`), Builder payloads lack the key entirely. Absent and present-but-non-string keys both dispatch to Builder — deliberate and tested. The absence arm takes a mandatory disqualifying-key guard (`none unless "connectionId" | "sqlTemplate"`) so a RawSql body whose `configType` the server stops sending lands losslessly in `Unknown` instead of silently becoming an empty builder tile.
- **Analyzer**: `rust_inventory.rs` now tracks `serde(default)` (bare and `default = "path"` forms), exposed via a narrow `model_fields_missing_serde_default()` — no new `FindingKind`, no report `schema_version` change, no Python changes, and `syn` stays out of the published crate's dependency graph.
- **The sweep**: `#[serde(default)]` on every remaining strict field (~365 fields across ClickStack, SCIM, ClickPipes, Postgres, quotas, RBAC) plus 12 bare `Option` fields, enforced by a new `every_model_field_carries_serde_default` test in `spec_coverage_test.rs` with an **empty exception list**.
- **Policy docs**: AGENTS.md's unexplained "Retain `#[serde(default)]`" line replaced with the full policy and rationale (so future review-bot "serde(default) masks required field" findings can be answered by citation); crate docs gain a "Tolerant deserialization" section including the GET→write-back caveat.

## Semantics decided during adversarial review

An adversarial review pass confirmed three real gaps, fixed in the final commit:

1. **Known-discriminator payload errors now fall back to `Unknown(value)`** instead of hard-failing. Previously one malformed element (e.g. a field whose *type* changed server-side) failed an entire `list` response — exactly the outage mode this PR exists to close. `Unknown` is now reachable via unrecognized discriminator values *or* undeserializable payloads, both round-tripping losslessly; the macro docs and tests pin both routes.
2. **RawSql-without-`configType` misroute** — fixed by the `none unless` guard described above.
3. **`ClickStackAlertChannel::default()` round-tripped to the wrong variant** (`ClickStackAlertChannelEmailType`'s `#[default]` sat on `Webhook` because the spec gives both variants the same two-value enum). `#[default]` moved to `Email`; a new cross-union invariant test asserts every `discriminated_union!` enum's `Default` round-trips to the same variant.

Residual (documented, intentional): a plain struct field whose type changes server-side still fails the response — `#[serde(default)]` only covers missing keys. Explicit JSON `null` on `Vec` fields also remains an error (out of scope per the issue).

## Verification

Every commit is independently green; on the final tree: `cargo test --workspace` (0 failures), `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check`, `cargo check --workspace --all-features`, `python3 -m unittest discover -s scripts/tests`, and `python3 scripts/check-openapi-drift.py --dry-run` confirms the sweep is analyzer-neutral.


sdairs:
Reviewed with focus on the idea, the policy, the standard, and the structure. Overall: approve — the policy is coherent, honestly documented (including what it does *not* cover), and enforced structurally via the empty-exception-list `every_model_field_carries_serde_default` test rather than by convention. Suggestions, in priority order:

1. **The `none unless` guard lists are a hand-maintained invariant with doc-only enforcement.** The macro docs say "If the spec ever gives the absence variant one of the guard keys, drop that key from the list" — but nothing fails if that drifts. If a Builder schema ever gains `connectionId`, those payloads silently start landing in `Unknown`. The analyzer inventory already knows every struct's wire fields; a check that each guard key is absent from the absence-variant's fields and present on some other variant's would make this structural, in the same spirit as the `VALUES` const check. Suggest a follow-up issue.

2. **The `assert_default_round_trips!` union list in `models_test.rs` is also hand-maintained** ("Add new unions with a `Default` impl here"). A new `discriminated_union!` with a `Default` won't fail anything if it's forgotten there. Same remedy as (1), lower priority since the failure it guards is rarer.

3. **Minor perf nit:** `deserialize_or_raw` clones the full `Value` on every call, including the success path. `&Value` implements `Deserializer`, so `T::deserialize(&value)` would avoid the clone and keep the original for the error path. Only matters for large dashboard payloads; fine to leave.

4. **Debuggability note:** `deserialize_or_raw` discards the serde error entirely, so when a payload-shape mismatch lands in `Unknown`, nothing records *why*. Lossless round-trip + `Display` showing the raw JSON makes this acceptable — but if support tickets ever hinge on "why is this tile Unknown," a `debug!`-level hook is where it would go.

5. **Non-blocking observation on the policy's cost:** failure moves from loud to silent — a dropped `id` on a list response is now `""` flowing into follow-up calls rather than a deserialize error. The docs acknowledge this and the daily drift job bounds exposure to ~one day; just worth remembering that the drift job is now the policy's correctness backstop, not merely hygiene, so its cadence shouldn't be relaxed later.

(5) needs no action; (3) and (4) are take-or-leave; (1) and (2) are worth follow-up issues.

sdairs:
I read the adversarial review above and checked its three findings against the code. It's technically accurate on all three — but I disagree with the severity ratings and the proposed architecture. My take:

**[P2, lib.rs:46–48] — correct, easy fix, should block.** "Bounded by the daily OpenAPI drift job" overstates the safety net: the analyzer compares spec ↔ Rust source, so a deployed server omitting a still-spec-required field is invisible to it — precisely the failure mode this PR masks. Weaken the sentence to scope the bound to *spec drift* and admit server-behavior drift is only covered by integration tests. Honest docs, no architecture change.

**[P1, `ScimUser.active`] — directionally right, overstated, and the fix already exists in this PR.** `active` is spec-required, so the default only fires on anomalous/partial representations. But `false` is a semantic default, not an inert one: `if user.active { allow() }` gets a fabricated deny (fail-closed for SCIM, but the class critique holds). The PR's own documented escape hatch — "if defaulting a specific field would be misleading, make it `Option<T>` with an `optionality_exemptions` entry" — was written but never applied. Suggest a targeted audit of semantic defaults (deny/enable flags, timestamps, quota numerics — maybe a dozen fields) rather than a blanket redesign.

**[P1, models_test.rs:4102] — the test is mislabeled, not wrong.** `ClickPipePostPubSubSource` is indeed only reachable via `ClickPipePostRequest`, and "dropped response fields" is the wrong frame. But the test asserts the crate's public Deserialize contract (sparse JSON deserializes; serialization unchanged), which is legitimately tested where it lives. Rename to something direction-neutral (`tolerates_sparse_json`) and fix the comment.

**Where I part ways with the review:**

- "The type system does not enforce required inputs" is strictly true and practically false. Every in-tree caller builds requests with struct literals naming all required fields (`ClickPipePostRequest { name, source, ..Default::default() }` — compiler-enforced). The real hole is the `..Default::default()` spread plus defaults on *leaf* credentials — and that **pre-dates this PR**: `main` today sends `{username: "", password: ""}` via `ClickPipeKafkaSchemaRegistryCredentials::default()` (commands.rs:1170) when `--schema-registry-url` is passed without credentials. That's a request-*construction* concern that strict *deserialization* (the review's option a) would not fix at all.
- The serde mechanism gap the review demonstrates (`from_str::<ClickPipePostRequest>("{}")` now succeeds) is real but second-order: nothing in this repo deserializes request bodies, and the caveat is already documented in lib.rs. It affects hypothetical downstream consumers of the published crate, bounded by semver + docs.
- "Classify models by request/response reachability" is heavier than it looks. `ClickStackSource` is the only shared GET/POST model in this surface, and a send-time validator must re-derive the *wire* requiredness contract (which already diverges from the type contract via `skip_serializing_if = "Vec::is_empty"`). That's a larger diff than this PR, protecting a caller pattern that doesn't exist in-tree.

**Suggested path — approve with two blocking items, not request-changes:**

1. Fix the lib.rs bound sentence (blocking, trivial).
2. Apply the existing exemption policy to `ScimUser.active` + audit the small class of semantic defaults (blocking, small).
3. Rename the PubSub test and fix its comment (non-blocking).
4. File a follow-up issue for `Default::default()` spreads + leaf-credential defaults as a request-construction concern — explicitly out of scope here; it needs `Option`-typed credentials or a build-time `validate()`, owned by the ClickPipe request builders.

The review's north star (responses usable without inventing facts; incomplete requests unsendable; missing data distinguishable from real `false`/`0`) is right, and this PR achieves the first leg cleanly. The request-side gap is a pre-existing construction hole, not a regression — blocking a well-tested response-tolerance fix on a request-architecture redesign lets the perfect be the enemy of the good.

sdairs:
Superseded by #316 (issue #314): the serde(default) sweep fabricated values indistinguishable from server-sent ones. The union-semantics commits from this branch (125ad7a, d57839e, and the carry-over parts of 6d1eb1b) live on in #316.

Issue #321, opened 2026-07-29
Local child-process passthrough exits bypass the telemetry tail


sdairs:
## Problem

#320 restructured `main` so every invocation falls through to a single `process::exit` preceded by `telemetry::finalize()`, with the hidden `telemetry send` child as the sole intended early exit. Three pre-existing paths still bypass that tail (found by the #320 adversarial review):

- `crates/clickhousectl/src/local/mod.rs:489` — foreground `local server start` exits with the server child's exit code
- `crates/clickhousectl/src/local/docker.rs:389` and `:522` — docker exec paths exit with the container command's exit code

These invocations produce no telemetry event when they end with a non-zero child exit code, and they skip the update-notice tail.

## Design question

All three deliberately pass an arbitrary child exit code through to the shell, which `Error::exit_code()` (gh-style closed set: 0/1/2/4) cannot represent. Fixing this means either an `Error` variant carrying a raw exit code (and deciding how that interacts with the documented gh-style codes), or a dedicated "exit code override" channel from handlers back to `main`'s tail.

## Non-goal

The success paths of these commands (child exits 0) already flow through the tail and are counted normally; only the non-zero passthrough is affected.

Issue #319, opened 2026-07-29
Redefine exit codes: move 'cancelled' off 2 to resolve clash with clap usage errors


sdairs:
## Context

The CLI documents gh-style exit codes (help text, CLAUDE.md): `0` success, `1` error, `2` cancelled, `4` auth required, implemented in `Error::exit_code()` (`crates/clickhousectl/src/error.rs`).

Independently, clap exits with its own hardcoded `2` for all usage errors (unknown subcommand, unknown flag, missing required arg). So `$? == 2` is ambiguous: "user cancelled a prompt" vs "malformed command line". gh itself does not have this clash — Cobra exits `1` on usage errors, so gh's `2` unambiguously means cancelled. We inherited gh's table but clap's usage-error behavior.

This was discovered during the telemetry-coverage work (follow-up to #283): the decision there is to **allow the clash** — parse errors keep clap's exit `2`, and the telemetry payload's `outcome` field disambiguates in analytics. This issue is the tailored fix for the shell-visible ambiguity.

## Decision

Keep clap's `2` = usage error (don't fight clap's convention; it matches argparse and much of the ecosystem). Move **cancelled** to a different code instead — proposal: `3` (currently unused; `4` stays auth required).

Risk is effectively zero today: `Error::Cancelled` is declared with `#[allow(dead_code)]` and never constructed anywhere in the codebase, so no invocation currently exits with a "cancelled" `2`. This is still a documented-contract change, so it rides a version bump and gets called out prominently.

## Scope

- `Error::exit_code()` mapping + doc comment in `crates/clickhousectl/src/error.rs`, and its tests
- Help text stating the convention (`after_help` for the cloud command in `crates/clickhousectl/src/cloud/cli.rs`)
- `CLAUDE.md` (exit-code convention paragraph)
- `README.md` if it states exit codes
- Telemetry docs page wording if it references exit-code meanings
- Changelog/release notes: explicit breaking-change entry

## Acceptance

- Exit code `2` means exactly one thing (usage error, from clap); cancelled has its own code, documented consistently everywhere the convention is stated.

PR #322, opened 2026-07-29
Telemetry: first-run notice and events cover every invocation

### LINKED ISSUE #320: Telemetry: first-run notice and events must cover every invocation (bare, --help, --version, parse errors)
## Problem

Telemetry (#283) has exactly one hook: `telemetry::finalize()` at the bottom of `main()`. It is only reached when clap parsing succeeds. Every other invocation exits before capture (`crates/clickhousectl/src/main.rs:41-64`):

| Invocation | clap outcome | Today |
|---|---|---|
| `clickhousectl` (bare) | missing-subcommand usage error | `e.exit()` before telemetry |
| `clickhousectl --help` (any level) | `DisplayHelp` | prints help, `process::exit(0)` before telemetry |
| `clickhousectl --version` | `DisplayVersion` | same |
| `clickhousectl hello` (typo) | `InvalidSubcommand` | `e.exit()` before telemetry |

This was originally deliberate ("a mistyped invocation never produces an event"), but it is the wrong call:

- No first-run notice is shown, and the consent marker is never written, for a user whose first-ever touch is bare/`--help` — their consent clock never starts until their first valid command.
- Failed invocations are exactly the signal we want: agents hallucinating commands, users fumbling the tree.

**Requirement: any invocation of the binary — bare, help, version, typo, anything — shows the first-run notice on first run and produces a telemetry event on subsequent runs (subject to the existing consent state machine, which is unchanged).**

### 1. Single-exit `main`

Restructure `main` so there is exactly one `std::process::exit`, preceded by exactly one `finalize()` call. Parse `Err` branches stop calling `e.exit()`/`process::exit` inline; they `e.print()` (clap keeps formatting/color), do their existing update-notice work, and fall through as `(invocation, exit_code)` data to the common tail. The structural invariant — you cannot add an exit path without visibly bypassing the tail — is the maintainability goal, not per-branch "remember to call finalize".

Sole exemption: the hidden `telemetry send` child keeps exiting early (a send must never trigger another send).

Parse-error paths must not spawn the update-cache refresh (matches today; a typo shouldn't cause network activity beyond the consented send).

### 2. Definition-anchored lossy capture for failed parses

Failed parses have no `ArgMatches`, so the existing `capture()` cannot run. Add `capture_lossy(root_cmd, argv, error_kind)`: walk argv tokens against the clap `Command` tree —

- token matches a defined subcommand name/alias → descend, record the **definition's** name
- token is `--long` matching a defined flag (or `-h`/`-V`) → record the **definition's** long name; if that arg takes a value (known from the `Arg` def), skip the next token (prevents a flag value that equals a sibling subcommand name being misrecorded as path)
- token matches nothing → **stop**; the unmatched token is never recorded

Privacy holds by construction: every recorded string is owned by the clap definitions, matched by equality — argv slices never enter the payload. Same "structurally impossible to leak values" guarantee as `capture()`, just lossy (valid prefix only). Covers all cases uniformly: bare → `command: ""`; `cloud --help` → `command: "cloud", flags: ["help"]`; `cloud service craete` → `command: "cloud service"` + error kind.

The success path keeps the existing exact `capture()`.

### 3. Payload: `outcome` + `suggestion`

- `outcome`: `"ok" | "help" | "version" | "invalid_subcommand" | "unknown_argument" | "missing_subcommand" | "missing_required" | "invalid_value" | "other_parse_error"` — straight mapping from clap's closed `ErrorKind` enum; can never carry user data.
- `suggestion: Option<String>`: clap's "did you mean" (`ContextKind::SuggestedSubcommand`/`SuggestedArg`). The suggestion is computed **from the definition set**, so it is definition-derived and safe.

**Decided: do NOT record the raw unmatched token.** Knowing what agents hallucinate is valuable, but a typo'd invocation is indistinguishable from a secret pasted into the wrong window, and the published notice/docs promise "never values or arguments". Error kind + valid prefix + suggestion + the existing `is_agent`/`agent` fields are the safe proxy. Default to safety; may revisit later with an explicit notice/docs change.

The ingest worker already accepts empty `command`; it gains `outcome` and `suggestion`. Deploy worker change before the CLI release.

### 4. Exit codes: allow the clash

clap usage errors keep clap's exit `2`, which numerically collides with the documented gh-style `2` = cancelled. **Decided: this change does not touch exit codes**; the payload's `outcome` field fully disambiguates in analytics. The shell-visible ambiguity is fixed separately in #319 (move "cancelled" off 2; `Error::Cancelled` is currently never constructed, so practical risk is nil). Help/version remain exit `0`.

### 5. Snapshot the executable path at startup (Linux post-self-update fix)

`spawn_send_child` currently resolves `std::env::current_exe()` at the **end** of the process. After a successful `clickhousectl update`, the binary has been atomically replaced; on Linux, `/proc/self/exe` then resolves to `.../clickhousectl (deleted)`, the spawn fails silently, and the event for a successful self-update is dropped (Linux only; macOS returns the path string, which now points at the new binary and works). This bug exists today, independent of the coverage problem.

Fix: `static EXE_PATH: OnceLock<PathBuf>` in `telemetry.rs`, populated by an explicit `telemetry::init()` at the top of `main` — same eager-snapshot pattern as `dotenv::init()`. At startup the path is always clean; after an update it names the new binary, which is valid to spawn. If `current_exe()` fails at startup the lock stays empty and the send is skipped, matching today's failure mode.

Companion contract: the child may be a **newer version** than the parent that built the payload, so the hidden `telemetry send` subcommand (payload via env var) is a stable cross-version interface — must never be removed or renamed. Pin with a comment in `cli.rs`.

### Non-changes

- Consent state machine (`decide()`, `telemetry.json`, DO_NOT_TRACK, fail-open-to-silent) is untouched; Missing→Notice / Enabled→Send handle the new paths automatically.
- `clickhousectl update` already flows through the normal tail (no early exit, no re-exec) — no changes beyond §5.
- `--version`'s forced update-cache refresh (blocking, up to 10s) is pre-existing behavior and out of scope.

## Tests

- Unit tests for `capture_lossy` in `telemetry.rs`: bare, root help, nested help, typo'd subcommand, unknown flag, flag-value token equal to a subcommand name, `-h`/`-V` shorts, `--` handling; plus a hostile-argv "payload never contains SECRET" test mirroring the existing one.
- Subprocess tests (pattern of `tests/cli_request_shape_test.rs`, using `CHCTL_TELEMETRY_DEBUG` + temp `HOME`): first run of `--help` prints the notice and writes the marker; second run of `clickhousectl hello` emits a payload with `outcome: "invalid_subcommand"` and no raw token. These pin the "any invocation counts" invariant end-to-end.

## Docs

- Update `telemetry.rs` module docs (the "mistyped invocation never produces an event" paragraph inverts).
- Update the telemetry docs page to state that help/version/failed invocations are recorded as such — the notice text ("command name, flag names, never values or arguments") stays true.

## Rollout

1. Worker: accept `outcome`/`suggestion` (backward compatible).
2. CLI change (this issue).
3. Docs page update.

Follow-up: #319 (exit-code redefinition).

sdairs:
Closes #320

Based on `main`, deliberately **not** stacked on #316: the two changesets are disjoint (#316 touches cloud models/commands and appends to `cli_request_shape_test.rs`; this PR touches `main.rs`, `telemetry.rs`, `src/cli.rs`, `tests/telemetry_test.rs`), so they merge cleanly in either order.

## What

Any invocation of the binary — bare, `--help`, `--version`, mistyped commands — now shows the first-run notice on first run and produces a telemetry event on subsequent runs, under the unchanged consent state machine.

- **Single-exit `main` (#320 §1):** exactly one `process::exit`, preceded by exactly one `finalize()`. Parse `Err` branches `e.print()` (clap keeps formatting), keep their existing help/version update-notice behavior, and fall through as `(invocation, exit_code)`. Usage errors keep clap's exit 2 and spawn no update-cache refresh. The hidden `telemetry send` child stays the sole intended early exit.
- **Definition-anchored lossy capture (§2):** `capture_lossy(root, argv, error)` walks argv against the built clap `Command` tree — subcommand tokens descend and record the *definition's* name, `--long` tokens resolve to defined flags (value-taking flags consume the next token, so a value equal to a sibling subcommand name is never misrecorded), and the walk stops at the first unmatched token. Every recorded string is owned by the clap definitions; argv slices never enter the payload.
- **Payload (§3):** new `outcome` (closed vocabulary: `ok | help | version | invalid_subcommand | unknown_argument | missing_subcommand | missing_required | invalid_value | other_parse_error`) and `suggestion` (clap's definition-derived did-you-mean). **The raw unmatched token is deliberately never recorded** — a typo is indistinguishable from a pasted secret.
- **Exit codes (§4):** untouched; the numeric 2-collision with "cancelled" is disambiguated by `outcome` in analytics and is #319's to fix shell-side.
- **Exe-path snapshot (§5):** `telemetry::init()` at the top of `main` snapshots `current_exe()` into a `OnceLock`, fixing the Linux post-self-update dropped event; the `telemetry send` child interface is pinned as stable cross-version in `cli.rs`.

## Verification

- Unit tests for `capture_lossy`: bare, root/nested help, `-h`/`-V`, typo + suggestion, unknown flag, flag-value-equals-subcommand-name, inline `=` values, `--` handling, hostile-argv `SECRET` never on the wire.
- Subprocess e2e (wiremock + temp `HOME`): first `--help` prints the notice and writes the marker; bare → `missing_subcommand`/exit 2; hallucinated subcommand → `invalid_subcommand` with no raw token; typo → `suggestion: "service"`.
- Gates green per commit: `cargo test -p clickhousectl` (429 unit + 18 e2e), clippy `-D warnings`, `cargo fmt --check`, `cargo check --workspace --all-features`, and the full suite with telemetry compiled out (`--no-default-features`).
- A fresh-context adversarial review probed hostile argv shapes (non-UTF-8 tokens, `--flag=SECRET`, closed pipes) and audited clap 4.6's suggestion contexts (`SuggestedSubcommand`/`SuggestedArg` are definition-derived; the input-embedding `Suggested*` kinds are not read). Its findings are fixed in the final commit: a closed stderr no longer turns exit 2 into a panic (writes are swallowed like clap's own `Error::exit`), and the pre-existing child-exit-code passthrough exits it spotted are acknowledged and tracked as #321.

## Rollout

The ingest worker must accept `outcome`/`suggestion` (backward compatible — deploy before the next CLI release). The docs page update ('help/version/failed invocations are recorded as such') is step 3; the notice text stays true and unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -889,14 +889,17 @@ The CLI checks for updates in the background (at most once per 24 hours) and cac
 
 Each event contains exactly:
 
-- the command path (e.g. `local start`)
+- the command path (e.g. `local start`) — for an invocation that fails to parse, the longest *valid* prefix of what was typed; a mistyped or unrecognized token is never recorded
 - the **names** of the flags passed (e.g. `json`, `org-id`) — never flag values, never positional arguments
-- the exit code (`gh`-style: 0 success, 1 error, 2 cancelled, 4 auth required)
+- how the invocation ended: `ok`, `help`, `version`, or a parse-error kind such as `invalid_subcommand` — plus, for typos, clap's "did you mean" suggestion (always the name of a real command or flag, never your input)
+- the exit code (`gh`-style: 0 success, 1 error, 2 cancelled, 4 auth required; parse errors keep clap's exit 2)
 - the CLI version, OS, and architecture
 - whether it ran in CI (`CI` env var)
 - whether it ran under a detected coding agent, and if so which one (e.g. `claude-code`)
 
-There is no install ID, no device ID, and no fingerprinting of any kind. The payload is built from the clap command definitions rather than the raw command line, so leaking an argument value is structurally impossible — the code that builds the event has no access to values at all.
+Every invocation counts, including `--help`, `--version`, a bare `clickhousectl`, and mistyped commands — failed invocations are recorded as such (that is how we find the confusing corners of the CLI), subject to the same consent flow as everything else.
+
+There is no install ID, no device ID, and no fingerprinting of any kind. The payload is built from the clap command definitions rather than the raw command line, so leaking an argument value is structurally impossible — the code that builds the event has no access to values at all. That holds for failed parses too: they are reconstructed by matching argv tokens against the defined command tree, so every recorded string is one the CLI itself defines.
 
 Nothing is ever sent before you have seen the notice or explicitly enabled telemetry with `clickhousectl telemetry enable`: the first run prints a one-time notice to stderr, records that it was shown in `~/.clickhouse/telemetry.json`, and sends nothing. Sending starts from the following run — or immediately if you opt in by running `telemetry enable`, which is explicit consent and skips the notice. The send happens in a short-lived detached process, so command latency is unaffected even when the endpoint is unreachable.
 
--- a/README.md
+++ b/README.md
@@ -889,17 +889,14 @@ The CLI checks for updates in the background (at most once per 24 hours) and cac
 
 Each event contains exactly:
 
-- the command path (e.g. `local start`) — for an invocation that fails to parse, the longest *valid* prefix of what was typed; a mistyped or unrecognized token is never recorded
+- the command path (e.g. `local start`)
 - the **names** of the flags passed (e.g. `json`, `org-id`) — never flag values, never positional arguments
-- how the invocation ended: for dispatched invocations `ok`, `error`, `cancelled`, `auth_required`, or `exec` (the CLI handed the process over to another program, e.g. `local client`; its exit status is unknown); for invocations that fail to parse, `help`, `version`, or a parse-error kind such as `invalid_subcommand` — plus, for typos, clap's "did you mean" suggestion (always the name of a real command or flag, never your input)
-- the exit code (`gh`-style: 0 success, 1 error, 2 cancelled, 4 auth required; parse errors keep clap's exit 2; a fixed 0 for `exec`, where the real exit code is not knowable)
+- how the invocation ended and its exit code
 - the CLI version, OS, and architecture
 - whether it ran in CI (`CI` env var)
 - whether it ran under a detected coding agent, and if so which one (e.g. `claude-code`)
 
-Every invocation counts, including `--help`, `--version`, a bare `clickhousectl`, and mistyped commands — failed invocations are recorded as such (that is how we find the confusing corners of the CLI), subject to the same consent flow as everything else. (One known gap: a few local commands that mirror a child process's exit code directly skip the event when the child exits non-zero; tracked in [#321](https://github.com/ClickHouse/clickhousectl/issues/321).)
-
-There is no install ID, no device ID, and no fingerprinting of any kind. The payload is built from the clap command definitions rather than the raw command line, so leaking an argument value is structurally impossible — the code that builds the event has no access to values at all. That holds for failed parses too: they are reconstructed by matching argv tokens against the defined command tree, so every recorded string is one the CLI itself defines.
+There is no install ID, no device ID, and no fingerprinting of any kind. The payload is built from the clap command definitions rather than the raw command line, so leaking an argument value is structurally impossible — the code that builds the event has no access to values at all.
 
 Nothing is ever sent before you have seen the notice or explicitly enabled telemetry with `clickhousectl telemetry enable`: the first run prints a one-time notice to stderr, records that it was shown in `~/.clickhouse/telemetry.json`, and sends nothing. Sending starts from the following run — or immediately if you opt in by running `telemetry enable`, which is explicit consent and skips the notice. The send happens in a short-lived detached process, so command latency is unaffected even when the endpoint is unreachable.
 
--- a/README.md
+++ b/README.md
@@ -885,7 +885,7 @@ The CLI checks for updates in the background (at most once per 24 hours) and cac
 
 ## Telemetry
 
-`clickhousectl` collects anonymous usage data to help us understand which commands matter and improve the CLI. Full details: <https://clickhouse.com/docs/interfaces/cli#telemetry>.
+`clickhousectl` collects anonymous usage data to help us understand which commands matter and improve the CLI. Full details: <https://clickhouse.com/docs/concepts/features/interfaces/cli#telemetry>.
 
 Each event contains exactly:
 
@@ -897,7 +897,7 @@ Each event contains exactly:
 - whether it ran in CI (`CI` env var)
 - whether it ran under a detected coding agent, and if so which one (e.g. `claude-code`)
 
-Every invocation counts, including `--help`, `--version`, a bare `clickhousectl`, and mistyped commands — failed invocations are recorded as such (that is how we find the confusing corners of the CLI), subject to the same consent flow as everything else.
+Every invocation counts, including `--help`, `--version`, a bare `clickhousectl`, and mistyped commands — failed invocations are recorded as such (that is how we find the confusing corners of the CLI), subject to the same consent flow as everything else. (One known gap: a few local commands that mirror a child process's exit code directly skip the event when the child exits non-zero; tracked in [#321](https://github.com/ClickHouse/clickhousectl/issues/321).)
 
 There is no install ID, no device ID, and no fingerprinting of any kind. The payload is built from the clap command definitions rather than the raw command line, so leaking an argument value is structurally impossible — the code that builds the event has no access to values at all. That holds for failed parses too: they are reconstructed by matching argv tokens against the defined command tree, so every recorded string is one the CLI itself defines.
 
--- a/README.md
+++ b/README.md
@@ -891,7 +891,7 @@ Each event contains exactly:
 
 - the command path (e.g. `local start`) — for an invocation that fails to parse, the longest *valid* prefix of what was typed; a mistyped or unrecognized token is never recorded
 - the **names** of the flags passed (e.g. `json`, `org-id`) — never flag values, never positional arguments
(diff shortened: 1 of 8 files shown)


sdairs:
**Finding: short flags (other than `-h`/`-V`) are invisible to `capture_lossy`.**

The lossy walker in `telemetry.rs` resolves `--long` flags via `get_long()` and hardcodes the two implicit shorts (`-h`/`-V`), but any other single-dash token falls through to the `else { break }` — the walk stops and nothing after it is recorded.

The CLI has real short flags today: `-v` (`local server start/postgres … --version`), `-F` (`--foreground`), `-p` (`--port`), `-q` (`--query`), `-e` (`--env`). So e.g. a failing `clickhousectl local client -q "select 1"`-style invocation would record only `command: "local client"` and lose everything after the short flag.

This is **privacy-safe** (fails toward recording less, never toward leaking a value), and it diverges from `capture`, which resolves `-h`/`-V` via normal definition lookup rather than special-casing. Not a blocker for this PR — but worth a follow-up to resolve shorts via `get_short()` the same way longs are resolved via `get_long()`, or at minimum a comment at the `else { break }` acknowledging that single-dash tokens stop the walk. Right now that limitation is implicit and only discoverable by reading the walker closely.

sdairs:
**The docs link the notice points users to has moved and needs updating.**

`DOCS_URL` (`telemetry.rs:42`), the `after_help` text on the `telemetry` command (`cli.rs:87`), and the README telemetry section (line 888) all still point to:

```
https://clickhouse.com/docs/interfaces/cli#telemetry
```

That URL now redirects; the canonical location is:

```
https://clickhouse.com/docs/concepts/features/interfaces/cli
```

Since this PR makes the first-run notice fire on *every* first invocation (bare, `--help`, typos — not just successful commands), far more users will see and follow that link than before. Worth updating all three spots to the new URL in this PR rather than shipping a notice that sends fresh users through a redirect.

sdairs:
## Structural review (no runtime checks)

Overall, the core direction is good: a common telemetry tail is the right shape, refusing to record unmatched tokens is the right privacy choice, and eagerly snapshotting `current_exe()` cleanly addresses the Linux self-update case. I would still request changes before approval because the implementation does not yet satisfy its central “every invocation reaches one telemetry tail” claim.

### Findings

1. **[P1] Successful `exec()` handoffs bypass telemetry entirely.**

   `run()` is not guaranteed to return to `main.rs`:

   - `crates/clickhousectl/src/local/mod.rs:299` uses `CommandExt::exec()` for `local client`.
   - `crates/clickhousectl/src/local/postgres.rs:626` uses it for the host-`psql` path.

   On success, the clickhousectl process image is replaced, so `telemetry::finalize()` is never reached. Those invocations get no first-run notice, marker, or event. These paths are also absent from follow-up #321, which only lists the three nested `process::exit` paths. Either finalize before an explicit handoff with a distinct outcome, or spawn/wait and propagate the child status through a typed top-level result.

2. **[P2] `capture_lossy` implements only a subset of clap’s grammar.**

   The walker handles canonical long names plus hard-coded `-h`/`-V`, but not long aliases such as `--fg` and `--config-file`, real short options such as `-v`, `-q`, and `-e`, short clusters, or attached short values. It stops at the first such token, so valid flags—and sometimes later useful context—are omitted. Bugbot already noted aliases; short options make this a broader parser-drift problem. The matcher should cover all spellings and value arity exposed by `Arg`, ideally with coverage generated from the built command tree.

3. **[P2] Suggestions are not definition-anchored by this code.**

   `suggestion_for_error()` directly clones strings from clap’s error context. They are safe with the currently locked clap implementation, but the public “never your input” promise depends on clap internals rather than an invariant enforced here. Validate any suggestion against the relevant command/argument definitions and emit the canonical definition-owned spelling. Also, clap 4.6 orders `SuggestedSubcommand` candidates from least to most similar, so `first()` records the weakest candidate; use the last candidate or model suggestions as a list.

4. **[P2] `outcome: "ok"` means “parsed”, not “succeeded”.**

   `capture()` runs before dispatch and hard-codes `ok`, so handler failures, authentication errors, and cancellations produce `outcome: "ok"` alongside a non-zero exit code. That conflicts with the README description of this field as “how the invocation ended” and will be easy to misuse in analytics. Prefer `parsed`/`dispatched`, or separate parse outcome from execution outcome.

5. **[P2] Cached update notices can still bypass finalization on closed stderr.**

   `main.rs` now swallows clap and telemetry-notice write failures, but `update::print_cached_update_notice()` still uses `eprintln!`, which panics if stderr is closed. With a cached newer version, `--help` or `--version` can therefore panic before capture/finalize. The update notice should use the same non-panicking write pattern.

### Architectural and rollout notes

- The “single-exit invariant” is currently a comment-level convention: lower layers still contain three acknowledged `process::exit` calls and two overlooked `exec()` handoffs. A typed `RunDisposition`/`ExitStatus` returned from handlers would make the boundary real and keep process termination in one place.
- The worker rollout must accept both old payloads where `outcome`/`suggestion` are absent and new payloads where `suggestion` is explicitly `null`. Delivery failures are intentionally silent, so a schema mismatch will silently drop new-version events.
- The external telemetry documentation should ship before or with the CLI release because disclosure now materially includes help, version, and failed invocations.
- The hidden cross-version `telemetry send` contract is protected only by a comment. A compatibility test or pre-parse child-mode guard would make self-update safer.
- README’s “Every invocation counts” statement is currently false because of the known exits and the two `exec()` paths above.

**Recommendation:** fix the `exec()` coverage before merge, and either fix the remaining P2 items or explicitly narrow the claims/docs and track them as concrete follow-ups.

iskakaushik (approved):

PR #345, opened 2026-07-31
Report invalid local versions directly

### LINKED ISSUE #340: `local use` with an unparseable version prints a double-wrapped error message
Found during 0.4.0 RC local-command testing (2026-07-30).

## Observed

```
$ chctl local use not.a.version
Error: No matching version found for: invalid version 'not.a.version': all parts must be numeric
(exit 1)
```

Two error layers are concatenated: the outer "No matching version found for:" (which expects a version string after the colon) wraps the inner parse error "invalid version '...': all parts must be numeric". The result reads as garbled — "no matching version found for: invalid version" is not a sentence.

## Expected

A parse failure should short-circuit before the lookup path adds its wrapper:

```
Error: invalid version 'not.a.version': all parts must be numeric
```

Purely an error-plumbing fix in the version-resolution path (the inner error is already the right message).

sdairs:
## Summary

- distinguish invalid version syntax from version lookup misses
- render parser diagnostics directly instead of prefixing them with `No matching version found for:`
- add focused unit coverage for invalid spec shapes and a subprocess regression test for `local use`

## Why

`parse_version_spec` used `Error::NoMatchingVersion` for both parser failures and genuine resolution misses. That error variant adds a lookup-specific prefix, so an input such as `not.a.version` produced a double-wrapped, garbled message.

## Impact

`clickhousectl local use not.a.version` now reports:

```
Error: invalid version 'not.a.version': all parts must be numeric
```

Genuine lookup misses continue to use the existing `No matching version found for:` message.

## Stack

This PR is stacked on #343 and should be reviewed as the delta from `codex/issue-323-openapi-drift`.

Closes #340.

## Validation

- `cargo fmt --all --check`

### DIFF
--- a/crates/clickhousectl/src/error.rs
+++ b/crates/clickhousectl/src/error.rs
@@ -42,6 +42,9 @@ pub enum Error {
     #[error("No matching version found for: {0}")]
     NoMatchingVersion(String),
 
+    #[error("{0}")]
+    InvalidVersion(String),
+
     #[error("Failed to execute ClickHouse: {0}")]
     Exec(String),
 
--- a/crates/clickhousectl/src/version_manager/spec.rs
+++ b/crates/clickhousectl/src/version_manager/spec.rs
@@ -34,7 +34,7 @@ pub fn parse_version_spec(input: &str) -> Result<VersionSpec> {
     let input = input.trim();
 
     if input.is_empty() {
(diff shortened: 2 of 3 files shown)


iskakaushik (approved):
LGTM

PR #348, opened 2026-07-31
Batch macOS process discovery

### LINKED ISSUE #329: Server discovery runs one `lsof` per clickhouse PID — ~29s command stall on machines with many clickhouse-local/chDB processes
Found during 0.4.0 RC local-command testing.

## Observed

`discover_clickhouse_processes()` (`crates/clickhousectl/src/local/discovery.rs`) does `pgrep -x clickhouse`, then for **each** PID shells out to `lsof -a -d cwd -Fn -p <pid>` to resolve its cwd. On macOS each `lsof` call costs ~0.27s, so total discovery time is linear in the number of processes *named* `clickhouse` on the machine — including ones the CLI didn't start.

On a machine with 106 such processes (chDB / clickhouse-local children spawned by an AI-agent app), every command that touches discovery stalled ~29s:

```
$ time chctl local server list        # zero project servers
{"servers": [], "total_servers": 0, "total_running_servers": 0}
1.65s user 9.11s system 37% cpu 28.847 total
```

Affected: `server list`, `server start`, `server stop`/`stop-all`, and named-server `client` lookup. `client --host/--port` is unaffected (skips discovery). Not telemetry-related (reproduces with `DO_NOT_TRACK=1`).

After killing the stray processes, the same commands take 0.06–0.36s — confirming the per-PID `lsof` loop is the entire cost.

This setup isn't exotic: chDB and clickhouse-local are exactly what coding agents spawn, so agent-heavy machines are the ones that hit it.

## Fix candidates

1. **Batch the lsof call**: `lsof` accepts a comma-separated PID list (`lsof -a -d cwd -Fn -p pid1,pid2,...`), and `-Fn` output includes `p<pid>` markers to attribute each cwd. One ~0.3s call regardless of N.
2. Or drop the subprocess entirely on macOS and use libproc (`proc_pidinfo`/`PROC_PIDVNODEPATHINFO`) for the cwd, mirroring the existing `/proc/<pid>/cwd` fast path on Linux.

Option 1 is the minimal change; option 2 also removes the dependency on `lsof` being present/fast.

sdairs:
## Summary

- resolve all ClickHouse process working directories with one batched `lsof` call on macOS
- parse `lsof -Fn` PID/path records and retain valid records if another process exits during discovery
- preserve the Linux `/proc/<pid>/cwd` fast path and add focused parser coverage

## Why

Process discovery previously launched one `lsof` subprocess per process named `clickhouse`. On agent-heavy machines with many clickhouse-local or chDB processes, that made local commands stall for tens of seconds even when none of those processes belonged to the current project.

## Impact

macOS discovery now pays the `lsof` startup cost once regardless of the number of ClickHouse processes. The CLI surface and Linux behavior are unchanged.

## Stack

This PR is stacked on #347 and should be reviewed as the delta from `codex/issue-337-org-autodetect`.

Closes #329.

## Validation

- `cargo fmt --all --check`
- `cargo test -p clickhousectl` (570 tests)
- verified batched two-PID `lsof -Fn` output on macOS

### DIFF
--- a/crates/clickhousectl/src/local/discovery.rs
+++ b/crates/clickhousectl/src/local/discovery.rs
@@ -4,7 +4,7 @@
 //! and command-line arguments to recover server metadata (project root, name,
 //! ports, version). Used for orphaned server recovery and global server listing.
 
-use std::process::Command;
+use std::{collections::HashMap, process::Command};
 
 /// A ClickHouse process discovered via OS-level process inspection.
 #[derive(Debug, Clone)]
@@ -23,10 +23,15 @@ pub struct DiscoveredProcess {
 /// meaning they were started by this CLI. Other ClickHouse processes are ignored.
 pub fn discover_clickhouse_processes() -> Vec<DiscoveredProcess> {
     let pids = find_clickhouse_pids();
+    let cwds = get_process_cwds(&pids);
     let mut discovered = Vec::new();
 
     for pid in pids {
-        if let Some(proc) = inspect_process(pid) {
+        let Some(cwd) = cwds.get(&pid) else {
+            continue;
+        };
+
+        if let Some(proc) = inspect_process(pid, cwd) {
             discovered.push(proc);
         }
     }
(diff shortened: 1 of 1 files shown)


iskakaushik (approved):
LGTM

Issue #328, opened 2026-07-30
Output polish: fix misleading and duplicate install confirmations


sdairs:
Found during 0.4.0 RC local-command testing. The Docker pull-output problem originally grouped here is now tracked separately in #353.

## 1. No-op install prints "Installed version X"

```
$ chctl local install 26.2          # 26.2.16.5 already installed
ClickHouse 26.2 is already installed as 26.2.16.5
Use --force to re-download the latest build
Installed version 26.2.16.5
```

Nothing was installed; the final line contradicts the hint above it. It should say something like `Using existing 26.2.16.5` (or print nothing beyond the hint).

## 2. Real install prints two confirmation lines

```
$ chctl local install 26.2 --force
Resolving 26.2...
Downloading ClickHouse 26.2...
Detecting version...
Installed ClickHouse 26.2.20.4
Installed version 26.2.20.4
```

The last two lines are duplicates from different layers; one should go.

sdairs:
Implemented in #354 (stacked on #352).

PR #349, opened 2026-07-31
Make repeated exact installs idempotent

### LINKED ISSUE #339: `local install` of an already-installed version: exact spelling errors (rc=1), partial spelling no-ops (rc=0)
Found during 0.4.0 RC local-command testing (2026-07-30); the copy half of this is #328, but the behavioral half was never filed.

## Observed

Installing a version that is already installed behaves oppositely depending on how you spell it:

```
$ chctl local install 26.2.16.5     # exact, already installed
Error: Version 26.2.16.5 is already installed
(exit 1)

$ chctl local install 26.2          # partial, resolves to 26.2.16.5, already installed
ClickHouse 26.2 is already installed as 26.2.16.5
Use --force to re-download the latest build
(exit 0)
```

Same user intent — "make sure this version is installed" — but the exact spelling is a hard error while the partial spelling is a friendly no-op.

## Why it matters

- Idempotent-install scripts (`chctl local install X && chctl local server start --version X`) break only when X is exact.
- The rc=1 path records the command as a failure in usage telemetry (#320's outcome classification), inflating error rates for what is arguably a success.

## Expected

Both paths should behave like the partial one: exit 0, say it's already installed, hint `--force`. (With #328's copy fix, neither path should claim "Installed" when nothing happened.)

sdairs:
## Summary

- treat an exact ClickHouse version that is already installed as a successful no-op
- emit the same already-installed message and `--force` hint used for partial version specs
- cover both exact and partial local-first installs with subprocess regression tests

## Why

The local-first install path special-cased exact version specs by returning `VersionAlreadyInstalled`, while partial specs returned the matching installed version successfully. That made equivalent idempotent install requests produce different exit codes and telemetry outcomes.

## Impact

Repeated exact-version installs now exit 0 without resolving or downloading anything, so scripts can safely use `local install` to ensure a version is present. The separate output-copy cleanup remains tracked by #328.

## Stack

This PR is stacked on #348 and should be reviewed as the delta from `codex/issue-329-batch-lsof`.

Closes #339.

## Validation

- `cargo fmt --all --check`
- `cargo test -p clickhousectl` (571 tests)

### DIFF
--- a/crates/clickhousectl/src/version_manager/install.rs
+++ b/crates/clickhousectl/src/version_manager/install.rs
@@ -9,17 +9,14 @@ use crate::version_manager::spec::VersionSpec;
 use std::os::unix::fs::PermissionsExt;
 
 /// Install a version spec, trying installed versions first before any remote call.
-/// Matches the UX of `install_resolved`'s post-resolve local check, but avoids the
-/// network round-trip when a local match exists.
+/// An installed match is a successful no-op regardless of whether the spec is
+/// exact or partial.
 pub async fn install_local_first(
     spec: &VersionSpec,
     platform: &Platform,
     force: bool,
 ) -> Result<String> {
     if !force && let Some(local) = try_resolve_local(spec) {
-        if matches!(spec, VersionSpec::Exact(_)) {
-            return Err(Error::VersionAlreadyInstalled(local));
-        }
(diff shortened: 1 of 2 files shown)


iskakaushik (approved):
LGTM

PR #351, opened 2026-08-01
Suggest --force when deleting a running service

### LINKED ISSUE #335: `service delete` on a running service should suggest --force
Found during 0.4.0 RC cloud-command testing (2026-07-30).

`cloud service delete` on a running service passes the API's CONFLICT text through without mentioning that the CLI already has the remedy:

```
$ chctl cloud service delete <running-id>
Error: CONFLICT: Only instance in one of the following states: 'provisioning','starting','awaking','idle','stopped','degraded','failed' can be terminated. Current state: 'running'
(exit 1)
```

`--force` exists precisely for this ("Stop the service first if it is running, then delete") and works well. When delete fails with this CONFLICT, append a hint:

```
Error: service is running and cannot be deleted. Use --force to stop it first, or `cloud service stop <id>`.
```

sdairs:
## Summary

- replace the raw running-service deletion conflict with an actionable `--force` hint
- include the exact `clickhousectl cloud service stop <id>` alternative in the error
- preserve unrelated conflicts and avoid suggesting `--force` when it was already passed
- add unit and subprocess regression coverage for the error path

## Why

`cloud service delete` currently passes through the Cloud API's `CONFLICT` response when the service is running, even though the CLI already supports the recovery path through `--force`. Users have to discover that remedy separately.

## Impact

Deleting a running service without `--force` now returns a concise error that explains both supported next steps. Other deletion failures retain their original API errors.

## Stack

This PR is stacked on #349 and should be reviewed as the delta from `codex/issue-339-idempotent-install`.

Closes #335.

## Validation

- `cargo fmt --all --check`

### DIFF
--- a/crates/clickhousectl/src/cloud/commands.rs
+++ b/crates/clickhousectl/src/cloud/commands.rs
@@ -1,4 +1,4 @@
-use crate::cloud::client::CloudClient;
+use crate::cloud::client::{CloudClient, CloudError};
 use crate::cloud::credentials;
 use crate::cloud::output::{ABSENT, or_absent, print_human};
 use clickhouse_cloud_api::models::{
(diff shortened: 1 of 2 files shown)


iskakaushik (approved):
LGTM

PR #436, opened 2026-08-20
Enforce global JSON query format conflict

### LINKED ISSUE #422: Global --json placement bypasses service query --format conflict
## Summary

`--json` is a global Cloud option and can be placed before or after nested subcommands, but its documented conflict with `service query --format` only works in some positions.

## Reproduction

This succeeds and emits CSV despite the explicit JSON request:

```console
clickhousectl cloud --json service query --id <service-id> --query 'SELECT 1' --format CSV
```

Moving the same global option after the nested command correctly produces a clap conflict:

```console
clickhousectl cloud service query --id <service-id> --query 'SELECT 1' --json --format CSV
error: the argument '--json' cannot be used with '--format <FORMAT>'
```

## Expected behavior

Both argument orderings should behave identically. Since help says `--json` cannot be combined with `--format`, both should fail before making a request.

This matters especially for agents, which naturally place global options immediately after `cloud` and may silently receive non-JSON output.

## Environment

Tested with `clickhousectl 0.4.1` built from PR #419 head `164d98154e77a491611c9a3b7de652d2b9caaffd` on macOS arm64.

sdairs:
## Summary

- reject explicit Cloud `--json` with service query `--format` in either argument order
- perform cross-level validation before runtime or network work
- preserve explicit formats when JSON mode comes only from coding-agent detection

## Testing

- `cargo fmt --all --check`
- focused subprocess tests for both flag orders and zero requests
- `git diff --check`

Closes #422

### DIFF
--- a/crates/clickhousectl/src/cloud/cli.rs
+++ b/crates/clickhousectl/src/cloud/cli.rs
@@ -44,6 +44,21 @@ pub struct CloudArgs {
     pub command: CloudCommands,
 }
 
+impl CloudArgs {
+    pub fn has_explicit_json_format_conflict(&self) -> bool {
+        self.json
+            && matches!(
+                &self.command,
+                CloudCommands::Service {
+                    command: ServiceCommands::Query {
+                        format: Some(_),
+                        ..
+                    }
+                }
+            )
+    }
+}
+
 #[derive(Subcommand)]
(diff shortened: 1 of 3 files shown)


sdairs:
bugbot review

iskakaushik (approved):

PR #441, opened 2026-08-20
Use concrete errors in Cloud handlers

### LINKED ISSUE #233: Cloud handlers: return a concrete Error instead of Box<dyn Error> to drop the exit-code downcast
Follow-up from PR #222.

## Background

PR #222 added `gh`-style exit codes. `CloudError` now carries a `kind` (`Auth` / `Generic`) so 401/403 and missing-creds map to exit `4`. But cloud command handlers (`cloud/commands.rs`, `cloud/postgres.rs`) and `run_postgres` return `Box<dyn std::error::Error>`, which **type-erases** the `CloudError`. To recover `kind` at the top level we currently downcast:

```rust
fn boxed_cloud_error_to_top_level(e: Box<dyn std::error::Error>) -> Error {
    match e.downcast::<cloud::CloudError>() {
        Ok(ce)     => cloud_error_to_top_level(*ce),    // kind recovered -> exit 4 possible
        Err(other) => Error::Cloud(other.to_string()),  // not a CloudError -> exit 1
    }
}
```

## Problem

The downcast only succeeds if the boxed concrete type is *exactly* `CloudError`. If any handler stringifies a `CloudError` before boxing — e.g. `client.foo().await.map_err(|e| format!("...: {e}"))?` instead of plain `?` — the downcast misses and exit `4` **silently** degrades to exit `1`. It still compiles, the message still prints, the process still exits non-zero; only the machine-readable auth signal is lost. Unit tests pin the router contract, but nothing structurally prevents a future handler from breaking it.

`Box<dyn Error>` was chosen for ergonomic `?` across heterogeneous error sources (CloudError, serde_json, uuid, chrono, io, plain validation strings) — that convenience is exactly what erases the type we now need to inspect.

## Proposed improvement

Have cloud handlers return a concrete, matchable error type instead of `Box<dyn Error>` — e.g. the top-level `Error` (or a dedicated cloud error enum) with `#[from]` impls for the parse/io/cloud sources. Then `kind` is carried as `Error::AuthRequired` vs `Error::Cloud` from the point of failure, `main` matches on it directly, and `boxed_cloud_error_to_top_level` + the downcast disappear entirely.

## Scope / cost

Touches the signature and error construction of every cloud handler in `cloud/commands.rs` and `cloud/postgres.rs`, plus `run_postgres`. Larger than #222, hence split out. No user-facing behavior change — purely making the exit-code routing structurally correct rather than downcast-dependent.

sdairs:
## Summary

- return concrete `CloudError` values throughout dispatched Cloud handlers and helpers
- remove boxed-error downcasting from the Cloud runtime boundary
- preserve auth classification and current messages through explicit source conversions

## Testing

- `cargo test -p clickhousectl --all-features`
- focused subprocess 401/403/500 exit-code tests
- `cargo clippy -p clickhousectl --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`

Closes #233

### DIFF
--- a/crates/clickhousectl/src/cloud/activity.rs
+++ b/crates/clickhousectl/src/cloud/activity.rs
@@ -1,4 +1,4 @@
-use crate::cloud::client::CloudClient;
+use crate::cloud::client::{CloudClient, Result as CloudResult};
 use crate::cloud::output::{or_absent, print_human};
 use crate::cloud::shared::{parse_date_only, resolve_org_id};
 use clap::Subcommand;
@@ -41,11 +41,7 @@ impl ActivityCommands {
     }
 }
 
-pub async fn run(
-    client: &CloudClient,
-    command: ActivityCommands,
-    json: bool,
-) -> Result<(), Box<dyn std::error::Error>> {
+pub async fn run(client: &CloudClient, command: ActivityCommands, json: bool) -> CloudResult<()> {
     match command {
         ActivityCommands::List {
             org_id,
@@ -74,7 +70,7 @@ async fn activity_list(
     from_date: Option<&str>,
     to_date: Option<&str>,
     json: bool,
-) -> Result<(), Box<dyn std::error::Error>> {
+) -> CloudResult<()> {
     let org_id = resolve_org_id(client, org_id).await?;
     let activities = client.list_activities(&org_id, from_date, to_date).await?;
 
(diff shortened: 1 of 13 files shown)


sdairs:
bugbot review

iskakaushik (approved):

PR #500, opened 2026-08-24
Allow direct local client connections to select a binary version

### LINKED ISSUE #469: Allow direct local client connections to select a binary version
## Problem

Direct `local client --host/--port` mode promises to connect to any ClickHouse server, but it can only launch the native client selected by `~/.clickhouse/default`.

On HEAD `93b3dd7`, an isolated HOME with one installed executable and no default file fails before connecting:

```console
$ clickhousectl local client --host remote
Error: No default version set. Run: clickhousectl local use <version>
```

Named mode does not have this requirement because it uses the version stored in server metadata. `local server start --version X` can therefore launch a managed server without changing the default, while direct mode cannot reuse that installed binary unless the user changes global selection.

## Expected behavior

Direct mode should offer an explicit client-binary selection path without requiring a global default change. An explicit `--version` is preferable to silently guessing among multiple installed versions.

## Acceptance criteria

- Direct mode accepts an explicit installed client version without changing the default.
- Behavior with no explicit version and zero, one, or multiple installed versions is deliberate and documented.
- A stale default and an explicit missing version produce actionable errors.
- Named mode continues to use the managed server's recorded version without requiring a default.
- Help explains the difference between connection selection (`--host`/`--port`) and local binary selection.
- Parser and subprocess tests cover the complete default/installed/explicit-version matrix.

sdairs:
## Summary

- add direct-only `local client --version <exact-installed-version>` binary selection without changing the configured default
- retain default-only resolution when direct `--version` is omitted, with no zero/one/multiple-install guessing and actionable missing/stale errors
- preserve named-server metadata version selection and document connection selection separately from local binary selection
- cover the parser and subprocess default/installed/explicit-version matrices

Closes #469

## Verification

- `cargo fmt --all --check`
- `cargo test -p clickhousectl` (756 passed)

## Stack

- Position: child 4 of 4 in gh-stack #489
- Base: #492 (`issue-466-local-client-selectors`)
- Parent PR is unchanged and should merge first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,12 @@ clickhousectl local client --name dev                # Connects to "dev" server
 clickhousectl local client --query "SHOW DATABASES"  # Run a query
 clickhousectl local client --queries-file schema.sql # Run queries from a file
 clickhousectl local client --host remote-host --port 9000  # Connect to a specific host/port
+clickhousectl local client --host remote-host --version 26.8.1.1760  # Use an exact installed client binary
 ```
+
+Named connections and direct connections select the client binary differently. A named connection uses the version recorded in the managed server's metadata, regardless of the global default. A direct connection (`--host`, `--port`, or both) uses the exact installed version passed with `--version`; if that flag is omitted, it uses the version selected by `local use`.
+
+Direct connections never infer a binary from the contents of `~/.clickhouse/versions`. With no valid default, omitting `--version` fails whether zero, one, or multiple versions are installed. Use `local list` to find an exact installed version, then pass it with `--version` or select it globally with `local use`. A missing explicit version and a default that points to a removed version both fail with repair instructions. Direct `--version` selection neither installs a binary nor changes the default.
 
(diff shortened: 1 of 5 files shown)


sdairs:
good but adds too much noise to agent context help. it is unlikely that thorough explanation of a --version flag is needed beyond the actual flag description

sdairs:
Addressed in [`30c6f10`](https://github.com/ClickHouse/clickhousectl/commit/30c6f105d05ac329c55cf9376b37c60872b74d3d): removed the redundant `--version` explanation from agent-context help, restored the existing concise connection/default wording, and retained the flag and its description. Verified with the focused help test, the local client selector integration suite, `cargo fmt --all --check`, `cargo check -p clickhousectl`, and clippy with warnings denied.

iskakaushik (approved):
LGTM

sdairs:
Superseded by #552 in stack #563.

PR #523, opened 2026-08-25
Make global ClickHouse symlink updates atomic

fallintoplace:
## Summary

- **Stage** the replacement symlink beside `~/.local/bin/clickhouse`
- **Rename** it over the old symlink atomically
- **Preserve** the old link when staging fails
- Keep regular-file and already-correct-link behavior unchanged

## Tests

- `cargo fmt --all --check`
- `cargo test -p clickhousectl --lib`
- `cargo test -p clickhousectl local::symlink::tests -- --nocapture`


sdairs (approved):
LGTM - thanks for contributing!

PR #481, opened 2026-08-24
Require an explicit Cloud integration decision

sdairs:
Closes #410

## Summary
- add a trusted default-branch controller that publishes `Cloud integration decision` only on the exact current PR head SHA
- accept successful selected suites, a successful no-suite plan that never entered the Cloud environment, or an audited maintainer override
- keep live execution on same-repository `pull_request` with the existing Dependabot exclusion and read-only token; the write-capable controller checks out only its own trusted revision and never executes PR code
- reject stale runs, unrelated labels, skipped admission, forks, Dependabot, incomplete job evidence, and non-PR Cloud runs
- require `maintain`/`admin`, exact SHA, actor-matching comment, timestamp, and reason or run link for the one-shot `skip-cloud-integration` override

## Stacked PR policy
Use SHA-bound overrides on lower PRs that link the successful manual `scope=all` top-stack run. A maintainer verifies each lower head is an ancestor of the tested top SHA and records a separate override on every lower exact SHA. The top run never silently covers lower checks.

## Verification
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` (68 passed)
- `cargo test -p clickhouse-cloud-api -p clickhouse-openapi-analyzer`
- `cargo clippy -p clickhouse-cloud-api -p clickhouse-openapi-analyzer --all-targets -- -D warnings`
- `cargo fmt --all --check`
- `actionlint` v1.7.10 on the new controller and changed test workflow
- `zizmor` v1.25.2 on the controller, live workflow, and changed test workflow (no findings)
- `git diff --check`

## Rollout
The `skip-cloud-integration` repository label has been created. Branch protection is intentionally unchanged until the controller is on `main`. After merge, add the required context without replacing existing checks:

```bash
gh api --method POST \
  repos/ClickHouse/clickhousectl/branches/main/protection/required_status_checks/contexts \
  -f 'contexts[]=Cloud integration decision'
```

The controller PR itself must merge before this one-time setting is activated.

PR #526, opened 2026-08-25
Isolate telemetry test mock endpoints

### LINKED ISSUE #525: Prevent first-run telemetry requests leaking between test mocks
## Problem

`first_run_writes_marker_prints_notice_sends_nothing` runs `local list` twice. The second invocation has enabled telemetry and starts a detached sender, but the test returns without waiting for that request. Its WireMock server can be dropped and its port reused by another parallel test before the child connects.

This makes unrelated no-send tests fail after receiving a payload for `command: local list`. Observed in, among others:

- https://github.com/ClickHouse/clickhousectl/actions/runs/32856844332/job/97830795286?pr=500
- https://github.com/ClickHouse/clickhousectl/actions/runs/32856016807

Issue #442 fixed the same lifecycle problem for `parent_never_waits_for_a_slow_endpoint`, but this second detached-child path remained.

## Acceptance criteria

- The first-run test waits until its enabled second invocation reaches its own mock before teardown.
- Every telemetry test that can spawn a detached sender either waits for its request or explicitly prevents sending.
- The telemetry subprocess suite passes repeatedly under normal parallel test execution.

sdairs:
## Summary

- give every telemetry test sandbox a unique ingest path instead of sharing `/v1/telemetry`
- filter received requests by that test-owned path so a delayed child hitting a reused port cannot satisfy or fail another test
- apply the same path isolation to the slow-endpoint case and add focused coverage for stale unmatched requests

Closes #525

## Verification

- `cargo test -p clickhousectl --test telemetry_test`
- telemetry test binary repeated 50 times (1,150 tests passed)

## Stack

- Position: child 33 of 33 in gh-stack #524
- Base: #481 (`issue-410-cloud-integration-decision`)
- Parent PR is unchanged and should merge first.

### DIFF
--- a/crates/clickhousectl/tests/telemetry_test.rs
+++ b/crates/clickhousectl/tests/telemetry_test.rs
@@ -34,20 +34,31 @@ fn clickhousectl_binary() -> PathBuf {
 struct Sandbox {
     home: tempfile::TempDir,
     mock: MockServer,
+    endpoint_path: String,
 }
 
 impl Sandbox {
     async fn new() -> Self {
+        let home = tempfile::tempdir().unwrap();
+        let endpoint_path = format!(
+            "/v1/telemetry/{}",
+            home.path().file_name().unwrap().to_string_lossy()
+        );
         let mock = MockServer::start().await;
         Mock::given(method("POST"))
-            .and(path("/v1/telemetry"))
+            .and(path(endpoint_path.clone()))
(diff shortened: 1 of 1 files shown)


iskakaushik (approved):
LGTM

sdairs:
Superseded by #531 in stack #563.

PR #530, opened 2026-08-26
Isolate subprocess tests from coding-agent environments

### LINKED ISSUE #445: Subprocess tests fail when run inside a coding agent: clear_agent_env misses Claude Code (and most other) detection markers
## Problem

`clear_agent_env` in `crates/clickhousectl/tests/cli_request_shape_test.rs` only removes `AGENT`, `AI_AGENT`, and the `OPENCODE*` variables before spawning the CLI binary. But `is-ai-agent` 0.5.0 detects agents via a much larger (private) marker table — for Claude Code alone: `CLAUDECODE`, `CLAUDE_CODE_ENTRYPOINT`, `CLAUDE_CODE_SESSION_ID`, `CLAUDE_CODE_EXECPATH` — plus markers for Cursor, Codex, Copilot, Gemini CLI, Amp, CodeBuddy, Goose, Replit, and a file signal (`/opt/.devin`).

So when the test suite runs inside Claude Code (or most other agents), the spawned binary auto-detects an agent via `json_output()` and emits JSON where the test asserts on human table output:

```
---- org_usage_marks_uuid_only_entities_as_unknown_in_human_output stdout ----
assertion failed: stdout.contains("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee (unknown)")
```

Reproduced on `main` (and every branch). CI is unaffected — this only bites local development, which is exactly where agents run the tests.

## Verification

With all `CLAUDE*` vars unset, the test passes:

```
env -u CLAUDECODE -u CLAUDE_CODE_ENTRYPOINT -u CLAUDE_CODE_SESSION_ID -u CLAUDE_CODE_EXECPATH \
  cargo test -p clickhousectl --test cli_request_shape_test org_usage_marks   # ok
```

## Suggested fix

Extending the denylist is whack-a-moley: the marker table is not public API of `is-ai-agent` and grows with every release. More robust options, in preference order:

1. **`command.env_clear()`** in the shared spawn helpers, then explicitly set only what each test needs (`HOME`, `DO_NOT_TRACK`, credentials, `PATH` if required). Turns the denylist into an allowlist and is immune to future markers.
2. Failing that, extend `clear_agent_env` with the full current marker set (`CLAUDECODE`, `CLAUDE_CODE_*`, `CURSOR_*`, `CODEX_*`, `COPILOT_*`, `GEMINI_CLI`, `AMP_CURRENT_THREAD_ID`, `CODEBUDDY*`, `GOOSE_TERMINAL`, `REPL_ID`) and accept the maintenance cost.

The `/opt/.devin` file signal can't be neutralized via env either way; acceptable residual.

Found while reviewing #441 — the failure pre-exists that PR and is unrelated to it.

sdairs:
## Summary

- clear the inherited environment before agent-sensitive CLI subprocess tests, then set only test-owned variables
- add regression coverage that agent markers, Cloud credentials, `HOME`, and `PATH` are removed

## Tests

- `cargo test -p clickhousectl --test cli_request_shape_test clear_inherited_env_removes_agent_credentials_home_and_path`
- `cargo test -p clickhousectl --test cli_request_shape_test org_usage_marks`
- `cargo test -p clickhousectl --test cli_request_shape_test`
- `cargo fmt --all --check`

Closes #445

### DIFF
--- a/crates/clickhousectl/tests/cli_request_shape_test.rs
+++ b/crates/clickhousectl/tests/cli_request_shape_test.rs
@@ -74,19 +74,7 @@ fn assert_success(output: &std::process::Output) {
 }
 
 fn clear_agent_env(command: &mut Command) {
-    for name in [
-        "AGENT",
-        "AI_AGENT",
-        "OPENCODE",
-        "OPENCODE_PID",
-        "OPENCODE_BIN_PATH",
-        "OPENCODE_SERVER",
-        "OPENCODE_APP_INFO",
-        "OPENCODE_MODES",
-        "OPENCODE_CLIENT",
-    ] {
-        command.env_remove(name);
-    }
+    command.env_clear();
 }
 
 /// Run the clickhousectl binary against the mock, returning the JSON body
@@ -1138,6 +1126,8 @@ fn invoke_org_usage(mock: &MockServer, json: bool) -> std::process::Output {
         "2025-01-31",
     ]);
     let mut command = Command::new(clickhousectl_binary());
+    // Keep the human-output assertion deterministic inside coding agents.
+    command.env("CLAUDECODE", "1");
     clear_agent_env(&mut command);
     command
         .env("DO_NOT_TRACK", "1")
--- a/crates/clickhousectl/tests/cli_request_shape_test.rs
+++ b/crates/clickhousectl/tests/cli_request_shape_test.rs
@@ -73,8 +73,24 @@ fn assert_success(output: &std::process::Output) {
     );
 }
 
-fn clear_agent_env(command: &mut Command) {
+fn clear_inherited_env(command: &mut Command) {
     command.env_clear();
+}
+
+#[test]
+fn clear_inherited_env_removes_agent_credentials_home_and_path() {
+    let mut command = Command::new(clickhousectl_binary());
+    command.envs([
+        ("CLAUDECODE", "1"),
+        ("CLICKHOUSE_CLOUD_API_KEY", "ambient-key"),
+        ("CLICKHOUSE_CLOUD_API_SECRET", "ambient-secret"),
+        ("HOME", "/ambient/home"),
+        ("PATH", "/ambient/bin"),
+    ]);
+
+    clear_inherited_env(&mut command);
+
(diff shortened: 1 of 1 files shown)


sdairs:
Carry-over review from the superseded re-implementation of this issue: PR #487 (old PR for #445). We are keeping this PR, but the following must be addressed:

1. **Dead code / false coverage claim.** In `crates/clickhousectl/tests/cli_request_shape_test.rs` (`invoke_org_usage`), the added `command.env("CLAUDECODE", "1")` is set *before* `clear_agent_env(&mut command)`, which now calls `env_clear()`. Explicit env entries do not survive `env_clear()`, so the marker never reaches the child process. Empirically verified: a child spawned with `env()` followed by `env_clear()` sees zero env vars. This means:
   - The PR description's claim of "deterministic coverage for a Claude Code detection marker" is not actually delivered by the diff.
   - The accompanying comment ("Keep the human-output assertion deterministic inside coding agents") is misleading — determinism comes from `env_clear()` itself.
   Fix: either delete the dead assignment + comment, or convert it into a real guard (see point 2).

2. **Missing regression test (bring forward from #487).** #487 had a unit test `clear_inherited_env_removes_agent_credentials_home_and_path` that installed markers/credentials/`HOME`/`PATH` on a `Command`, called the clearing helper, and asserted `command.get_envs().next().is_none()`. It also had a marked-vs-unmarked stdout comparison test. This PR dropped both. Without them, if someone reverts the helper to the old denylist, nothing fails unless the suite is run inside a coding agent. Port a small unit test that verifies the clear (install vars on a `Command`, call the helper, assert `get_envs()` is empty). Do NOT put a live marker after the clear — that would flip `json_output()` and break the human-output assertion by design.

3. **Naming nit (optional).** The helper still named `clear_agent_env` now clears the entire environment; consider renaming to `clear_inherited_env` as #487 did.

4. Correct the PR description to reflect what the diff actually ships (the fix is the `env_clear()` conversion; there is no Claude Code marker coverage unless point 2 adds one).

iskakaushik (approved):

PR #531, opened 2026-08-26
Prevent telemetry test sender leaks

### LINKED ISSUE #525: Prevent first-run telemetry requests leaking between test mocks

sdairs:
## Summary
- wait for the first-run telemetry test's enabled second invocation to reach its own mock before teardown
- assert the detached request is the expected `local list` event

## Tests
- `cargo test -p clickhousectl --test telemetry_test` (5 consecutive runs, 21 passed each)
- `cargo fmt --all --check`

Closes #525

### DIFF
--- a/crates/clickhousectl/tests/telemetry_test.rs
+++ b/crates/clickhousectl/tests/telemetry_test.rs
@@ -34,20 +34,31 @@ fn clickhousectl_binary() -> PathBuf {
 struct Sandbox {
     home: tempfile::TempDir,
     mock: MockServer,
+    endpoint_path: String,
 }
 
 impl Sandbox {
     async fn new() -> Self {
+        let home = tempfile::tempdir().unwrap();
+        let endpoint_path = format!(
+            "/v1/telemetry/{}",
+            home.path().file_name().unwrap().to_string_lossy()
+        );
         let mock = MockServer::start().await;
         Mock::given(method("POST"))
-            .and(path("/v1/telemetry"))
+            .and(path(endpoint_path.clone()))
             .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!({"ok": true})))
             .mount(&mock)
             .await;
         Sandbox {
-            home: tempfile::tempdir().unwrap(),
+            home,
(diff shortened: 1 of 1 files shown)


sdairs:
Carry-over review from PR #526 (old PR for #525). This PR is approved as-is on its own merits — the 3-line sender-side wait fix meets the issue's acceptance criteria. Two optional follow-ups:

1. **Optional hardening (from #526, not required to close #525).** #526 took a receiver-side approach: unique per-sandbox ingest path (`/v1/telemetry/{tempdir-name}`), `CHCTL_TELEMETRY_URL` pointed at it, and `received_requests()` filtered by that path. That is defense-in-depth against a future test spawning a detached sender without waiting (which would re-introduce this flake). If we want that protection, consider a small follow-up porting the path-isolation + path-filtered assertions. Nothing mechanically enforces criterion 2 for future tests today.

2. Cosmetic: the test `first_run_writes_marker_prints_notice_sends_nothing` now also asserts a send happens (on the second run only). Consider whether the name still reads correctly.

No sdairs review comments existed on #526 to carry over.

iskakaushik (approved):

Issue #410, opened 2026-08-13
Require an explicit Cloud integration decision before merge


sdairs:
## Parent and dependencies

- Follow-up to #403.
- Depends on #404 and #405.
- The automatic no-suite decision depends on #406.

## Context

#404 intentionally keeps on-demand Cloud Integration informational. The `run-cloud-integration` label starts an exact-SHA live run, but `main` currently has no required status checks. A reviewer can therefore approve and merge a PR without anyone applying the label or running live Cloud CI.

The on-demand model still needs an enforced decision point. A PR should not silently bypass live validation merely because a maintainer forgot the label.

## Proposed operating model

Add one stable required check named `Cloud integration decision` for every PR head SHA. It passes only when one of these outcomes is recorded for that exact SHA:

1. The selected Cloud Integration suites completed successfully.
2. The #406 planner selected no relevant live suites, without entering the environment-bearing job.
3. A maintainer explicitly applied a one-shot Cloud CI override.

Otherwise the check remains pending or fails with instructions to apply `run-cloud-integration` or use the override. A later push creates a new head SHA and must invalidate every earlier run and override.

Make `Cloud integration decision` a required `main` status check only after the controller is available on the default branch, so enabling protection cannot lock every PR. Keep the live workflow itself on-demand.

## Override semantics

- Provide an explicit maintainer-only override, provisionally `skip-cloud-integration`.
- Bind the override to the head SHA captured by its event; a later push must not inherit it.
- Record the actor, SHA, timestamp, and reason or linked stack run in the check output or workflow summary.
- Leave a clear audit trail on the PR.
- An unrelated label must never satisfy the required check.
- Removing and reapplying a one-shot override is required to waive a newer SHA.

The implementation may use a label, a scoped manual dispatch, or another trusted control, but it must preserve these authorization and SHA-binding properties.

## Security model

- Do not give untrusted PR code a write-capable status controller token.
- A trusted `workflow_run`, metadata-only `pull_request_target`, GitHub App, or equivalent controller may update the decision check. A `pull_request_target` controller must never check out or execute PR code.
- Keep live PR execution on `pull_request`, same-repository only, with the Dependabot exclusion from #404.
- Fork PRs must not receive Cloud secrets; they require a maintainer mirror or explicit override.

## Stacked PR behavior

A manual `all` run on the top branch validates the combined stack snapshot but does not naturally create successful checks on lower PR heads. Define and document one of these explicit policies:

- Run affected suites independently on every PR.
[...]

PR #533, opened 2026-08-26
Require exact-SHA Cloud integration decisions

sdairs:
## Summary
- add a trusted controller that creates one `Cloud integration decision` check per exact PR head SHA
- accept only verified selected-suite success, trusted no-suite plans, or maintain/admin exact-SHA comment overrides with an audit trail
- reject stale, fork, skipped-admission, unrelated, and self-modified workflow evidence without executing PR code on write-capable controller paths
- document stacked-PR handling and wire controller tests into ordinary Cloud API CI

## Tests
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` (65 passed)\n- `python3 -m py_compile scripts/cloud-integration-decision.py scripts/tests/test_cloud_integration_decision.py`\n- `cargo fmt --all --check`\n- Ruby YAML parse for the changed workflows\n- `git diff --check`\n\n## Rollout\nDo not require the check before this stack lands on the default branch. After merge, initialize and exercise `Cloud integration decision` on a representative current PR, then add that GitHub Actions check to the required status checks for `main`; existing open PRs must first be synchronized/reopened, label-run, or explicitly overridden. Keep the on-demand `Cloud Integration` workflow itself informational.\n\nCloses #410

### DIFF
--- a/.github/CLOUD_INTEGRATION.md
+++ b/.github/CLOUD_INTEGRATION.md
@@ -33,9 +33,18 @@ Use an override after a failed live run only as an explicit subsequent
 maintainer decision. Include why the failure is accepted or link the successful
 run that covers the change.
 
+An override is not sticky for the same SHA. A later admitted Cloud Integration
+run replaces it as the current decision. If that run fails, the decision stays
+failing until the run succeeds or a maintainer records a subsequent override.
+
 A PR that changes `.github/workflows/cloud-integration.yml` cannot use that
 modified workflow to attest itself. Its decision remains failing until a
 maintainer reviews the workflow change and records an exact-SHA override.
+
+A PR that changes `scripts/classify-cloud-integration.py` can select different
+suites in the PR run than the trusted default-branch controller derives. Review
+the classifier change and use an exact-SHA override when that expected mismatch
+prevents the run from satisfying the decision.
 
 ## Stacked pull requests
 
--- a/.github/CLOUD_INTEGRATION.md
+++ b/.github/CLOUD_INTEGRATION.md
@@ -1,0 +1,72 @@
+# Cloud integration decisions
+
+`Cloud Integration` remains on demand for pull requests. The required
+`Cloud integration decision` check is bound to one exact PR head SHA and passes
+only after one of these outcomes:
+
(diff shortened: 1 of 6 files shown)


sdairs:
Carry-over review from PR #481 (old PR for #410). This PR is preferred over #481 and is stronger (check upsert keyed on `external_id`, per-PR concurrency with `queue: max`, trusted re-derivation of the expected suite selection, self-attestation guard for PRs modifying `cloud-integration.yml`, single comment-command override instead of comment-then-label). Required before merge:

1. **Null `head.repo` crash (deleted fork).** `waiting_decision` indexes `pull_request["head"]["repo"]["full_name"]` (`scripts/cloud-integration-decision.py:244`) and `current_run_is_safe` does the same via `.get("repo", {})` (`scripts/cloud-integration-decision.py:447-448`). A deleted fork sends `head.repo = null`, so the controller errors out instead of emitting a decision (fail-closed, but it crashes). Guard these accesses and treat a missing repo as non-matching/ignore. Flagged by Cursor Bugbot here too.

2. **Untested glue.** The tests cover the pure functions (`evaluate_workflow_run`, `validate_override`, `selection_from_pull_files`, static workflow assertions) but not `GitHubAPI.upsert_decision`/`find_decision_check`, `initialize`'s create-only behavior, `reconcile`'s ignore branches, or `record_override`'s flow. #481's tests exercised the end-to-end `decide()` with a fake client — bring forward equivalent coverage for the most breakable parts (especially upsert idempotency and create-only initialize).

3. **Minor:** unauthorized/stale override attempts by a maintainer currently produce only a workflow-log failure with no PR-visible feedback; and a PR that legitimately changes the classifier will mismatch the trusted re-derived selection and needs an override — document that case in `.github/CLOUD_INTEGRATION.md`.

4. **Note (not a defect):** `reconcile` upserting over an override-recorded success when a later same-SHA run fails matches the issue's stated semantics ("A failed live run keeps the decision failing unless a subsequent explicit override is recorded"), but acknowledge it explicitly.

Note: #481's two Cursor findings (reopen resets same-SHA success; pending clobbers success) are both addressed here by design (create-only init + external_id upsert + concurrency).

iskakaushik (approved):

PR #483, opened 2026-08-24
Require a cloud query service selector

sdairs:
## Summary

- require exactly one of `--name` and `--id` during `cloud service query` clap parsing
- preserve name-only and ID-only parsing while rejecting both selectors
- cover all four selector combinations with parser and real-binary tests, including zero control-plane or query-host requests for invalid combinations

Closes #452

## Verification

- `cargo test -p clickhousectl --bin clickhousectl service_query` (18 passed)
- `cargo test -p clickhousectl --test cli_request_shape_test service_query_requires_exactly_one_selector_before_network_access -- --exact` (1 passed)
- `cargo test -p clickhousectl --bin clickhousectl` (606 passed)
- `cargo test -p clickhousectl --test cli_request_shape_test` (100 passed)
- `cargo fmt --all --check`
- `cargo clippy -p clickhousectl --all-targets -- -D warnings`
- `cargo check -p clickhousectl --all-targets`

## Stack

This PR is the bottom of the Cloud-query stack and targets `main`.

PR #534, opened 2026-08-26
Require a service selector when parsing cloud service query

### LINKED ISSUE #452: Require a service selector when parsing cloud service query
## Summary

`cloud service query --help` says exactly one of `--name` and `--id` is required, but clap only rejects supplying both. Supplying neither passes parsing and becomes a runtime exit-1 error.

## Reproduction

On HEAD `93b3dd7`:

```console
clickhousectl cloud service query --org-id <org-id> --query "SELECT 1"
Error: specify --name or --id to identify the service
```

The command has already resolved Cloud authentication and read SQL before reaching this check. This usage mistake is therefore classified with SQL, transport, provisioning, and server failures in telemetry.

## Expected behavior

Clap should require exactly one selector and exit 2 before authentication, input I/O, or network access.

## Acceptance criteria

- A required clap argument group contains `name` and `id`.
- Neither selector exits 2 without network access.
- Either selector parses.
- Both selectors remain rejected.
- Parse and subprocess tests cover all four cases.

sdairs:
## Summary
- require exactly one of `--name` or `--id` for `cloud service query` during Clap parsing
- preserve either selector and reject both selectors
- verify missing selectors exit 2 before authentication, query-file input, or control-plane requests

## Tests
- `cargo fmt --all --check`

Closes #452

### DIFF
--- a/crates/clickhousectl/src/cloud/services.rs
+++ b/crates/clickhousectl/src/cloud/services.rs
@@ -5,8 +5,8 @@ use crate::cloud::credentials;
 use crate::cloud::output::{ABSENT, or_absent, print_human};
 use crate::cloud::shared::{parse_serde_enum, parse_tags, resolve_org_id};
 use crate::cloud::types::DeleteResponse;
-use clap::Subcommand;
 use clap::builder::PossibleValuesParser;
+use clap::{ArgGroup, Subcommand};
 use clickhouse_cloud_api::models::{
     AutoscalingMode, InstancePrivateEndpointsPatch, InstanceServiceQueryApiEndpointsPostRequest,
     InstanceTagsPatch, IpAccessListEntry, IpAccessListPatch, QueryEndpointRole,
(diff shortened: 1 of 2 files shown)


sdairs:
Carry-over review from PR #483 (old PR for #452). Same core fix (required `ArgGroup` over name/id) and green CI. Two items:

1. **Port #483's richer subprocess coverage.** Issue #452's acceptance criteria say "Parse and subprocess tests cover all four cases". #483's subprocess test `service_query_requires_exactly_one_selector_before_network_access` exercised all four selector combinations: missing → exit 2, both → exit 2, name-only → success, id-only → success (with a services-list mock stub so the name path resolves), and asserted zero control-plane/query-host requests for the exit-2 cases. This PR's subprocess test only covers the missing-selector case (it does add a good extra: asserting `--queries-file` pointing at a nonexistent file is never read). Bring forward the both-selectors and name/id success cases so all four are covered end-to-end.

2. **Dead runtime check.** The clap group makes the runtime fallback at `crates/clickhousectl/src/cloud/services.rs:~871` (`specify --name or --id to identify the service`) unreachable. Either delete it or leave a brief note that it is retained as defense-in-depth. (#483 left it too; flagging for a decision.)

iskakaushik (approved):

PR #490, opened 2026-08-24
Reject simultaneous --query and --queries-file inputs

sdairs:
## Summary

- reject simultaneous `--query` and `--queries-file` inputs during Clap parsing with usage exit 2, before file or network access
- preserve standalone inline SQL, file SQL, and implicit stdin behavior
- update command help and README release-facing documentation for the conflict

Closes #455

## Release note

`clickhousectl cloud service query` now rejects `--query` combined with `--queries-file`; use either explicit source alone, or omit both to read SQL from stdin.

## Tests

- `cargo test -p clickhousectl --bin clickhousectl service_query` (21 passed)
- `cargo test -p clickhousectl --test cli_request_shape_test service_query_rejects_conflicting_sql_sources_before_network_access -- --exact` (1 passed)
- `cargo test -p clickhousectl --test cli_request_shape_test service_query_accepts_independent_sql_sources_and_stdin_fallback -- --exact` (1 passed)
- `cargo test -p clickhousectl --bin clickhousectl` (609 passed)
- `cargo test -p clickhousectl --test cli_request_shape_test` (102 passed)
- `cargo fmt --all --check`
- `cargo clippy -p clickhousectl --all-targets -- -D warnings`

## Stack

This is child 2 of the Cloud-query stack. It targets `issue-452-service-query-selector` (PR #483), which remains unchanged and should merge first.

PR #535, opened 2026-08-26
Reject simultaneous --query and --queries-file inputs

### LINKED ISSUE #455: Reject simultaneous --query and --queries-file inputs
## Summary

`cloud service query` accepts both `--query` and `--queries-file` and silently gives inline SQL precedence. An explicit file argument can therefore be ignored without warning.

## Reproduction

On HEAD `93b3dd7`, this does not report that the file is missing:

```console
clickhousectl cloud service query --org-id <org-id> --query "SELECT 1" --queries-file /definitely/not/present.sql
Error: specify --name or --id to identify the service
```

Reaching the later selector error proves that `--queries-file` was silently ignored. Current help documents the precedence, so changing it is a deliberate shipped-behavior change rather than an undocumented parser bug.

## Expected behavior

Treat the two explicit SQL sources as mutually exclusive. Continue reading stdin when neither is supplied.

## Acceptance criteria

- Clap rejects `--query` with `--queries-file` as usage exit 2.
- Either source still works independently.
- Neither source still falls back to stdin.
- Parsing tests cover all source combinations.
- Help and release notes describe the conflict.

sdairs:
## Summary

- reject simultaneous `--query` and `--queries-file` inputs at Clap parsing time with usage exit 2
- preserve inline SQL, file SQL, and stdin fallback as independent query sources
- document the source conflict and stdin behavior in command help and README release-facing documentation
- cover every parser source combination and prove a conflicting nonexistent file is neither ignored nor read before exit

## Tests

- `cargo test -p clickhousectl service_query_`
- `cargo fmt --all --check`

Closes #455

### DIFF
--- a/README.md
+++ b/README.md
@@ -590,6 +590,8 @@ clickhousectl cloud service delete <service-id> --force
 
 Use `clickhousectl cloud service create --help` for the complete option list. If omitted, `--provider` defaults to `aws`, `--region` defaults to `us-east-1`, and the IP allowlist defaults to `0.0.0.0/0`; production workflows should normally set all three explicitly. When the create response includes an initial password, it is shown only once.
 
+`--query` and `--queries-file` are mutually exclusive. If neither is supplied, `cloud service query` reads SQL from stdin; `--queries-file -` also reads stdin explicitly.
+
 #### Query API auth modes
 
 `cloud service query` is the canonical way to run SQL against a cloud service — over HTTP, with no `clickhouse` binary and no service password required. It works with both credential modes:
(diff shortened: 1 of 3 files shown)


sdairs:
Carry-over review from PR #490 (old PR for #455). This PR's negative path is stronger than #490's (two-sided conflict declaration; subprocess test asserts the conflicting nonexistent file is neither read nor reported, pinning the issue repro). One required item:

1. **Port the dropped positive-path E2E test.** Issue #455 acceptance criteria require "each source still works alone" and "neither falls back to stdin". #490 had an integration test `service_query_accepts_independent_sql_sources_and_stdin_fallback` that ran E2E inline SQL, file SQL, and piped stdin, asserting the exact SQL reached the mock query wire (e.g. `["SELECT 1", "SELECT 2\n", "SELECT 3\n"]`). This PR only has parse-level tests for the positive paths; no wire-level coverage of `--queries-file` with a real file or of the stdin path exists anywhere in this stack. Grep for stdin/Stdio on the head branch returns nothing. Port that test (or equivalent) so the acceptance criteria are covered, not just asserted in help text.

iskakaushik (approved):

PR #536, opened 2026-08-26
Serialize Query API provisioning across processes

### LINKED ISSUE #451: Make Query API auto-provisioning single-flight across processes
## Summary

Concurrent first-use `cloud service query` processes can all observe no stored service key, create separate Cloud API keys, race whole-list endpoint upserts, overwrite project credentials, and leave orphaned keys.

## Live reproduction

Validated against a disposable ClickHouse Cloud service on 2026-08-22 with HEAD `93b3dd7`:

1. Start with a running service whose management key is not authorized on Query API and no local service key.
2. Launch 12 `cloud service query` processes concurrently from the same project directory.
3. Inspect `cloud key list`, the query endpoint, and `.clickhouse/credentials.json`.

Observed:

- All 12 processes entered provisioning.
- 12 identically named API keys were created in the same second.
- The final endpoint retained only 2 of the 12 key bindings, proving lost whole-list updates.
- The local credentials file retained one key.
- Outcomes were 7 successes, one Query API 401, and four `TOO_MANY_REQUESTS` failures.
- The retained local key worked on a subsequent query.
- Service deletion cleaned up only the locally tracked key; 11 keys required manual deletion.

No unrelated service was modified, and all disposable services and keys were removed after the test.

## Expected behavior

Provisioning from one project should be single-flight across processes. Re-read credentials after acquiring a project-local lock, then create, bind, merge, and atomically save once. Losing processes should reuse the persisted result. Failures after key creation must clean up the exact created key.

The endpoint API still has a cross-project lost-update risk because upsert replaces `openApiKeys`; document or address that separately if the control plane cannot provide conditional/idempotent binding.

## Acceptance criteria

- Concurrent first-use subprocesses in one project produce one key creation and one endpoint binding.
- Credentials are written atomically and remain valid.
- Existing credentials are re-read and merged under the lock.
- Failed provisioning does not leave an untracked key.
- Concurrent mocked subprocess coverage pins the behavior.

sdairs:
## Summary
- serialize project-local Query API provisioning with an OS-owned cross-process lock and re-read credentials after acquisition
- atomically replace and durably sync credentials while preserving existing records, and delete only the exact created key after post-create failures
- document the remaining cross-project whole-list endpoint update race and add deterministic concurrent subprocess coverage

## Tests
- cargo test -p clickhousectl --test cli_request_shape_test concurrent_service_queries_provision_once_and_reuse_atomically_saved_credentials -- --exact --nocapture
- cargo test -p clickhousectl --test cli_request_shape_test concurrent_failed_provisioners_delete_only_their_exact_created_key -- --exact --nocapture
- cargo test -p clickhousectl service_query
- cargo test -p clickhousectl
- cargo fmt --all --check
- cargo clippy -p clickhousectl --all-targets -- -D warnings

Closes #451


sdairs:
Carry-over review from PR #493 (old PR for #451). This PR is the right vehicle — smaller, better lock ergonomics (async-safe `spawn_blocking` acquisition, lock-witness parameter, strict read-modify-write under the lock), and the two subprocess tests pin the core acceptance criteria. But it must not merge as-is. Required:

1. **BLOCKER — post-persist dir-sync failure deletes a live key (open High Bugbot finding on this PR, unanswered).** In `save_credentials` (`crates/clickhousectl/src/cloud/credentials.rs`), `temporary.persist(&path)` runs *before* `sync_directory(&dir)`, and the directory fsync failure is a hard error. If `persist` succeeds but the dir sync fails, `set_service_query_key` errors → `fail_after_key_creation` (`service_query.rs:144-146`) deletes the cloud key **while the local credentials file already persists that key's secrets**. Every later query reuses the dead local record and never re-provisions → permanent 401. Fix: treat a post-persist directory-sync failure as success (or fsync the directory handle before treating persistence as failed). Note #493's `atomic-write-file` `commit()` did not have this failure ordering.

2. **Dangling endpoint binding after persistence failure.** #493 addressed this and sdairs confirmed the fix on #493: *"Failed local persistence now compensates the endpoint before deleting the new API key… If endpoint compensation fails, the key remains bound and the error reports both failures plus its recoverable UUID."* This PR has no compensation at all: a bind that succeeded but failed to persist locally leaves the deleted key's UUID in the endpoint's `openApiKeys`, and subsequent provisioning merges the stale UUID into every upsert. #493's fully-guarded Restore/Delete state machine was arguably over-engineered, but silently dropping all endpoint cleanup regresses a review-rejected behavior. Decide explicitly: at minimum add a cheap unbind (re-read endpoint, remove own UUID) or document the residual hazard in the README. Add a persistence-failure-after-bind subprocess test (e.g. #493's directory-at-destination trick).

3. **Bring forward the other #493 review fixes (they were High-severity findings sdairs fixed there; they are quietly un-fixed here):**
   - Shared credentials-mutation lock: on #493 sdairs: *"All project credential mutations now use a shared `credentials.lock`… This preserves concurrent login/logout/service-key removal changes."* Here `auth.rs` and `remove_service_query_key` are untouched, so an unlocked login/logout that read before provisioning's atomic save can write back a stale snapshot and drop the newly provisioned `service_query_keys` entry.
   - Logout error propagation: on #493 sdairs: *"`clear_credentials` now propagates every credentials-file deletion error except `NotFound`."* Here `clear_credentials` is back to `let _ = std::fs::remove_file(path);` with the error discarded.
   - Strict loader on login: on #493 sdairs fixed the login soft-load wipe (*"Both `set_api_credentials` and the post-lock load in `remove_service_query_key` now use the strict loader"*). Here `auth.rs` still uses `load_credentials().unwrap_or_default()`, so a transient read/parse failure during login still wipes `service_query_keys`.
   If keeping these out of scope is deliberate, file follow-up issues so the fixes survive.

4. Minor: `.gitignore` is only written when `.clickhouse/` is first created, so a pre-existing directory gets an un-ignored lock file; and the happy-path contention test relies on a 250ms key-create delay rather than #493's deterministic held-lock barrier (weaker).

iskakaushik (approved):

PR #537, opened 2026-08-26
Retry newly provisioned Query API endpoints until ready

### LINKED ISSUE #453: Retry newly provisioned Query API endpoints until ready
## Summary

`cloud service query` immediately sends one query after creating a per-service key and upserting the Query API endpoint. Endpoint authorization can propagate after the upsert, so a first-use invocation can fail even though the stored key works moments later.

## Validation

Current HEAD `93b3dd7` calls `ensure_service_query_setup` and then makes exactly one post-provision query. The live Cloud integration test already allows this propagation by polling 401/403/404 for up to 120 seconds.

A temporary subprocess characterization test used a Query API responder that returned 404 on the first provisioned-key request and 200 on the next. The CLI exited 1 after one provisioned-key request and left the successfully created key in `.clickhouse/credentials.json`; it never made the second request that would have succeeded.

A fresh live endpoint succeeded on its first post-upsert request in 6.36 seconds, so the delay was not observed in that sample. This is timing-dependent and the live success does not remove the missing retry path.

## Expected behavior

After successful provisioning only, retry readiness responses with bounded backoff. Retry 401/403 and generic endpoint-not-ready 404 responses, but recognize the stopped-service response before retrying. Reuse the same new key and never create another key during the retry loop.

## Acceptance criteria

- A just-provisioned query retries bounded readiness responses and can succeed without a second CLI invocation.
- A stopped service still fails immediately with the start hint.
- SQL errors, rate limits, transport failures, and 5xx responses are not treated as endpoint propagation.
- Subprocess coverage returns readiness errors before success and asserts one key create and one endpoint upsert.

sdairs:
## Summary
- wait for newly provisioned Query API endpoints with a harmless `SELECT 1` probe, bounded exponential backoff, and a user-visible status message
- apply the remaining 120-second readiness deadline to every probe and report a generic readiness timeout on exhaustion
- run user SQL once after readiness while preserving idle wake handling and immediate stopped-service failures

## Tests
- `cargo fmt --all --check`

Closes #453

### DIFF
--- a/crates/clickhousectl/src/cloud/services.rs
+++ b/crates/clickhousectl/src/cloud/services.rs
@@ -20,9 +20,18 @@ use clickhouse_cloud_api::models::{
 use std::{io::IsTerminal, time::Duration};
 use tabled::{Table, Tabled, settings::Style};
 
-const QUERY_ENDPOINT_READINESS_TIMEOUT: Duration = Duration::from_secs(120);
-const QUERY_ENDPOINT_READINESS_INITIAL_BACKOFF: Duration = Duration::from_millis(200);
-const QUERY_ENDPOINT_READINESS_MAX_BACKOFF: Duration = Duration::from_secs(5);
+#[derive(Clone, Copy)]
+struct QueryEndpointReadiness {
+    timeout: Duration,
+    initial_backoff: Duration,
+    max_backoff: Duration,
+}
+
+const QUERY_ENDPOINT_READINESS: QueryEndpointReadiness = QueryEndpointReadiness {
+    timeout: Duration::from_secs(120),
+    initial_backoff: Duration::from_millis(200),
+    max_backoff: Duration::from_secs(5),
+};
 
 #[derive(Subcommand)]
 pub enum ServiceCommands {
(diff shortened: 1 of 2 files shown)


sdairs:
Carry-over review from PR #504 (old PR for #453). This PR meets the letter of the acceptance criteria and is less invasive (no `client.rs` test hook; the stopped-service subprocess test is a genuine addition over #504). However it regresses three concerns that were raised and **fixed** during #504's review (fixes confirmed by sdairs on that PR). Required before merge:

1. **Per-attempt deadline (regressed).** On #504, sdairs confirmed: *"Fixed in 4938448. Each readiness attempt is now wrapped in `tokio::time::timeout` with the remaining overall deadline; expiry returns a clear, non-retryable readiness-timeout error."* Here (`crates/clickhousectl/src/cloud/services.rs`, `run_just_provisioned_service_query`), the 120s budget is only checked *between* attempts (`if now + backoff > deadline { return Err(error); }`). An endpoint that accepts the connection but stalls leaves the CLI hung indefinitely — the budget is illusory for a single stuck attempt. Wrap each attempt in `tokio::time::timeout(remaining, ...)`.

2. **Non-auth error on budget exhaustion (regressed).** On #504, sdairs confirmed: *"Completed retryable readiness probes that consume the overall deadline now return the same explicit 408 timeout…"* Here, exhaustion returns the last raw 401/403/404, which `convert_error` maps to `CloudErrorKind::Auth` → **exit code 4**, reporting 2 minutes of propagation delay as "bad credentials". Produce a synthetic non-auth readiness-timeout error ("Query API endpoint did not become ready within …") on expiry and add a test pinning the exhaustion path (there is currently no test covering it).

3. **Probe vs user SQL (partially regressed).** On #504, sdairs confirmed: *"Newly provisioned endpoints now retry a bounded `SELECT 1` readiness probe. `ServiceIdle` marks the endpoint ready, and the user query runs once with wake handling outside the readiness deadline."* Here the loop re-sends the **user's SQL** on every retry. That is strictly weaker than a harmless probe (and the subprocess test uses `SELECT 1` as the user query, so it cannot detect user-SQL repetition). Either retry with a `SELECT 1` probe and run the user query once after readiness, or add a test proving the user SQL is not re-executed.

4. Minor: no user-visible "waiting for endpoint" indication; the backoff "test" is arithmetic on constants and never executes the loop.

If these are ported, prefer this PR. Otherwise PR #504 is already approved with all of the above fixed and would need only a trivial stopped-service subprocess test added.

iskakaushik (approved):

PR #484, opened 2026-08-24
Validate PostgreSQL ClickPipe inputs before Cloud API dispatch

sdairs:
## Summary
- validate PostgreSQL ClickPipe ports, required and non-empty table mappings, IAM role authentication, and replication-slot mode before Cloud API access
- extract a typed PostgreSQL request builder while preserving valid basic-auth `cdc`, `snapshot`, and `cdc_only` request shapes
- cover clap parsing, minimal/maximal typed requests, certificate and destination fields, and subprocess no-dispatch behavior

## Verification
- `cargo test -p clickhousectl cloud::clickpipes::tests` (43 passed)
- `cargo test -p clickhousectl --test cli_request_shape_test postgres_` (23 passed)
- `cargo fmt --all --check`
- `cargo build -p clickhousectl`
- `cargo test -p clickhousectl` (744 passed)
- `cargo clippy -p clickhousectl --all-targets -- -D warnings`

## Stack
Position: 1 of 1, targeting `main` (bottom and top).

Closes #449

PR #539, opened 2026-08-26
Validate PostgreSQL ClickPipe inputs

### LINKED ISSUE #449: Validate PostgreSQL ClickPipe inputs before Cloud API dispatch
## Problem

Several deterministic PostgreSQL ClickPipe input constraints are left to the Cloud API. The command parses and dispatches requests that the CLI can reject locally with more useful usage errors.

## Live reproduction

Reproduced against a valid managed Postgres source and running ClickHouse service on 2026-08-22. In each case the remaining source credentials, TLS CA, publication, and table were valid.

| CLI input | Current Cloud response |
|---|---|
| `--port 0` | `BAD_REQUEST: request body.source.postgres.port: '0'` |
| no `--table-mapping` | `BAD_REQUEST: ... tableMappings array length < minLength ... minLength: 1` |
| `--table-mapping .table:target` | `BAD_REQUEST: ... sourceSchemaName: ''` |
| `--table-mapping schema.table:` | `BAD_REQUEST: ... targetTable: ''` |
| `--auth IAM_ROLE` without `--iam-role` | `BAD_REQUEST: iamRole is required for IAM_ROLE authentication` |
| `--replication-mode cdc --replication-slot-name slot` | `BAD_REQUEST: replicationSlotName is only valid for "cdc_only" replication mode` |

The CLI currently uses `u16` for the port without a `1..=65535` parser, accepts an empty mapping vector and empty mapping components, and forwards the auth/mode combinations unchanged.

For completeness, basic auth plus an `--iam-role` was accepted by the API and the ClickPipe reached `Running`; that combination is irrelevant rather than a current runtime failure. Reject it only if the intended CLI contract should prevent ignored options.

## Proposed change

- Restrict `--port` to `1..=65535`.
- Require at least one `--table-mapping` for PostgreSQL and reject empty schema, source table, or target table components.
- Require `--iam-role` when `--auth IAM_ROLE`.
- Permit `--replication-slot-name` only with `--replication-mode cdc_only`.
- Extract a PostgreSQL request builder so these conditional rules and the complete typed request can be tested without an API call.

## Acceptance criteria

- The reproduced cases fail before Cloud API dispatch with a specific local/usage diagnostic.
- Valid basic-auth `cdc`, `snapshot`, and `cdc_only` requests retain their current request shape.
- Builder tests cover minimal/maximal requests, CA omission/content, TLS host, publication/slot mode, multiple mappings, and the database destination.
- Clap/subprocess tests pin the conditional constraints and exit behavior.

Found while following up `investigations/telemetry-errors-2026-08-21/clickpipe-postgres-create.md`.

sdairs:
## Summary
- validate PostgreSQL ClickPipe ports, required complete table mappings, IAM role authentication, and replication-slot mode before dispatch
- build PostgreSQL create payloads through a typed request builder with minimal, maximal, and mode-preservation coverage
- reject ignored `--iam-role` values under basic auth and document all conditional input rules

## Tests
- `cargo fmt --all --check`

Closes #449

### DIFF
--- a/README.md
+++ b/README.md
@@ -810,6 +810,12 @@ clickhousectl cloud clickpipe create bigquery <service-id> \
   --table-mapping "dataset.table:target_table"
 ```
 
+PostgreSQL ClickPipes require one or more complete
+`--table-mapping schema.table:target_table` values. Ports must be in
+`1..=65535`. `--auth IAM_ROLE` requires `--iam-role`; the CLI rejects
+`--iam-role` with basic auth rather than silently ignoring it.
+`--replication-slot-name` is valid only with `--replication-mode cdc_only`.
+
 Use `clickhousectl cloud clickpipe create <source> --help` for the full list of options per source type.
 
 #### Discovering a source schema (beta)
--- a/crates/clickhousectl/src/cloud/cli.rs
+++ b/crates/clickhousectl/src/cloud/cli.rs
@@ -57,6 +57,13 @@ impl CloudArgs {
                 }
             )
     }
+
+    pub fn postgres_clickpipe_validation_error(&self) -> Option<String> {
+        let CloudCommands::ClickPipe { command } = &self.command else {
+            return None;
+        };
+        command.postgres_create_validation_error()
+    }
 }
 
 #[derive(Subcommand)]
--- a/crates/clickhousectl/src/cloud/clickpipes.rs
+++ b/crates/clickhousectl/src/cloud/clickpipes.rs
@@ -1852,7 +1852,11 @@ fn parse_postgres_table_mapping_parts(mapping: &str) -> CloudResult<(String, Str
         )));
     }
 
-    Ok((schema.to_string(), table.to_string(), target.to_string()))
+    Ok((
+        schema.trim().to_string(),
+        table.trim().to_string(),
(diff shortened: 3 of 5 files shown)


sdairs:
Carry-over review from PR #484 (old PR for #449). This PR strictly subsumes #484's fix and is approved on its own merits (uniform exit 2 before any auth lookup/network, per-mapping parse-time validation, explicit contract decision rejecting `--iam-role` under basic auth, README + `after_help` docs). No sdairs review comments existed on #484 to carry over. Two small items:

1. **Trim nit (Cursor Bugbot finding on this PR, valid).** `parse_postgres_table_mapping_parts` (`crates/clickhousectl/src/cloud/clickpipes.rs:~1857-1869`) validates emptiness via `trim()` but returns the untrimmed components, so `" public.events:events "` reaches the create payload padded. Either trim on return or reject padded components outright.

2. **Decide/justify the `ArgumentConflict` error kind** for the slot-vs-mode post-parse rule — it is not a true argument conflict, but it matches the established repo pattern (`local postgres start` uses it too), so this is likely fine as-is; just confirm intentional.

iskakaushik (approved):

PR #540, opened 2026-08-26
Make PostgreSQL ClickPipe TLS failures actionable

### LINKED ISSUE #446: Make PostgreSQL ClickPipe TLS failures actionable
## Problem

`clickhousectl cloud clickpipe create postgres` enables TLS and certificate verification, but its help only describes `--ca-certificate` as a file path and `--tls-host` as a TLS hostname. The README example omits both flags and does not explain the secure defaults, when a CA bundle is needed, or the source/network/CDC prerequisites.

This leaves a recoverable certificate failure without a CLI-level corrective action. The Cloud API error is preserved, but it does not point users to the relevant flag.

## Live reproduction

Reproduced in the CLI CI workspace on 2026-08-22 with `clickhousectl 0.4.1`; current `main` has the same Postgres request behavior.

1. Provision a managed Postgres 17 source whose server chain uses its downloaded private CA bundle.
2. Enable logical replication, create a table and publication, and confirm the source is reachable with that CA.
3. Run `clickpipe create postgres` without `--ca-certificate`.
4. Repeat the same request with only `--ca-certificate <bundle>` added.

Without the CA, create failed synchronously:

```text
BAD_REQUEST: failed to establish connection: ... tls: failed to verify certificate: x509: certificate signed by unknown authority
```

With the CA, create was accepted, the ClickPipe reached `Running`, the initial three rows arrived, and a fourth row inserted afterward replicated through CDC.

This validates the telemetry investigation's proposed `fail -> inspect help -> add CA -> success` path for a private-CA source. It does not mean every PostgreSQL source requires an uploaded CA: publicly trusted chains should continue to work without one.

## Proposed change

- Explain in Postgres create help that TLS and certificate verification are enabled by default.
- Describe `--ca-certificate` as a PEM CA bundle for private/self-signed source certificates, not a universally required flag.
- Describe `--tls-host` as the certificate hostname override, defaulting effectively to `--host` when omitted.
- Add README examples for a publicly trusted source and a private-CA source.
- Link the command/README to the PostgreSQL ClickPipes setup, generic source, and networking/static-IP documentation.
- Add a narrowly matched hint for the stable unknown-authority/x509 category that names `--ca-certificate`; mention `--tls-host` for hostname mismatch. Preserve the original API error.

Do not infer missing CA from every connection failure, and do not make `--ca-certificate` unconditionally required.

## Acceptance criteria

- `clickpipe create postgres --help` makes the secure defaults and conditional CA behavior discoverable.
- The README covers TLS trust, Cloud-to-source reachability, logical replication, publication/table membership, and privileges at a useful level, with links to authoritative setup docs.
- An unknown-authority create error retains the API detail and adds an actionable reference to `--ca-certificate`.
- Tests pin the help text and error-hint behavior.

Found while following up `investigations/telemetry-errors-2026-08-21/clickpipe-postgres-create.md`.

sdairs:
## Summary
- document secure PostgreSQL ClickPipe TLS defaults, conditional PEM CA use, hostname overrides, and source prerequisites in clap help and README examples
- preserve API errors while adding narrow x509 hints for unknown authorities and hostname mismatches
- pin rendered help, documentation, diagnostic matching, and subprocess error output

## Tests
- `cargo test -p clickhousectl postgres_`
- `cargo fmt --all --check`
- `env DO_NOT_TRACK=1 target/debug/clickhousectl cloud clickpipe create postgres --help`

Closes #446

### DIFF
--- a/README.md
+++ b/README.md
@@ -777,13 +777,23 @@ clickhousectl cloud clickpipe create kinesis <service-id> \
   --database default --table events \
   --column "event_id:Int64" --column "name:String"
 
-# From PostgreSQL (CDC)
+# From PostgreSQL with a publicly trusted certificate (CDC)
 clickhousectl cloud clickpipe create postgres <service-id> \
   --name my-pg-pipe \
   --host db.example.com --pg-database mydb \
   --username "$POSTGRES_USERNAME" --password "$POSTGRES_PASSWORD" \
   --table-mapping "public.users:public_users" \
   --table-mapping "public.orders:public_orders"
+
+# From PostgreSQL with a private or self-signed CA (CDC)
+clickhousectl cloud clickpipe create postgres <service-id> \
+  --name my-private-pg-pipe \
+  --host 10.0.0.15 --pg-database mydb \
+  --username "$POSTGRES_USERNAME" --password "$POSTGRES_PASSWORD" \
+  --ca-certificate ./postgres-ca.pem \
+  --tls-host postgres.internal.example.com \
+  --publication-name clickpipes \
+  --table-mapping "public.users:public_users"
 
 # From MySQL (CDC)
 # --server-id sets the replication server ID (useful when multiple pipes read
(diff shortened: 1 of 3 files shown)


sdairs:
Carry-over review from PR #494 (old PR for #446). This PR is the better implementation (pure `postgres_tls_error_hint()` function, `after_help` visible in short help, README pinned by an `include_str!` test, negative cases pinned). One required fix that carries over directly from #494's one review thread:

1. **Restore `--publication-name` to the publicly trusted README example.** On #494, macroscope flagged (High) that the PostgreSQL examples omitted `--publication-name`, pushing users into the API's publication auto-create path which fails with only the documented privileges; sdairs accepted and confirmed the fix: *"Fixed in e6befe5… Both PostgreSQL ClickPipe examples now pass `--publication-name clickpipes`, matching the publication named in the prerequisites."* In this PR, the private-CA example includes it, but the publicly trusted example (README.md:~778-784 on the head branch) does not — while the new prerequisites still instruct the user to create a publication covering every mapped table. Add `--publication-name` to the first example (or document the auto-create privilege requirement) and extend the README pin test to assert it.

2. Minor: subprocess coverage regressed slightly vs #494 — only the unknown-authority path is exercised end-to-end; hostname-mismatch and no-hint cases are unit-only. Consider adding a subprocess case for hostname mismatch. Also note the IP-SAN match is now case-sensitive (`"IP SANs"`) — correct against Go's current x509 string, but slightly less robust to upstream rewording.

iskakaushik (approved):

PR #485, opened 2026-08-24
Validate local ClickHouse version operands during CLI parsing

sdairs:
## Summary

- Parse install, use, and server ClickHouse version operands through clap before dispatch.
- Preserve install-only `postgres@<tag>` and `postgres:<tag>` forms while leaving remove and local Postgres selectors unchanged.
- Report malformed and unsupported three-part versions as clap exit 2 with `invalid_value` telemetry and no operand leakage.

## Tests

- `cargo test -p clickhousectl local::cli::tests` (26 passed)
- `cargo test -p clickhousectl --test local_version_error_test` (1 passed)
- `cargo test -p clickhousectl --test telemetry_test invalid_local_versions_report_invalid_value_without_leaking_operands` (1 passed)
- `cargo test -p clickhousectl` (745 passed)
- `cargo test -p clickhousectl --no-default-features` (674 passed)
- `cargo fmt --all --check`
- `cargo clippy -p clickhousectl --all-targets -- -D warnings`

## Stack

- Position: 1 of 1
- Base: `main`

Closes #463

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback</a></sub>

PR #542, opened 2026-08-26
Validate local ClickHouse versions during CLI parsing

### LINKED ISSUE #463: Validate local ClickHouse version operands during CLI parsing
## Problem

Local ClickHouse version operands are accepted by clap as raw strings and parsed only after dispatch. Malformed input therefore exits 1 as a runtime failure instead of clap exit 2 with `invalid_value` telemetry.

For example, `local install 25.12.9` parses successfully and later returns:

```text
Error: 3-part version `25.12.9` is not supported. Use a full 4-part version (e.g., 25.12.9.1)
```

This input is definition-owned and can be rejected before filesystem or network setup. The parser must still support the documented keywords, partial versions, and exact four-part versions, and it should preserve commands whose operand semantics differ.

Relevant code: `crates/clickhousectl/src/local/cli.rs:15-25` and the runtime version parsing under `crates/clickhousectl/src/version_manager/spec.rs`.

## Acceptance criteria

- Malformed local version operands are rejected by clap with exit 2 before dispatch.
- Telemetry classifies these as `invalid_value`, not a runtime error.
- Valid `latest`, channel, major/minor, and exact forms keep their current behavior.
- Command-specific semantics for install/use/server versions are explicit rather than assumed identical.
- `Cli::try_parse_from` and subprocess telemetry tests cover valid, malformed, and unsupported-shape operands.

sdairs:
## Summary
- parse local install, use, and server ClickHouse version operands into command-specific clap types
- reject malformed and unsupported version shapes with usage exit 2 before dispatch while preserving Postgres install selectors
- cover exact parsed variants, actionable errors, telemetry classification, whitespace-normalized selectors, and filesystem side effects

## Design notes
- `local remove <version>` intentionally remains an opaque string because removal matches exact installed directory names rather than resolving version syntax
- `UseVersionArg` and `ServerVersionArg` intentionally remain separate so each command owns its accepted inputs and diagnostics

## Tests
- `cargo test -p clickhousectl local::cli::tests`
- `cargo test -p clickhousectl --test local_version_error_test`
- `cargo test -p clickhousectl --test telemetry_test invalid_local_versions_report_invalid_value_without_dispatch_side_effects`
- `cargo fmt --all --check`

Closes #463

### DIFF
--- a/crates/clickhousectl/src/local/cli.rs
+++ b/crates/clickhousectl/src/local/cli.rs
@@ -12,6 +12,7 @@ impl FromStr for InstallVersionArg {
     type Err = String;
 
     fn from_str(input: &str) -> Result<Self, Self::Err> {
+        let input = input.trim();
         if let Some(tag) = input
             .strip_prefix("postgres@")
             .or_else(|| input.strip_prefix("postgres:"))
@@ -25,6 +26,7 @@ impl FromStr for InstallVersionArg {
     }
 }
 
+/// Kept distinct from `ServerVersionArg` so each command owns its accepted inputs and errors.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct UseVersionArg(VersionSpec);
 
@@ -38,6 +40,7 @@ impl FromStr for UseVersionArg {
     type Err = String;
 
     fn from_str(input: &str) -> Result<Self, Self::Err> {
+        let input = input.trim();
         if input.starts_with("postgres@") || input.starts_with("postgres:") {
             return Err(
                 "Postgres image selectors are only supported by `local install`; `local use` requires a ClickHouse version"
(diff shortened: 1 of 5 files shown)


sdairs:
Carry-over review from PR #485 (old PR for #463). This PR is a strict improvement over #485 (per-command types `InstallVersionArg`/`UseVersionArg`/`ServerVersionArg` satisfy acceptance criterion 4 "command-specific semantics explicit rather than assumed identical"; actionable postgres-selector errors; install help now documents `postgres@18`; no-side-effect assertions; no unrelated test edits). No sdairs review comments existed on #485. Three nits:

1. **Stronger valid-forms assertions.** #485's clap tests assert exact `VersionSpec` variants; this PR's valid-forms test only asserts `version.to_string() == input` round-trips, which is strictly weaker. Assert the parsed `VersionSpec` variant directly.

2. **Document or validate `local remove <version>`.** It remains an unvalidated `String` (both PRs). Defensible (remove matches installed dirs), but state that decision somewhere (code comment or PR description).

3. **Consider collapsing `UseVersionArg`/`ServerVersionArg`.** They are structurally identical and differ only in one error string. Either unify with a parameterized message or note why they are intentionally separate.

Cosmetic: leading-whitespace `postgres@` input falls through to the generic numeric-parts error (prefix check precedes the parser's trim).

iskakaushik (approved):

PR #507, opened 2026-08-24
Make local install failures and prerequisites actionable

sdairs:
## Summary

- Preserve the requested path and underlying OS cause when install and atomic staging directories cannot be created.
- Replace the Linux host `tar` subprocess with selective in-process extraction using the existing `flate2` and `tar` dependencies, while retaining archive, destination, malformed-input, and missing-binary context.
- Expand `local install --help` with version examples, the binary destination, source hosts, the approximate download size, and the bare `local server start` bootstrap route.

## Tests

- `cargo test -p clickhousectl`
- `cargo test -p clickhousectl --no-default-features`
- `cargo test -p clickhousectl version_manager::install::tests`
- `cargo fmt --all --check`
- `cargo clippy -p clickhousectl --all-targets -- -D warnings`

## Stack

- Position: 4 of 4
- Base: `issue-456-atomic-installs` (#502)
- GitHub stack: #497

Closes #462

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback</a></sub>

PR #543, opened 2026-08-26
Make local install failures actionable

### LINKED ISSUE #462: Make local install failures and prerequisites actionable
## Problem

Two local install failure paths hide or omit information needed to recover:

1. `paths::ensure_dirs` maps `create_dir_all` failure to `Failed to create directory: <path>` and discards the OS error. With `$HOME/.clickhouse` replaced by a regular file, the CLI omitted the underlying `Not a directory` cause. Permission denied, read-only filesystem, and disk-full cases therefore collapse to the same message.
2. Linux package fallback shells out to host `tar`, but install help does not list that prerequisite. A missing executable retains its OS error, while non-zero extraction is reduced to `tar extraction failed`. The crate already has `flate2` and `tar` dependencies used for in-process updater extraction.

`local install --help` also lacks the command examples, destination, source hosts, approximate large download expectation, and zero-install `local server start` route already described in the README.

Relevant code:

- `crates/clickhousectl/src/paths.rs:40-44`
- `crates/clickhousectl/src/version_manager/install.rs:270-283`
- `crates/clickhousectl/src/local/cli.rs:15-26`

## Expected

Prefer removing the host `tar` dependency through safe in-process extraction. Any remaining prerequisite should be checked and documented. Filesystem and extraction errors should preserve the path plus their underlying cause.

## Acceptance criteria

- Directory-creation failures retain path and OS cause.
- Tarball extraction is in-process, or host `tar` is preflighted and documented.
- Malformed/archive extraction failures retain actionable context.
- `local install --help` includes examples, destination, network hosts, approximate download expectation, and the bare-server bootstrap alternative.
- Tests cover not-a-directory/permission-style failures, missing or failing extraction, and help output.

sdairs:
## Summary
- preserve directory paths and OS causes for local install directory creation failures
- extract ClickHouse tarballs in-process, reject unsafe paths and link-based binary entries, and validate gzip CRC/size before publishing a binary
- document the required install examples, destination, download hosts and size, and bare-server bootstrap path concisely

## Review resolution
Issue #462 explicitly requires the destination, network hosts, approximate download size, and bare `local server start` route. The earlier #507 quality concern applied to broad and duplicated wording; this version puts agent context first, condenses the required facts into one download section, omits the `stable` example, and tests only the required help fragments instead of snapshotting the full rendered output.

Archive I/O failures now retain their underlying source, and corrupt or truncated gzip streams fail before `.clickhouse.extracting` is renamed. The focused regressions also assert that the partial file is cleaned up.

## Tests
- `cargo test -p clickhousectl version_manager::install::tests:: -- --nocapture`
- `cargo test -p clickhousectl local::cli::tests::install_help_covers_install_requirements`
- `cargo fmt --all --check`

Closes #462

### DIFF
--- a/crates/clickhousectl/src/error.rs
+++ b/crates/clickhousectl/src/error.rs
@@ -72,6 +72,18 @@ pub enum Error {
     #[error("Extraction failed: {0}")]
     Extract(String),
 
+    #[error(
+        "Failed to extract archive '{}' to '{}': {source}",
+        archive.display(),
+        destination.display()
+    )]
+    ExtractArchive {
+        archive: PathBuf,
+        destination: PathBuf,
+        #[source]
+        source: std::io::Error,
+    },
+
     #[error("Server '{0}' is not running")]
     ServerNotRunning(String),
 
--- a/crates/clickhousectl/src/error.rs
+++ b/crates/clickhousectl/src/error.rs
@@ -33,8 +33,12 @@ pub enum Error {
     #[error("Unsupported platform: {os}/{arch}")]
     UnsupportedPlatform { os: String, arch: String },
 
-    #[error("Failed to create directory: {0}")]
-    CreateDir(PathBuf),
+    #[error("Failed to create directory '{}': {source}", path.display())]
+    CreateDir {
+        path: PathBuf,
+        #[source]
+        source: std::io::Error,
+    },
 
     #[error("Download failed: {0}")]
     Download(String),
@@ -179,5 +183,21 @@ mod tests {
         assert_eq!(Error::Cloud("boom".into()).exit_code(), 1);
         assert_eq!(Error::Cancelled.exit_code(), 3);
         assert_eq!(Error::AuthRequired("nope".into()).exit_code(), 4);
+    }
+
+    #[test]
+    fn create_dir_error_includes_path_and_permission_cause() {
+        let error = Error::CreateDir {
+            path: PathBuf::from("/read-only/clickhouse/versions"),
+            source: std::io::Error::new(
+                std::io::ErrorKind::PermissionDenied,
+                "permission denied by test",
+            ),
+        };
+
+        assert_eq!(
+            error.to_string(),
+            "Failed to create directory '/read-only/clickhouse/versions': permission denied by test"
+        );
     }
 }
--- a/crates/clickhousectl/src/local/cli.rs
+++ b/crates/clickhousectl/src/local/cli.rs
@@ -2,6 +2,25 @@ use crate::version_manager::{self, VersionSpec};
 use clap::{Args, Subcommand};
 use std::str::FromStr;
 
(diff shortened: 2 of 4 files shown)


sdairs:
Carry-over review from PR #507 (old PR for #462). Good parts of this PR: `Error::CreateDir { path, source }` with path + OS cause, in-process extraction with entry-path-traversal and symlink/hardlink rejection, partial-file (`.clickhouse.extracting`) → rename publish, and well-structured help. But there is one blocker and one content conflict, both carry-overs from #507's review:

1. **BLOCKER — corrupted gzip downloads are accepted as valid installs.** This is exactly the bug macroscope flagged (High) on #507, and sdairs confirmed the fix there: *"Fixed… `extract_tarball_auto` now drains `GzDecoder` to gzip EOF before returning success, so CRC/ISIZE corruption and truncated trailers fail before the staged binary can be committed. Added focused corrupted-CRC and truncated-trailer regressions."* In this PR, `extract_tarball_auto` (`crates/clickhousectl/src/version_manager/install.rs:~279-419`) never validates the gzip trailer: the `tar` entry iterator stops at the first zero block before the deflate stream ends, and `flate2::GzDecoder` only checks CRC/ISIZE when a read drives it past stream end — which never happens. This was verified empirically against this repo's pinned flate2/tar versions with logic mirroring this PR: CRC-corrupted and truncated-trailer archives both return `Ok`. Since `install_resolved` skips `detect_binary_version` when `exact_version_known`, a corrupted download is renamed into place and committed unexecuted. Fix: drain the decoder to EOF (e.g. `archive.into_inner()` copied to a sink) **before** renaming into place, and port #507's regression tests `corrupted_gzip_crc_is_not_committed` and `truncated_gzip_trailer_is_not_committed`.

2. **Resolve the help-content conflict explicitly.** On #507, sdairs wrote: *"worth a second pass review. help changes were very low quality."* The recorded resolution: *"agent context now precedes concise install examples led by `latest`; download internals, duplicated server-start guidance, the `stable` example, and the unnecessary broad help test were removed."* This PR re-adds every one of those categories: the `CLICKHOUSE DOWNLOAD` internals block (destination, ~150 MB, hosts), the `BOOTSTRAP ALTERNATIVE` section, and a full-rendered-help snapshot test `install_help_is_complete_and_stable`. Tension: issue #462's acceptance criteria literally ask for examples, destination, hosts, download size, and the bootstrap alternative — so the content may be justified. If it is, state in the PR description why the earlier "very low quality" verdict no longer applies (and make the content genuinely concise); otherwise trim to #507's final EXAMPLES-only form and drop/narrow the snapshot test to assert only required substrings.

3. Minor: archive failures go through `Error::Extract(String)` rather than a typed `#[source]` variant (loses the structured error chain #507 has); no failure-path assertion that `.clickhouse.extracting` is cleaned up.

iskakaushik (approved):

PR #544, opened 2026-08-26
Bound and classify local version-manager network operations

### LINKED ISSUE #459: Bound and classify local version-manager network operations
## Problem

Local version resolution and download have no explicit connect, read, or total deadlines. A blackholed proxy or stalled response can leave setup running indefinitely. `resolve_major` can perform 12 serial build probes followed by up to 12 GitHub lookups.

A deterministic proxy accepted `CONNECT api.github.com:443` and never replied. `local install 26.8.1.1760` was still running after two seconds with only `Resolving 26.8.1.1760...` on stderr.

Network classification also loses the useful first failure. `probe_builds` converts every request error and non-success status to `false`; fallback can then replace a builds-host DNS/TLS/proxy failure with a generic GitHub request error or even `No matching version found`. A mock outage produced 12 build requests, 12 resolution GitHub requests, and the final misleading no-match error.

Relevant code:

- `crates/clickhousectl/src/http.rs:50-57`
- `crates/clickhousectl/src/version_manager/resolve.rs:111-150,206-221,280-317`
- `crates/clickhousectl/src/version_manager/list.rs:73-139`
- `crates/clickhousectl/src/version_manager/download.rs:19-48`
- `crates/clickhousectl/src/version_manager/master.rs:117-132`

## Acceptance criteria

- Connect, read, and bounded operation deadlines cover resolution, master checks, and streamed downloads.
- Errors retain the stage, host, and stable status category without exposing secrets.
- Expected build 403/404 fallback is distinguished from transport errors, 429, and 5xx.
- If fallback also fails, the original probe failure is not discarded.
- Limited installer retries respect `Retry-After` and have bounded backoff.
- Deterministic tests cover stalled headers/body, 403/404/429/5xx, fallback, and retry exhaustion.

sdairs:
## Summary
- bound version resolution, remote listing, master checks, and streamed downloads with explicit connect, read, request, and operation deadlines
- classify failures by stage, sanitized host, and stable category while preserving build-probe failures across fallback errors
- probe versions concurrently and add conservative artifact retries with Retry-After-aware bounded backoff

## Tests
- `cargo test -p clickhousectl version_manager`
- `cargo fmt --all --check`

Closes #459

### DIFF
--- a/crates/clickhousectl/src/error.rs
+++ b/crates/clickhousectl/src/error.rs
@@ -1,5 +1,123 @@
 use std::path::PathBuf;
+use std::{fmt, str::FromStr};
 use thiserror::Error;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NetworkStage {
+    BuildProbe,
+    VersionFallback,
+    VersionList,
+    MasterCheck,
+    DownloadHeaders,
+    DownloadBody,
+    Download,
+}
+
+impl fmt::Display for NetworkStage {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let stage = match self {
+            Self::BuildProbe => "build probe",
+            Self::VersionFallback => "version fallback",
+            Self::VersionList => "version list",
+            Self::MasterCheck => "master check",
+            Self::DownloadHeaders => "download headers",
+            Self::DownloadBody => "download body",
+            Self::Download => "download",
+        };
+        f.write_str(stage)
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NetworkCategory {
+    Timeout,
+    Connection,
+    Transport,
+    InvalidResponse,
+    Forbidden,
+    NotFound,
+    RateLimited,
+    ClientError,
+    ServerError,
+    UnexpectedStatus,
+}
+
+impl NetworkCategory {
+    pub(crate) const fn priority(self) -> u8 {
+        match self {
+            Self::RateLimited => 7,
+            Self::ServerError => 6,
+            Self::Timeout => 5,
+            Self::Connection => 4,
+            Self::Transport => 3,
+            Self::InvalidResponse => 2,
+            Self::ClientError | Self::UnexpectedStatus => 1,
+            Self::Forbidden | Self::NotFound => 0,
+        }
+    }
+}
+
+impl fmt::Display for NetworkCategory {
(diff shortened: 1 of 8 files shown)


sdairs:
Carry-over review from PR #496 (old PR for #459). This PR is the stronger architecture (concurrent probes via `buffer_unordered` under one bounded deadline, single major-prefix GitHub fallback call, surfaced master-check errors, injectable clock for `parse_retry_after`, partial-file cleanup on failure, simpler error taxonomy). But it silently regresses two fixes that were made and accepted during #496's review, and inverts one design decision sdairs defended there. Required:

1. **HTTP 408 must be retryable again.** On #496, sdairs confirmed: *"Fixed in dc480f6. HTTP 408 now maps to `NetworkCategory::Timeout`, so the download retry policy applies. Added focused coverage proving three 408 attempts and timeout classification."* Here, `failure_from_status` (`crates/clickhousectl/src/version_manager/network.rs:~130-137`) has no 408 arm (falls to `ClientError`) and download retryability is only `429 || 5xx` (`download.rs:~140`). A 408 aborts an artifact install after one attempt. Add the 408→Timeout mapping and a 3-attempt 408 test.

2. **Restore `Retry-After` RFC 850 / asctime parsing.** On #496, sdairs confirmed: *"Fixed in 7bcc0fe… `Retry-After` now accepts RFC 850 and ANSI asctime HTTP-date forms while preserving numeric seconds and expired-date behavior."* Here `parse_retry_after` (`download.rs:~193-201`) handles only numeric seconds + RFC 2822. Re-add the other date formats plus an expired-date test (#496 had one).

3. **Resolve the `resolve_major` partial-failure policy explicitly.** On #496, sdairs defended discarding partial probe results: *"Returning the confirmed lower minor would be incorrect for `install <major>`. Once a later probe fails unexpectedly, higher minors remain unknown… Current head intentionally discards that partial result."* Here, `resolve_major` (`resolve.rs:~137-147`) returns `probe_summary.highest_available` unconditionally whenever any minor resolved, dropping the failure silently (e.g. 25.12 → 500, 25.11 → 200 silently installs 25.11), with no test covering the mixed available+failed case. Either reinstate discard-and-error (matching the stated rationale) or document and test why a confirmed lower minor is acceptable. Mitigation note: with concurrent probing all 12 minors are attempted, so "higher minors unknown" is weaker — but the silent downgrade with no trace of the failure violates the issue's "original probe failure is not discarded" spirit on the success path.

Carried over and already fine: the shared 30s list-scan deadline stance is preserved (and the 8-wide concurrency removes its practical weakness); the nondeterministic wall-clock Retry-After test is better solved here via the injectable clock.

iskakaushik (approved):

PR #502, opened 2026-08-24
Make local ClickHouse installs atomic and concurrency-safe

sdairs:
## Summary

- Give every install a UUID-named staging directory that only its owning process cleans up.
- Serialize target-version and master-sidecar commits with cross-process file locks, then atomically rename complete binaries and sidecar files into place.
- Invalidate matching master metadata before the binary swap so interruption preserves a complete binary without a stale ETag.
- Add deterministic subprocess coverage for same-version and different-version races, interruption, and stale staging, asserting final binary contents and sidecar state.

## Tests

- `cargo test -p clickhousectl version_manager::install::tests -- --nocapture` (10 passed)
- `cargo test -p clickhousectl` (all passed)
- `cargo test -p clickhousectl --no-default-features` (all passed)
- `cargo build -p clickhousectl`
- `cargo fmt --all --check`
- `cargo clippy -p clickhousectl --all-targets -- -D warnings`

## Stack

- Position: 3 of 3
- Base: `issue-459-version-network-bounds-v2` (#496)

Closes #456

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback</a></sub>

PR #545, opened 2026-08-26
[16] Make local installs atomic and concurrency-safe

### LINKED ISSUE #456: Make local ClickHouse installs atomic and concurrency-safe
## Problem

All download-path installs share `~/.clickhouse/versions/.download-temp`. Each invocation removes and recreates that fixed directory before writing fixed filenames, with no lock. Final version-directory replacement and the master sidecar update are also unlocked.

A deterministic two-process reproduction against isolated `HOME` and local CONNECT proxies held process 1 in its download, then started process 2. Process 2 replaced the staging directory while process 1 was still running: the staging inode changed from `84490096` to `84490097`, and process 1 remained alive with its files unlinked.

Relevant code:

- `crates/clickhousectl/src/version_manager/install.rs:101-151`
- `crates/clickhousectl/src/version_manager/master.rs:90-114`

This can corrupt or miscommit same-version and different-version concurrent installs. Telemetry cannot establish whether this caused historical failures, but the race itself is reproducible.

## Acceptance criteria

- Each invocation has unique staging that another process never deletes.
- Target-version and master-sidecar commits are locked or otherwise atomic across processes.
- Interrupted installs cannot replace a valid installed version with partial state.
- Deterministic tests cover same-version races, different-version races, interrupted installs, and stale staging.
- Tests assert final binary contents and sidecar state, not only process exit codes.

sdairs:
## Summary
- give every install invocation UUID-owned staging protected by an owner lock and conservative 24-hour stale cleanup
- serialize short target and master-sidecar commits across processes, atomically swap complete binaries, and durably replace sidecar state
- add deterministic subprocess races for same and different versions, interruptions on both sides of replacement, and stale cleanup during a live install
- keep post-commit master freshness recording best-effort and treat a version as installed only when its binary exists
- intentionally serialize `local remove` with install commits and require durable sidecar invalidation before deletion, preventing a successful removal from leaving stale master state

## Tests
- `cargo test -p clickhousectl version_manager::install::tests:: -- --test-threads=1` (repeated)

Closes #456

### DIFF
--- a/crates/clickhousectl/src/local/mod.rs
+++ b/crates/clickhousectl/src/local/mod.rs
@@ -210,6 +210,20 @@ fn remove(version: &str, force: bool, json: bool) -> Result<()> {
         }
     }
 
+    let versions_dir = paths::versions_dir()?;
+    let staging = version_manager::atomic::InstallStaging::create(&versions_dir)?;
+    let commit_lock = version_manager::atomic::CommitLock::acquire_blocking(&versions_dir)?;
+    if !version_dir.exists() {
+        return Err(Error::VersionNotFound(version.to_string()));
+    }
+
+    version_manager::master::invalidate_version(
+        &commit_lock,
+        &versions_dir,
+        staging.path(),
+        version,
+    )?;
+
     // Check if this is the default version
     if let Ok(default) = version_manager::get_default_version()
         && default == version
@@ -221,14 +235,7 @@ fn remove(version: &str, force: bool, json: bool) -> Result<()> {
     }
 
     std::fs::remove_dir_all(&version_dir)?;
-
-    // If the removed dir was recorded as the installed master build, clear the
-    // master sidecar record so a later `latest` resolve doesn't see a stale
-    // entry pointing at a now-deleted binary. Best-effort, like the install
-    // overwrite path: a sidecar failure must not fail a successful removal.
-    if let Ok(platform) = version_manager::platform::Platform::detect() {
-        let _ = version_manager::master::clear_record_for_version(&platform, version);
-    }
+    version_manager::atomic::sync_directory(&versions_dir)?;
 
     let out = output::RemoveOutput {
         version: version.to_string(),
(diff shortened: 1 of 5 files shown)


sdairs:
Carry-over review from PR #502 (old PR for #456). This PR is strictly more complete than #502 (no unsafe, real stale-staging GC with owner-lock probing, interruption tested on both sides of the rename, power-loss-durable fsyncs, cross-platform sidecar invalidation, single commit-lock design). No sdairs review comments existed on #502 (its only review was an LGTM), so no carry-over concerns. Three items to address before merge:

1. **`record_install` failure now hard-fails a completed install.** In `commit_staged_install_locked` (`crates/clickhousectl/src/version_manager/install.rs:~225-235`), the post-rename `record_install` uses `?` propagation. Both the pre-existing code and #502 deliberately kept the post-commit sidecar write best-effort ("a sidecar write failure must not fail an otherwise-successful install"). A sidecar write error after the binary is safely committed now reports the install as failed even though the version is usable, and the user may retry with `--force`. Restore best-effort semantics for the post-commit record write, or document why it should fail the install.

2. **`ensure_installed` pre-check inconsistency.** `install_resolved` was upgraded to check `binary_path(...).exists()`, but `ensure_installed` (`install.rs:~245-247`) still checks `version_dir.exists()`. #502 fixed both. A legacy empty version dir (from pre-fix partial installs) would be treated as installed by `server start --version`. Align `ensure_installed` to `binary_path`.

3. **Confirm the `local remove` scope expansion is intended.** Taking the commit lock and invalidating sidecar records before `remove_dir_all` in `local/mod.rs:~214-238` is beyond #456's literal criteria. It is small and same-class (concurrent remove vs install commit), defensible — but it changes remove semantics: a sidecar write error now blocks removal entirely (was best-effort). Confirm intentional.

Minor: the `checkpoint: impl FnMut(...) -> Result<()>` injection param is test-only and its `Err` arm is never exercised in production (stylistic, fine).

iskakaushik (approved):

Issue #461, opened 2026-08-22
Keep local install integration workflow triggers aligned with installer dependencies


sdairs:
## Problem

The specialized `.github/workflows/test-install.yml` workflow runs live install and URL checks, but its pull-request path filter only includes:

- `crates/clickhousectl/src/version_manager/**`
- `crates/clickhousectl/src/main.rs`
- `crates/clickhousectl/src/cli.rs`
- `crates/clickhousectl/Cargo.toml`
- the workflow itself

Installer behavior also depends on shared local CLI/dispatch/output, `paths.rs`, `error.rs`, `http.rs`, relevant subprocess tests, and lockfile dependency changes. Changes there receive general CLI CI but can skip the specialized cross-platform download/install workflow.

Using all of `src/local/**` would over-trigger on unrelated local Postgres changes, so this should be an intentional exact mapping.

## Acceptance criteria

- The workflow path filter includes each shared source, test, and lockfile path that can change installer behavior.
- The mapping is documented near the filter or checked by a focused classifier test.
- Unrelated local/Postgres-only changes do not trigger the expensive install matrix.
- Renamed or newly introduced installer dependencies fail closed or require an explicit mapping update.

PR #512, opened 2026-08-24
Keep local install workflow triggers aligned

sdairs:
## Summary

- Replace the broad local install workflow filter with an exact mapping of installer-affecting source, subprocess test, manifest, lockfile, toolchain, and workflow paths.
- Keep unrelated local and Postgres-only files explicitly classified out while covering shared dispatch, output, HTTP, path, server-discovery, and inherited version-manager dependencies.
- Add a fail-closed inventory test, run by both the broad CLI workflow and install workflow, so new or renamed non-Cloud CLI Rust and subprocess test files require an explicit mapping decision.

## Tests

- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'`
- `cargo test -p clickhousectl`
- `cargo fmt --all --check`
- `cargo clippy -p clickhousectl --all-targets -- -D warnings`
- Ruby YAML parsing for `.github/workflows/test-cli.yml` and `.github/workflows/test-install.yml`

## Stack

- Position: 5 of 5
- Base: `issue-462-install-errors-help` (#507)
- GitHub stack: #497

Closes #461

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback</a></sub>

PR #546, opened 2026-08-26
Keep local install workflow triggers aligned

sdairs:
## Summary
- expand the live install workflow path filter to the current ClickHouse installer dependency surface
- keep unrelated local Docker/Postgres sources and tests outside the expensive matrix
- add a fail-closed classifier and static workflow mapping coverage in install and broad CLI CI

## Tests
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'`\n- `python3 scripts/tests/test_classify_install_integration.py`\n- Ruby YAML parse of `.github/workflows/test-install.yml` and `.github/workflows/test-cli.yml`\n- classifier smoke checks for run, skip, and unknown fail-closed paths\n\nCloses #461


sdairs:
Carry-over review from PR #512 (old PR for #461). This PR is preferred over #512 and is approved on its own merits (meets all four acceptance criteria; CI fully green including the install matrix). No sdairs review comments existed on #512. For completeness, reasons this PR wins over #512, plus minor items:

Why preferred over #512:
- #512's inventory listed paths that don't exist in either tree (`.cargo/config`, `.cargo/config.toml`, `rust-toolchain`, `rust-toolchain.toml`) and had an internal inconsistency (`telemetry.rs` NO_INSTALL but `telemetry_test.rs` INSTALL). This PR's inventory matches its actual tree.
- This PR's `version_manager/**` glob absorbs file renames (e.g. `lock.rs` → `atomic.rs` between stack bases) without mapping churn, whereas #512's no-glob exact mapping makes every rename a fail-closed CI break requiring a mapping edit.
- #512's CI is red (from its stacked base, not its own diff); this PR is green.

Minor items (optional follow-ups):
1. The argparse CLI (`main()`/`classify_paths()` in `scripts/classify-install-integration.py:~114-128`) is never invoked by any workflow in this diff — currently dead code. It matches the `classify-cloud-integration.py` convention, so keep it if it will be wired up, otherwise note it is reserved for future use.
2. The inventory test only globs `*.rs`; non-`.rs` files under `tests/` (future snapshots) aren't covered. #512 inventoried all files. Consider broadening if snapshot tests land.
3. Unlike #512, no test pins the exact install/no-install mapping sets — a file could be moved between `INSTALL_EXACT_PATHS` and `NON_INSTALL_EXACT_PATHS` and only the workflow-equality test would catch it (it would, but less review-visible). Consider pinning the sets.

iskakaushik (approved):

PR #547, opened 2026-08-26
Validate local Postgres start options

### LINKED ISSUE #460: Validate local Postgres start options before Docker work
## Problem

`local postgres start` performs Docker recovery/connect/image work before validating several user inputs.

Deterministic Docker API reproductions on HEAD `93b3dd7` showed:

- `--port 0` inspected and pulled `postgres:18` before returning the port error.
- With an explicitly occupied host port, the CLI proceeded to container create/start and surfaced Docker port-allocation failure instead of rejecting the port locally.
- `--version 18garbage` passed validation and reached image/container operations because validation only reads leading digits.
- `-e NO_EQUALS` and duplicate generated variables were sent unchanged in the container environment.
- Recovery contacted Docker before invalid-name and unsupported-major errors, so even these invalid inputs do not guarantee zero Docker requests.

Relevant code: `crates/clickhousectl/src/local/postgres.rs:86-186,369-389` and `crates/clickhousectl/src/local/docker.rs:232-269,791-823`.

## Expected

Validate definition-owned input before recovery, Docker access, image pulls, filesystem creation, or other side effects. Omitted ports may auto-select; explicitly requested ports should never silently move.

`POSTGRES_PASSWORD` is currently an advertised `-e` use, so reserved-variable behavior needs an explicit precedence/compatibility decision rather than a blanket ban.

## Acceptance criteria

- Name, full supported tag grammar, port zero/availability, and every `KEY=VALUE` are validated before side effects.
- Invalid input makes zero Docker API requests.
- A bound explicit port returns an actionable local error before create/start.
- Generated/reserved environment variable precedence is documented and deterministic.
- Coverage includes clap parsing, bound-port tests, malformed tags/env, and a fake Docker socket asserting zero requests.

sdairs:
## Summary
- validate local Postgres names, complete supported image tags, nonzero ports, and environment assignments before command execution
- reject occupied explicit ports locally and make generated environment variable precedence deterministic while preserving `-e POSTGRES_PASSWORD=...` compatibility
- defer omitted-port auto-selection until a fresh create is confirmed, so exhausted default ports cannot block resuming a stored instance
- preserve the stored-settings note when resume ignores password overrides supplied through `--env`
- add clap, bound-port, malformed input, duplicate/reserved variable, zero-request fake Docker socket, and resume regression coverage
- update local Postgres edge-case expectations for clap-time port and version diagnostics

## Validation layering
Clap parsers own individual value validation, and the post-parse hook owns cross-argument conflicts so invalid CLI input exits with usage code 2. The handler preflight intentionally revalidates at the side-effect boundary as defense in depth for non-CLI callers; explicit ports are checked there, while omitted ports remain unresolved until the fresh-create path.

## Tests
- `cargo test -p clickhousectl --test local_postgres_start_validation_test`
- `cargo fmt --all --check`
- direct CLI verification of the updated port-zero and unsupported-version diagnostics

The full Docker-backed `scripts/test-postgres-integration.sh` was not run locally because the configured OrbStack Docker socket was unavailable.

Closes #460


sdairs:
Carry-over review from PR #486 (old PR for #460). The clap-time validation in this PR is better UX than #486's handler preflight (exit 2, env-key grammar, excellent reserved-key messages like "use `--user` instead of `--env`"), and the zero-request fake-socket criterion is met. But this PR has real regressions of work done during #486's review. Required:

1. **BLOCKER — omitted-port resolution regresses the resume fix sdairs landed on #486.** On #486, sdairs adjudicated the port-preflight concern: *"Partially valid. The exhausted omitted-port case could incorrectly block resume and is fixed in b538018: an omitted `--port` now remains unresolved through preflight and auto-selection runs only after the existing-instance branch determines this is a fresh create. I intentionally kept explicit-port validation in preflight per #460."* Here, `validate_start_options` calls `resolve_port(port)?` unconditionally (`crates/clickhousectl/src/local/postgres.rs`, `StartPreflight.host_port: u16`, not `Option`) *before* `recover_current_project_servers()` and the resume branch. An exhausted 5432–5532 auto-select range now aborts a resume that would reuse the stored binding — exactly the bug fixed in `b538018`. Fix: keep an omitted port as `Option` through preflight and auto-select only after the resume branch determines this is a fresh create (explicit ports may still validate in preflight, per the issue). Cursor Bugbot flagged this here; unanswered.

2. **Restore the resume regression test.** #486 had `exhausted_auto_port_range_does_not_block_resume` (exhausts 5432–5532, asserts resume still succeeds with the stored binding). This PR dropped it entirely — no resume-path coverage exists. Re-add it.

3. **Resume password-note regression (Cursor Bugbot #2 here, valid).** Preflight strips `-e POSTGRES_PASSWORD=` out of `extra_env` before the resume-override check, so password-via-`--env` alone no longer triggers the "resuming with stored settings" note. #486 handled this with a pre-preflight `has_extra_env` flag. Restore it.

4. **Fix the red `local postgres edge cases` CI job.** `scripts/test-postgres-integration.sh` expects `--port 0 is not allowed` (old runtime check) and the old `is not supported` version wording; this PR moved port-zero rejection into clap (`0 is not in 1..=65535`) and changed version-error wording without updating the script. Update the script expectations. Note: this also currently breaks the same CI job on every PR stacked above this one.

5. **Error text quality.** Port-zero now shows clap's generic `0 is not in 1..=65535` instead of #486's actionable "`--port 0` is not allowed; pick a specific port or omit the flag". Consider a `value_parser` wrapper that emits the actionable message.

6. Mild over-engineering note: the `main.rs` post-parse hook + `LocalArgs::postgres_start_validation_error()` plus full re-validation in `validate_start_options` duplicates what the clap parsers already enforce. #486's single preflight function was leaner. Keep if the layering is intentional defense-in-depth, but confirm.

iskakaushik (approved):

PR #550, opened 2026-08-26
Make local Postgres startup transactional

### LINKED ISSUE #464: Make local Postgres startup transactional
## Problem

Failures after creating a local Postgres container leave partial state.

Deterministic Docker API reproductions on HEAD `93b3dd7` showed:

- Create succeeded and start returned HTTP 500; the request log ended at `POST /containers/<id>/start` with no container delete.
- A simulated partial PGDATA marker survived 30 failed running-state polls. The handler stopped/deleted the container and metadata but retained the fresh data directory.
- Metadata is saved before readiness; errors while saving metadata also return without rollback.

Relevant code:

- `crates/clickhousectl/src/local/docker.rs:255-269`
- `crates/clickhousectl/src/local/postgres.rs:219-245`

A later retry may remove a leaked managed container through name recovery, but the failed invocation itself is not transactional. Whether a particular partial real PGDATA makes retries persistently fail remains workload-dependent; the retained state is directly reproduced.

## Acceptance criteria

- Any error after fresh container creation attempts container removal.
- Fresh failed initialization removes or quarantines partial PGDATA under a documented policy.
- Existing data is preserved on resume failure.
- Metadata-write and readiness failures use the same rollback boundary.
- Cleanup failure retains the primary error and adds cleanup diagnostics.
- Tests cover create-success/start-failure, initialization timeout, metadata failure, retry, and cleanup failure.

sdairs:
## Summary
- split fresh Postgres container creation from startup and keep start, metadata, and readiness behind one rollback boundary
- remove PGDATA and metadata only after the exact new container is gone, while preserving pre-existing data and recovery metadata on incomplete cleanup
- keep privileged PGDATA removal shell-free and cover metacharacter-heavy names as one literal `rm` argument
- treat Docker 404 during removal as already removed so retries do not report spurious cleanup failures
- document the fresh-start cleanup ownership policy and stabilize fake-Docker rollback coverage

## Tests
- `cargo test -p clickhousectl` (706 unit tests passed, 1 ignored; all integration binaries passed)
- `cargo test -p clickhousectl --test local_postgres_readiness_test` (12 passed)
- `cargo fmt --all --check`
- `env -u AGENT -u OPENCODE -u OPENCODE_PID scripts/test-postgres-integration.sh /Users/al/ch/ai/clickhousectl/target/debug/clickhousectl` (15 passed)

Closes #464


sdairs:
Carry-over review from PR #501 (old PR for #464). This PR's single-rollback-boundary design is clean and simpler, and resume data preservation by construction is good. But it silently drops two fixes produced by #501's review cycle. Required:

1. **BLOCKER — shell-injection sink re-introduced and now reachable from this PR's new rollback path.** On #501, macroscope flagged (Critical) that the rollback's `remove_host_dir_blocking` fallback `sh -c "rm -rf /work/{basename}"` executes shell metacharacters from a user-controlled server name and can delete sibling instances; sdairs confirmed the fix: *"Fixed in 3fec2d1… The privileged fallback now invokes `rm` directly with argv (`rm -rf -- /work/<basename>`), so the user-controlled basename is never interpreted by `sh -c`. Added focused tests for normal cleanup and a metacharacter-heavy basename remaining one literal argument."* That fix commit is not in this stack: `crates/clickhousectl/src/local/docker.rs:~999-1002` on this branch still builds `format!("rm -rf /work/{}", basename)` and runs `["sh", "-c", cmd]`, and this PR's new `rollback_fresh_postgres` (`postgres.rs:~429-432`) feeds a user-controlled name into it. `validate_server_name` blocks only `/`, `\`, NUL, `.`, `..` — `;`, `$()`, backticks, spaces all pass. Port the argv-based `rm` fix and its two unit tests (`privileged_remove_passes_metacharacters_as_one_argument`, `remove_host_dir_removes_normal_directory`).

2. **Do not delete PGDATA/metadata when container removal fails.** Here the three rollback steps are independent best-effort (`postgres.rs:~421-432`), so if `remove_container` fails, the instance dir and metadata are still deleted — then label-based recovery (`recover_current_project_servers()` → `recover_project_postgres_blocking`) resurrects the orphaned *labeled* container as a managed instance over deleted data. Cursor Bugbot flagged this (Medium) here, and the test `cleanup_failure_keeps_primary_start_error_and_adds_diagnostics` currently pins the bad behavior (`assert!(!fresh_instance_dir(...).exists())` after a failed remove). #501 deliberately retains data + recovery metadata in this case; adopt that policy and fix the test.

3. **Guard against deleting an instance dir that contained data before the attempt.** Unconditional instance-dir deletion on the fresh path loses data that predates the attempt (e.g. user deleted the metadata json — which the README calls "non-destructive" — leaving the data dir). Port #501's `fresh_instance_dir_is_disposable` predicate (~20 lines); this serves issue #464's "existing data is preserved" acceptance criterion beyond just the resume path.

4. Make `remove_container` 404-tolerant (#501 did) so removing an already-gone container doesn't add spurious "could not remove newly-created container" diagnostics.

5. Fix the red `local postgres edge cases` job (inherited from #547's script-expectation drift, but confirm resolved before this merges).

iskakaushik (approved):

PR #551, opened 2026-08-26
Validate local client selectors and reject port zero

### LINKED ISSUE #466: Validate local client selectors and reject port zero
## Problem

`local client` accepts selectors that are silently ignored or invalid:

- `--name` can be combined with `--host` or `--port`; direct mode wins and the name is ignored.
- `--port 0` parses as a valid `u16` and is forwarded to the native client.

Reproduced on HEAD `93b3dd7` with an isolated fake binary:

```console
$ clickhousectl local client --name ignored --host remote --port 0 --query Q
client
--host
remote
--port
0
--query
Q
```

Port-only direct mode is already documented and should remain defined rather than treated as an ambiguity.

## Expected behavior

Clap should reject conflicting selector modes and port zero before binary or project-state resolution.

## Acceptance criteria

- `--name` conflicts with both `--host` and `--port` in either argv order.
- `--port 0` is a usage error; ports 1 and 65535 remain valid.
- Host-only defaults to port 9000 and port-only remains documented direct `localhost` mode.
- Parse tests cover every selector combination, both orders, zero, maximum port, and a nonnumeric value.
- The same validation decision is applied deliberately to analogous local Postgres client selectors where applicable.

sdairs:
## Summary
- reject mixed named and direct local client selectors during clap parsing
- reject port zero while preserving host-only and port-only direct-mode defaults
- apply equivalent validation to local Postgres clients without changing pass-through arguments
- document selector behavior and cover parse-time and fake-child execution paths

## Tests
- `cargo fmt --all --check`

Closes #466

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,11 @@ clickhousectl local client --name dev                # Connects to "dev" server
 clickhousectl local client --query "SHOW DATABASES"  # Run a query
 clickhousectl local client --queries-file schema.sql # Run queries from a file
 clickhousectl local client --host remote-host --port 9000  # Connect to a specific host/port
+clickhousectl local client --host remote-host         # Direct mode; port defaults to 9000
+clickhousectl local client --port 19000               # Direct mode; host defaults to localhost
 ```
+
+`--name` selects a managed server and cannot be combined with direct `--host` or `--port` selectors.
 
 ### Creating and managing ClickHouse servers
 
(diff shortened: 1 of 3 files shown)


sdairs:
Carry-over review from PR #492 (old PR for #466). This PR is strictly superior to #492 and approved on its own merits: same minimal clap mechanism, the `--version`+direct-mode concern (the only review-driven change #492 needed, raised by macroscope and fixed by sdairs in `e3c74bc`) is handled from the first commit here and covered in both orders end-to-end, README docs added per repo convention, exhaustive 10-order permutation tests, passthrough-preservation guards, and filesystem-level proof that invalid input fails before any project-state creation. No further changes required for this PR itself.

One dependency note: this PR's `local postgres edge cases` CI job fails (3 cases: `install_rejects_latest`, `port_zero_rejected`, `unsupported_majors`) — inherited from the stack base (#547 changed port-zero rejection wording to clap's `0 is not in 1..=65535` and version wording, without updating `scripts/test-postgres-integration.sh`). That script reconciliation belongs to #547; once it lands and the stack rebases, this job should go green here.

iskakaushik (approved):

PR #552, opened 2026-08-26
Allow direct local client connections to select a binary version

### LINKED ISSUE #469: Allow direct local client connections to select a binary version

sdairs:
## Summary
- add direct-mode `local client --version` selection for installed ClickHouse binaries without modifying the global default
- intentionally let partial numeric selectors choose the newest installed match, while exact selectors require that installed version
- define actionable zero, one, multiple, stale-default, missing-explicit, and named-mode behavior
- document connection selection versus local binary selection and reject named-mode version overrides during clap parsing
- cover parser behavior and the complete isolated fake-binary subprocess matrix

## Tests
- `cargo fmt --all --check`

Closes #469

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,9 +226,14 @@ clickhousectl local client --queries-file schema.sql # Run queries from a file
 clickhousectl local client --host remote-host --port 9000  # Connect to a specific host/port
 clickhousectl local client --host remote-host         # Direct mode; port defaults to 9000
 clickhousectl local client --port 19000               # Direct mode; host defaults to localhost
+clickhousectl local client --host remote-host --version 26.8.1.1760  # Use an installed client binary
 ```
 
-`--name` selects a managed server and cannot be combined with direct `--host` or `--port` selectors.
+`--name` selects the connection and local client binary from managed server metadata, so named mode does not need a global default. It cannot be combined with direct `--host` or `--port` selectors, and named mode does not accept `--version`.
+
+In direct mode, `--host` and `--port` select the server connection while `--version` independently selects an already installed local client binary. Numeric selectors such as `26`, `26.8`, and `26.8.1.1760` select the newest installed match. This does not install a binary or change `~/.clickhouse/default`.
+
+Without `--version`, direct mode uses the valid default. If no default exists, zero installed versions is an error, one installed version is used without creating a default, and multiple installed versions require either `--version` or `local use`. A default that names a missing binary is an error; repair it with `local use`, or bypass it for one direct connection with `--version`.
 
 ### Creating and managing ClickHouse servers
 
--- a/crates/clickhousectl/src/error.rs
+++ b/crates/clickhousectl/src/error.rs
@@ -140,6 +140,26 @@ pub enum Error {
     #[error("No default version set. Run: clickhousectl local use <version>")]
     NoDefaultVersion,
 
+    #[error(
+        "No ClickHouse client versions are installed. Run `clickhousectl local install <version>` before making a direct connection."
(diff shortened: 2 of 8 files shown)


sdairs:
Carry-over review from PR #500 (old PR for #469). Functionally this PR is stronger than #500 (parse-time typed validation with genuinely good postgres/latest rejection messages, deliberate+documented zero/one/multiple fallback policy, broader subprocess matrix, default-file non-mutation asserted). Two required items:

1. **Trim the `--version` help noise — sdairs' only review comment on #500 applies here, more strongly.** On #500, sdairs wrote: *"good but adds too much noise to agent context help. it is unlikely that thorough explanation of a --version flag is needed beyond the actual flag description"*, and then addressed it: *"removed the redundant `--version` explanation from agent-context help, restored the existing concise connection/default wording, and retained the flag and its description."* This PR re-adds exactly that explanation in `crates/clickhousectl/src/local/cli.rs:~246-254` (`after_help`): the connection-vs-binary sentence *and* a full resolution-semantics paragraph ("Direct mode without --version uses a valid default… sole installed version… stale default must be repaired…"), and it locks the noise in with a test (`clickhouse_client_help_distinguishes_connection_and_binary_selection`, `cli.rs:~971-984`) asserting six substrings — so trimming later breaks the suite. Issue #469's criterion 5 does require help to distinguish connection selection from binary selection, so *some* text is warranted — but #500's accepted end state satisfied that with the flag description alone ("Installed local client version for direct host/port mode (e.g. 25, 25.12, or 25.12.9.61). Does not change the default.") plus the README. Cut the resolution-semantics lines from `after_help`, keep the detail in README and the flag description, and adjust the pinned test accordingly.

2. **Fix the IO-error masking (open low-severity Cursor finding here).** `try_resolve_local` (`crates/clickhousectl/src/version_manager/resolve.rs:~35`) swallows `list_installed_versions()` IO errors via `.ok()?`, so an explicit `--version` against an unreadable `~/.clickhouse/versions` reports `ClientVersionNotInstalled`, while the no-default branch propagates the IO error. Inconsistent and misleading. Propagate the error in the explicit-version path too (#500's explicit path used `list_installed_versions()?` and did not have this).

Cosmetic: partial-version "newest installed match" is a mild expansion beyond the issue's "explicit exact version" framing — it is documented and user-directed, so acceptable, but confirm intentional.

iskakaushik (approved):

PR #553, opened 2026-08-26
Preserve native query and query-file multiplicity

### LINKED ISSUE #470: Preserve native query and query-file multiplicity in local client
## Problem

The native ClickHouse 26.8.1.1760 client documents repeatable `--query` and multiple paths for `--queries-file`. The `local client` convenience flags use one `Option<String>` each, narrowing that native interface.

On HEAD `93b3dd7`, both of these fail in clap with usage exit 2 before the native client runs:

```console
clickhousectl local client --query Q1 --query Q2
clickhousectl local client --queries-file a.sql --queries-file b.sql
```

Users can work around this by moving the options after `--`, but the documented wrapper flags do not preserve native multiplicity. When one query and one file are supplied, the wrapper always emits query first and file second regardless of source argv order.

## Expected behavior

Preserve the supported native multiplicity through the convenience interface and define the interaction between inline queries and query files.

## Acceptance criteria

- Multiple inline queries can be represented without dropping or merging values.
- Multiple query files can be represented in the native client's supported argv shape.
- The behavior when query and file inputs are combined is deliberate: preserve source order where meaningful or reject the combination with a clear usage error.
- Pass-through arguments remain after wrapper-generated arguments.
- Clap and fake-child argv tests cover minimal and maximal queries/files, combined forms, ordering, empty values, and `--` pass-through.
- Supported ClickHouse client versions used by clickhousectl are checked so the wrapper does not promise an unsupported shape.

sdairs:
## Summary
- preserve repeatable `--query` plus multi-path and repeated `--queries-file` values in native client argv order
- reject mixed inline/file sources at clap time and guard repeated queries on pre-23.9 clients
- document the native contract and cover help text, empty values, multi-path files, and `--` pass-through with fake children

## Tests
- `cargo test -p clickhousectl clickhouse_client`
- `cargo test -p clickhousectl --test local_client_selectors_test`
- `cargo fmt --all --check`

Closes #470

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ clickhousectl local client --name dev                # Connects to "dev" server
 clickhousectl local client --query "SHOW DATABASES"  # Run a query
 clickhousectl local client --query "SELECT 1" --query "SELECT 2" # Run queries in order
 clickhousectl local client --queries-file schema.sql # Run queries from a file
-clickhousectl local client --queries-file schema.sql --queries-file seed.sql # Run files in order
+clickhousectl local client --queries-file schema.sql seed.sql # Run files in order
 clickhousectl local client --host remote-host --port 9000  # Connect to a specific host/port
 clickhousectl local client --host remote-host         # Direct mode; port defaults to 9000
 clickhousectl local client --port 19000               # Direct mode; host defaults to localhost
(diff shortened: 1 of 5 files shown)


sdairs:
Carry-over review from PR #506 (old PR for #470). This PR is the more complete implementation: it is the only one that implements issue #470's final acceptance criterion (version check so the wrapper doesn't promise an unsupported shape — the <23.9.1.1854 gate in `crates/clickhousectl/src/local/mod.rs:~324-337`, correctly placed after resolution for both named and direct paths, with boundary tests at 23.9.1.1853/1854). #506 punted on this with manual source inspection. Preferred over #506.

Carry-over analysis of the one substantive comment on #506: Cursor Bugbot claimed the single-flag multitoken `--queries-file` forwarding was wrong; sdairs refuted it with `ClientBase.cpp` source citations across 5 versions (`multitoken()`, so one flag with many paths is native-valid) while noting "Repeated flags may also parse, but one flag per file is not required." This PR adopts one-flag-per-path (`mod.rs:~311-313`), which is inside sdairs' validated envelope, and proves forwarding of that shape with a fake child. No adverse carry-over.

Two minor regressions vs #506 to decide on:
1. The wrapper no longer accepts multi-path-after-one-flag input (`--queries-file a.sql b.sql`, via `num_args = 1..`). The issue's failing examples used repeated flags, so acceptance criteria are still met — but confirm the narrowing is intentional.
2. #506 had a help-text assertion test; this PR does not. Consider adding one.

Note: the 23.9.1.1854 floor is sourced only via the linked changelog + pinned tests (plausible; repeated files predate repeatable queries). Fine, but keep the citation in the README as-is.

CI note: `local postgres edge cases` fails on this branch but identically on its stack base; the diff touches no Postgres code — inherited from #547's script drift, not this PR.

iskakaushik (approved):

PR #514, opened 2026-08-24
Make omitted local server selection deliberate

sdairs:
## Summary

- preserve omitted ClickHouse server names through clap and dispatch while retaining positional and compatibility `--name` forms
- make bare project-scoped `server stop` prefer an existing `default`, otherwise select the sole known ClickHouse server, succeed as a no-op with none, and require a name or `stop-all` with multiple non-default servers
- keep bare `server remove` conservative: remove only an existing `default`, never infer a custom server, and return safe available-state guidance
- distinguish implicit selection failures from explicit unknown-name typos in human and structured JSON output without expanding the bounded local error-code taxonomy
- document the stop/remove policy and update the structured-error example

Closes #473

## Tests

- isolated subprocess coverage for zero, one, many, default, custom, running, stopped, Postgres-only, positional, and compatibility `--name` states
- `cargo fmt --all --check`
- `cargo build -p clickhousectl`
- `cargo check -p clickhousectl --no-default-features`
- `cargo test -p clickhousectl`
- `cargo clippy -p clickhousectl --all-targets -- -D warnings`

## Stack

- Position: child 7 of 7 in gh-stack #489
- Base: #510 (`issue-475-local-structured-errors`)
- Parent PR is unchanged and should merge first.

PR #557, opened 2026-08-26
Make omitted local server selection deliberate

### LINKED ISSUE #473: Make omitted local server name selection deliberate and actionable
## Problem

`local server stop [NAME]` and `local server remove [NAME]` materialize the literal string `default` via clap's `default_value` when the name is omitted. Handlers cannot distinguish an omitted name from an explicit `default`, so bare `stop` in a project with only a custom-named server silently targets the wrong name and yields an unhelpful `Server 'default' not found` runtime error.

Reproduced on HEAD `93b3dd7`:

```console
# Project with only a custom-named server "dev":
$ clickhousectl local server start dev
{...}

$ clickhousectl local server stop
Error: Server 'default' not found        # exits 1, targets literal "default"

$ clickhousectl local server stop dev
{"name":"dev","already_stopped":false}    # exits 0
```

The same applies to `local server remove`. Bare `local server stop` in an empty project also yields the same unhelpful error.

Relevant code:
- `crates/clickhousectl/src/local/cli.rs:253-304` — `default_value = "default"` on name arguments
- `crates/clickhousectl/src/local/mod.rs:715-795` — receives only `String`, no omission signal
- `Error::ServerNotFound` cannot tell whether the name was explicit or implicit

## Expected behavior

Preserve `Option<String>` through clap parsing and dispatch so omitted input receives deliberate selection and diagnostics.

For stop, define safe zero/one/many behavior: stop the existing `default`, stop the sole known ClickHouse server when exactly one exists, return a successful no-op when none exist, and require a name or `stop-all` when multiple non-default servers exist.

For remove, keep deletion conservative: operate only on an existing `default`, otherwise fail with available-state guidance and safe next commands. Explicit unknown names should continue to fail.

## Acceptance criteria

- Dispatch retains whether the name was omitted, positional, or supplied through any compatibility form
- Omitted stop has documented zero/one/many/default/custom behavior
- Omitted remove never guesses among custom servers and gives actionable guidance
- Explicit unknown names continue to catch typos
- Human and JSON errors distinguish implicit selection from an explicit unknown name
- Isolated subprocess tests cover zero, one, many, default, custom, running, and stopped states

Follows up on #326 (made teardown names optional, selected literal `default`) and #324 (retained stopped server identity).

sdairs:
## Summary
- select `default` or a sole ClickHouse server deliberately for an omitted project-scoped `server stop` name, while keeping omitted `stop --global` pinned to literal `default`
- keep omitted `server remove` conservative and exclude Postgres metadata and data remnants from ClickHouse selection
- keep compatibility `--name` dispatch hidden from help and preserve positional dispatch
- intentionally include `selection` for both implicit and explicit ClickHouse stop/remove outcomes; Postgres output omits the optional field
- document the selection policy and the `server_selection_required` structured error code

## Tests
- `cargo test -p clickhousectl --test local_server_selection_test`
- `cargo test -p clickhousectl --test local_postgres_readiness_test -- --test-threads=1`
- `cargo test -p clickhousectl --test local_server_name_compatibility_test`
- `cargo test -p clickhousectl --test local_server_stopped_test`
- `cargo test -p clickhousectl --test local_structured_errors_test -- --test-threads=1`
- `cargo fmt --all --check`

Closes #473

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,7 +263,7 @@ clickhousectl local server list
 clickhousectl local server list --global                  # List running ClickHouse servers across all projects
 
 # Stop servers
-clickhousectl local server stop                           # Stop "default"
+clickhousectl local server stop                           # Stop "default", or the sole ClickHouse server
 clickhousectl local server stop dev                       # Stop by name
 clickhousectl local server stop default --global          # Stop from any project
 clickhousectl local server stop default --global --project /path/to/project  # Disambiguate
@@ -271,7 +271,7 @@ clickhousectl local server stop-all                       # Stop all ClickHouse 
 clickhousectl local server stop-all --global              # Stop all ClickHouse servers system-wide
 
 # Remove a stopped server and its data
-clickhousectl local server remove                         # Remove "default"
+clickhousectl local server remove                         # Remove "default" only; never guesses a custom name
 clickhousectl local server remove test                    # Remove by name
 
 # Write connection env vars to .env file
@@ -282,6 +282,8 @@ clickhousectl local server dotenv --local --user default --database mydb  # Incl
 ```
 
 Stopping a server preserves its data and identity metadata, so it remains visible in `server list` with a `stopped` status. Version and ports are shown only while running because they are resolved again on each start. Starting the same name resumes the existing data directory.
+
+Without a name, `server stop` selects an existing `default`, then a sole known ClickHouse server. It succeeds without changing anything when none exist, and requires a name or `server stop-all` when multiple non-default servers exist. Bare `server remove` is deliberately stricter: it removes an existing `default` only and otherwise requires an explicit name, even when there is just one custom server.
 
 **Server naming:** Without a name, the first server is called "default". If "default" is already running, a random name is generated (e.g. "bold-crane"). Pass a name positionally for stable identities you can start/stop repeatedly.
 
--- a/README.md
+++ b/README.md
@@ -979,6 +979,7 @@ The schema and meanings of existing codes are stable. New optional fields or cod
 | Code | Meaning |
 | ---- | ------- |
 | `server_not_found` | The selected local server does not exist |
+| `server_selection_required` | A server name is required because omission is ambiguous or unsafe |
 | `server_not_running` | The selected local server exists but is stopped |
 | `server_running` | The operation requires a stopped server |
 | `invalid_version` | The version selector is invalid |
--- a/crates/clickhousectl/src/error.rs
+++ b/crates/clickhousectl/src/error.rs
@@ -310,6 +310,16 @@ pub enum Error {
     #[error("Server '{0}' not found")]
     ServerNotFound(String),
 
+    #[error(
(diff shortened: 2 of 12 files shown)


sdairs:
Carry-over review from PR #514 (old PR for #473). This PR is preferred as the base: better dispatch modeling (`ServerNameInput { Omitted, Positional, NameFlag }` preserves how the name was supplied), better ordering (lock → recover → enumerate), implicit/explicit `selection` markers in both human and JSON output directly satisfy the acceptance criteria, and `{available}` counts in guidance. #514 should be closed in favor of it. But this PR regresses all five of sdairs' help-trim comments on #514 and has scope creep. Required:

1. **Re-apply sdairs' five #514 help trims verbatim (all regress here).** On #514, sdairs left five pending inline comments on `crates/clickhousectl/src/local/cli.rs`, all trimming help over-engineering; he then fixed them in commit `299d3b9` ("Trim local teardown help"): removed `--name` prose, hid the `--name` flag (`hide = true`), removed the "Idempotent…"/"typos" lines, and added a test asserting help never contains `--name`. In this PR:
   - `cli.rs:~422-423` and `cli.rs:~468-469` still carry "The compatibility form `--name dev` remains accepted, but cannot be combined with a positional name." → delete.
   - `cli.rs:~427-428` still carry "Idempotent: a server that exists but is already stopped exits 0 (no error)." and "An unknown server name still errors so typos are caught." → delete (sdairs marked both "remove"/"uncessary - does not assist in operating the CLI").
   - The `--name` flags here are NOT hidden (`cli.rs:~436`, `cli.rs:~479`, no `hide = true`) → hide them.
   - Port the `server_teardown_help_hides_compatibility_name_flags` test asserting help never contains `--name`.

2. **Revert omitted-name selection for `stop --global`.** #514 deliberately left `--global` stop unchanged (omitted name → literal `default`). This PR extends the selection policy to `stop --global` (`mod.rs` global arm), so an omitted global stop now stops a sole custom server *in another project* without a name — a real cross-project behavior change outside issue #473's scope (the issue is about project-scoped commands). Cursor Bugbot flagged "Global stop picks foreign sole server"; unanswered. Revert to fail-closed literal `default`, or explicitly justify the change in the PR description.

3. **Fix Postgres-remnant selection pollution (regression of a #514 fix).** On #514, sdairs fixed: an orphaned `*-pg*` data dir without metadata was miscounted as ClickHouse (commit `a3e2b8e`, with a regression test using a fake Docker socket proving a failed fresh Postgres start can't pollute ClickHouse selection). Here `list_clickhouse_server_names_locked` (`server.rs:~414-434`) counts such dirs again, and there is no equivalent test. Port the rollback fix (if not in this stack's base) and the fake-Docker regression test; add a Postgres-only project no-op exclusion case.

4. **Document the new JSON error code.** `server_selection_required` is emitted but the README structured-error table (head `README.md:~978`) still lists only the old codes; this PR's README diff doesn't touch the table. Add it (or remove the code if not kept).

5. Minor: `selection` is added to *all* stop/remove JSON output (including explicit calls, and Postgres outputs via `selection: None`) — a broader output-contract change than the issue requires; confirm intentional.

CI note: `local postgres edge cases` fails identically on both this and #514 — pre-existing stack-base breakage from #547, not this PR.

iskakaushik (approved):

PR #558, opened 2026-08-26
Explain managed client project scope errors

### LINKED ISSUE #467: Make local client managed-state errors explain project scope
## Problem

Managed `local client` lookup reads `.clickhouse/servers` from the exact current directory. It does not search parent directories, but failures do not identify that project scope or give a recovery path.

On HEAD `93b3dd7`, with stopped `default` metadata in a parent project:

```console
$ cd parent && clickhousectl local client
Error: Server 'default' is not running

$ cd parent/child && clickhousectl local client
Error: Server 'default' not found
```

The child-directory error does not explain that a server exists one level above or that lookup is scoped to the current directory. Coding-agent mode and explicit `local --json` still receive only the same plain error.

## Expected behavior

Managed-client failures should identify the selected mode and project scope and provide safe next commands, without changing the project-local state model.

## Acceptance criteria

- Missing and stopped managed-server errors explain that lookup used the current project directory.
- Errors suggest applicable recovery commands such as `local server list`, returning to the project root, `local server start [name]`, or direct `--host` mode.
- Agent/JSON mode exposes a stable structured error code and safe fields; human mode remains concise.
- Tests cover an empty project, a child of a valid project, an explicit wrong name, stopped metadata, and a selected missing binary.
- Telemetry remains privacy-safe: no server name, path, version, or raw message is sent.

sdairs:
## Summary
- add managed-client-specific missing, stopped, selected-binary, and project-state diagnostics with canonical exact-project scope
- keep lock and metadata failures inside the managed error contract with redacted JSON and actionable human recovery guidance
- document the dedicated `managed_client_*` JSON codes and trim constant-only scope, mode, and guidance fields
- keep server names, paths, versions, and raw errors out of telemetry

## Tests
- `cargo test -p clickhousectl --test local_client_project_scope_errors_test`
- `cargo test -p clickhousectl --test local_client_selectors_test`
- `cargo test -p clickhousectl --test local_server_stopped_test`
- `cargo test -p clickhousectl --features telemetry --test telemetry_test managed_client_failure_details_never_reach_telemetry`
- `cargo clippy -p clickhousectl --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`

Closes #467

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,8 @@ clickhousectl local client --host remote-host --version 26.8.1.1760  # Use an in
 
 `--name` selects the connection and local client binary from managed server metadata, so named mode does not need a global default. It cannot be combined with direct `--host` or `--port` selectors, and named mode does not accept `--version`.
 
+Without `--host` or `--port`, managed client lookup uses `.clickhouse/servers` from the canonical current directory only. It does not search parent directories. If lookup fails, return to the project root that owns the server, inspect that project's servers with `local server list`, or use direct mode.
+
 In direct mode, `--host` and `--port` select the server connection while `--version` independently selects an already installed local client binary. Numeric selectors such as `26`, `26.8`, and `26.8.1.1760` select the newest installed match. This does not install a binary or change `~/.clickhouse/default`.
 
 Without `--version`, direct mode uses the valid default. If no default exists, zero installed versions is an error, one installed version is used without creating a default, and multiple installed versions require either `--version` or `local use`. A default that names a missing binary is an error; repair it with `local use`, or bypass it for one direct connection with `--version`.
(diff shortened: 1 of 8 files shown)


sdairs:
Carry-over review from PR #518 (old PR for #467). This PR is better on its primary targets (clearer human text naming mode + exact dir + "parent projects are not searched" + selection-appropriate recovery; selection-aware `start` vs `start <name>` guidance; exact full-output equality subprocess tests incl. a symlinked-CWD case and an explicit no-interpolation assertion; a real E2E telemetry test stronger than #518's in-process one). But it regresses the one defect sdairs personally signed off as fixed on #518. Required:

1. **BLOCKER — lock/metadata failures escape unannotated (regression of sdairs' confirmed fix on #518).** On #518, both bots raised that `ServerLock::acquire` wasn't passed through `with_managed_client_scope`, and sdairs confirmed: *"Fixed in c12947b… `ServerLock::acquire` now maps through `with_managed_client_scope`, and the focused subprocess test covers an invalid `.locks` path in both JSON and human modes."* In this PR, the managed branch (`crates/clickhousectl/src/local/mod.rs:~300-302`) does `lock_metadata()?`, `recover_current_project_servers_locked(&metadata_lock)?`, and `server_entry_locked(server_name, &metadata_lock)?` with plain `?` — no mapping to the managed error. Those calls return raw `ServerMetadataRead/Utf8/Parse/Write`, `InvalidServerName`, io errors, or a raw `Error::ServerNotFound` (stale-reload path), with no mode/project context and no recovery commands. A corrupt `default.json` (`ServerMetadataParse`) or unwritable `.clickhouse/servers` yields exactly the unhelpful error issue #467 was opened to eliminate — violating criteria 1–2 on that failure class. The closed 3-variant `ManagedClientErrorKind` enum makes this structurally hard: there is no variant to carry a lock/metadata failure. Fix: add a fourth kind (e.g. `ProjectStateUnavailable`) wrapping `ServerMetadata*`/lock errors from those three calls, or re-introduce a scope-attachment mechanism like #518's wrapper. Also port a subprocess test mirroring #518's `invalid_lock_directory_keeps_managed_scope_in_json_and_human_errors` — there is currently no coverage for these paths at all.

2. **Decide the JSON contract churn (reviewer decision, not a blocker).** This PR introduces three new `managed_client_*` codes and a divergent envelope (drops `command`, adds `project_scope`/`server`/`guidance`) vs #518's "keep `server_not_found` + add `mode`/`project`". Nothing consumes these yet (unshipped), so it is tolerable — but choose deliberately and document in the README, since agent consumers may key on the existing `server_not_found` vocabulary.

3. Minor over-engineering: single-variant enums serialized as constants (`LocalProjectScopeKind::ExactCurrentProject`, `parent_projects_searched` which can only be `false`, `LocalClientMode::Managed`) plus a 6-action guidance enum. (#518 had a single-variant `LocalClientMode::Managed` too, so fairness noted — but consider trimming fields that can only ever hold one value.)

4. Minor: `entry.info.ok_or_else(...)` (`mod.rs:~310-312`) is effectively unreachable (`server_entry_locked_with` always returns `info: Some`); the managed branch checks binary existence then the shared code re-checks it — harmless duplication, tidy if easy.

iskakaushik (approved):

PR #559, opened 2026-08-26
Expose project scope in local server errors

### LINKED ISSUE #477: Expose project scope in local server stop/remove/remove errors
## Problem

Project-local server state is rooted at exactly `current_dir()/.clickhouse`; local commands do not discover a parent project. Server-not-found errors expose only the selected name, so the same server appears to vanish when a user or coding agent changes into a child directory.

Reproduced on HEAD `93b3dd7`. A project at `/tmp/myproject/` has a running `custom-svr`:

```console
# From project root — works:
$ cd /tmp/myproject
$ clickhousectl local server stop custom-svr
{"name":"custom-svr","already_stopped":false}

# From child directory — fails with no scope info:
$ cd /tmp/myproject/child
$ clickhousectl local server stop custom-svr
Error: Server 'custom-svr' not found
```

The child error does not identify the project directory used for lookup, state that parent projects are not searched, or suggest recovery commands such as returning to the project root or using `local server list --global`.

Relevant code:
- `crates/clickhousectl/src/init.rs:4-8` — `.clickhouse` under exact current directory
- `crates/clickhousectl/src/local/server.rs:662-706` — requires recovered process root equals exact CWD
- `crates/clickhousectl/src/error.rs:71-81` — omits project scope and recovery guidance

This is distinct from #467 which covers the same diagnostic gap for `local client` specifically. This issue covers `stop`, `remove`, and `list`.

## Expected behavior

First make exact-CWD selection visible and actionable in stop/remove/list errors. Then consider a stable `--project <PATH>` flag accepted by all project-local server commands.

## Acceptance criteria

- Not-found errors identify the canonical project directory used for lookup
- Errors state that parent `.clickhouse` directories are not searched and suggest recovery commands
- Tests cover project root versus child, explicit project selection if introduced, symlinked CWD, nested `.clickhouse`, running and stopped state

sdairs:
## Summary
- report the canonical exact-CWD project scope for local server stop and remove not-found errors
- explain missing `.clickhouse` state for omitted-name stop and remove, including the project root where the server was started
- add actionable human and structured guidance to empty project lists while keeping global list output distinct
- document the no-parent-search behavior and cover nested projects, symlinks, server state, and telemetry redaction

## Scope
- Unlike superseded PR #517, this intentionally limits new project-scoped errors to stop/remove paths that cannot find current-project state or a selected server. Running-server removal and selection errors with existing `.clickhouse` state remain self-evidently scoped and retain their existing envelopes.
- `server list` metadata and I/O failures also retain the shared redacted `io_error` envelope, so this PR does not add project-path detail to that case.

## Tests
- `cargo fmt --all --check`

Closes #477

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,8 @@ clickhousectl local server dotenv --local --user default --database mydb  # Incl
 
 Stopping a server preserves its data and identity metadata, so it remains visible in `server list` with a `stopped` status. Version and ports are shown only while running because they are resolved again on each start. Starting the same name resumes the existing data directory.
 
+Project-local server commands select `.clickhouse` under the exact current working directory. They do not search parent directories, so running `list`, `stop`, or `remove` from a child directory selects a different project scope. Change to the project root first. There is intentionally no project-path override for project-local commands; `server stop --global --project <project-root>` is only for an explicitly confirmed server found with `server list --global`.
+
 Without a name, `server stop` selects an existing `default`, then a sole known ClickHouse server. It succeeds without changing anything when none exist, and requires a name or `server stop-all` when multiple non-default servers exist. Bare `server remove` is deliberately stricter: it removes an existing `default` only and otherwise requires an explicit name, even when there is just one custom server.
 
 **Server naming:** Without a name, the first server is called "default". If "default" is already running, a random name is generated (e.g. "bold-crane"). Pass a name positionally for stable identities you can start/stop repeatedly.
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ clickhousectl local server dotenv --local --user default --database mydb  # Incl
 
 Stopping a server preserves its data and identity metadata, so it remains visible in `server list` with a `stopped` status. Version and ports are shown only while running because they are resolved again on each start. Starting the same name resumes the existing data directory.
 
-Project-local server commands select `.clickhouse` under the exact current working directory. They do not search parent directories, so running `list`, `stop`, or `remove` from a child directory selects a different project scope. Change to the project root first. There is intentionally no project-path override for project-local commands; `server stop --global --project <project-root>` is only for an explicitly confirmed server found with `server list --global`.
+Project-local server commands select `.clickhouse` under the exact current working directory. They do not search parent directories, so running `list`, `stop`, or `remove` from a child directory selects a different project scope. Change to the local project root where the server was started first; this is where `.clickhouse` typically lives. There is intentionally no project-path override for project-local commands; `server stop --global --project <project-root>` is only for an explicitly confirmed server found with `server list --global`.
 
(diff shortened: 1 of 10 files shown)


sdairs:
Carry-over review from PR #517 (old PR for #477). This PR is preferred over #517: typed `ProjectServerNotFound` error raised at the source instead of a post-hoc string wrapper; paths kept out of JSON `message` (matching the README contract, where #517 embedded the canonical path in the message and hard-overwrote `command`); explicit telemetry-redaction test (#517 lacked one); empty-list guidance via success output. Green CI where it matters. Three items:

1. **Answer sdairs' open review thread from #517 (never implemented there either, so an open design question, not a regression).** On #517, sdairs wrote: *"what about if we check for the presence of a .clickhouse dir in the CWD when a local command is run - and if there isn't one - we can tailor the output to give the message that we dont search for parent .clickhouse dirs must be run from project root"*, then scoped it: *"I think this belongs in this PR rather than a separate issue… keep it narrow to commands that require existing project state, such as `server stop` and `server remove`. When `.clickhouse` is absent from the current directory, the error should explain that clickhousectl does not search parent directories and that `.clickhouse` typically lives in the local project root where the server was started. It should not apply to commands such as `server start` that can initialize project state."* Against that:
   - Narrow to stop/remove, exclude start: **satisfied** (`ProjectServerCommand` has exactly Stop/Remove).
   - State parents are not searched: **satisfied** (`error.rs:~275`, JSON `parent_projects_searched: false`).
   - State `.clickhouse` typically lives in the project root **where the server was started**: **not satisfied** — closest text is "Return to the project root and run …" and guidance "Change to the project root that owns the server". Add the "where the server was started" framing.
   - Keyed on `.clickhouse` absence: **not satisfied** — no `local_dir().exists()` check; same message either way. The explicit-name stop/remove cases still get the full scoped error from a `.clickhouse`-less dir, but two silent paths survive: omitted-name `stop` in a non-project dir succeeds as a noop with zero scope signal (`ServerStopNoopOutput` has no scope field, `local/mod.rs:~910-918`), and omitted-name `remove` errors without scope. Decide: either attach `project_scope` to those two outputs or explicitly close the thread as addressed-by-other-means.

2. **Update the README structured-error section.** The new `project_scope`/`server`/`guidance` body and the now-absent `command` field on stop/remove `server_not_found` are undocumented (README still shows the old `{code, message, command}` example). #517 did update that section.

3. **Justify the narrower error coverage vs #517 (note in PR description).** #517 also wrapped `ServerRunningCannotRemove`, selection-required, and list metadata/IO errors. Here only the two not-found sites are scoped. Defensible (those other errors can only fire when `.clickhouse` state exists in CWD, so scope is self-evident), but list metadata errors are the one case where #517 was strictly more informative — call the narrowing out explicitly.

CI note: postgres edge-cases failures are pre-existing on the stack base (#547 script drift), not this PR.

iskakaushik (approved):

PR #521, opened 2026-08-24
Add local server state machine subprocess tests

sdairs:
## Summary

- add one table-driven real-binary suite for the complete project-local ClickHouse server state machine
- cover fresh, default, generated-name, custom-only, multi-server, and exact-CWD root/child transitions
- assert structured output and errors together with metadata, process liveness, and data-directory state in isolated temporary HOME and project directories
- disable telemetry and seed the temporary update cache so the suite does not touch user state or depend on network access

Closes #479

## Tests

- `cargo test -p clickhousectl --test local_server_state_machine_test` (4 consecutive corrected runs; 25.29s-25.53s)
- `cargo test -p clickhousectl` (639 unit tests plus all integration suites)
- `cargo fmt --all --check`
- `cargo clippy -p clickhousectl --all-targets -- -D warnings`

## Stack

- Position: child 12 of 12 in gh-stack #519
- Base: #520 (`issue-478-local-lifecycle-help`)
- Parent PR should merge first; this PR is not intended to merge independently.

PR #562, opened 2026-08-26
Add local server state machine subprocess tests

### LINKED ISSUE #479: Add table-driven subprocess test suite for local server state machine
## Problem

Local ClickHouse lifecycle tests cover important individual regressions but not the complete project state machine. Current coverage includes clap defaults/conflicts, start argument forwarding, readiness failures, retained stopped metadata, bare stop of a running `default`, and one running-remove error.

Missing transitions include: fresh projects, omitted-name selection, custom-only state, wrong CWD, repeated commands, and stopped removal. These are exactly where the August 2026 telemetry investigation found ambiguity.

Existing anchors: `local_server_start_args_test.rs`, `local_server_readiness_test.rs`, `local_server_stopped_test.rs`, and stop-state unit tests in `local/mod.rs`.

## Expected behavior

Add one table-driven isolated subprocess suite using temporary `HOME` and project directories plus the existing fake-binary/process patterns. It should exercise the shipped CLI rather than only pure state classifiers.

## Acceptance criteria

- Cover fresh list/stop/remove behavior
- Cover default start, repeated start, list, stop, repeated stop, running remove, stopped remove, and post-remove list
- Cover a complete custom-name lifecycle
- Cover omitted and explicit names across zero, one, many, default, and custom-only state
- Cover project root versus child CWD and explicit project selection if introduced
- Cover unknown names and safe actionable errors
- Assert exit status, stdout/stderr or structured JSON, persisted metadata, process state, and data-directory state
- Keep telemetry disabled and leave real user state untouched

sdairs:
## Summary
- add one table-driven, isolated subprocess suite for the local ClickHouse server state machine
- cover fresh, default, custom, omitted/explicit selection, exact-CWD, explicit global project, and generated-name transitions
- exercise bare starts through a selected fake `latest` default without network resolution
- assert exact output contracts, metadata, process liveness, data persistence/removal, telemetry isolation, update-cache isolation, and no project server state under `HOME`

## Tests
- `cargo test -p clickhousectl --test local_server_state_machine_test` (3 runs, including 2 concurrent runs)
- `cargo build --workspace`
- `cargo check --workspace --all-targets`
- `cargo fmt --all --check`
- `cargo test --workspace` (blocked by two inherited parent-owned help-text assertions in `local::cli::tests`; those files are outside this PR diff)

Closes #479

Supersedes #521.

### DIFF
--- a/crates/clickhousectl/tests/local_server_state_machine_test.rs
+++ b/crates/clickhousectl/tests/local_server_state_machine_test.rs
@@ -5,9 +5,9 @@ use std::collections::BTreeSet;
 use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Output};
-use std::time::{Duration, Instant};
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
-const VERSION: &str = "25.12.9.61";
+const VERSION: &str = "latest";
 const WAIT_TIMEOUT: Duration = Duration::from_secs(5);
 
 fn clickhousectl_binary() -> PathBuf {
@@ -73,6 +73,9 @@ impl Fixture {
         .expect("install fake ClickHouse wrapper");
         std::fs::set_permissions(&binary, std::fs::Permissions::from_mode(0o755))
             .expect("make fake ClickHouse wrapper executable");
+        std::fs::write(home.join(".clickhouse/default"), VERSION)
+            .expect("select fake latest version");
+        seed_update_cache(&home);
 
         // Process discovery is part of the shipped lifecycle. Deterministic
         // stand-ins expose only this fixture's live fake processes, avoiding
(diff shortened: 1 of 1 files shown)


sdairs:
Carry-over review from PR #521 (old PR for #479). This PR is strictly better than #521: CI green where #521 is red; it fixes #521's fatal `exec sleep` × managed-identity-inspection interaction by construction (the new base uses plain `kill(pid, 0)` liveness, and all `pgrep`/`lsof`/`ps` discovery is shimmed at the front of PATH keyed off a recorded pid file, so global-discovery scenarios are deterministic); stronger isolation (`env_clear()` + per-scenario fixtures + `ProcessGuard` + no-recorded-PID-alive check); exact full-JSON equality for responses and structured errors plus exact-string human output; full metadata-file equality; data-marker persistence across restart; and explicit `--global --project <root>` stop from a child CWD. Verified: suite passes locally (1 passed). No sdairs over-engineering/test-design comments existed on #521. Three items before merge:

1. **Honor sdairs' one substantive review comment on #521: "always use latest".** On #521, sdairs' comment was *"always use latest"* (addressed there by installing and selecting a fake `latest` default and starting without `--version`, keeping determinism without network resolution). This PR does the opposite: `const VERSION: &str = "25.12.9.61"` and every start appends `--version 25.12.9.61 --no-wait`. Determinism is preserved (local-first install finds the pre-seeded version) so the *spirit* is honored, but the bare-start / default-version-resolution path (`get_default_version()` → bare start) is never exercised, and the literal instruction is not followed. Either add a bare-start case backed by a fake `latest` default (write `.clickhouse/default` + a `versions/latest` copy of the fake binary), or document in the PR description why pinning the version is intentional.

2. **Add the omitted-repeated-start → generated-name transition.** #521 covered it ("omitted repeated start generates a name"); this PR does not exercise `resolve_name_locked`'s random adjective-noun generation anywhere. That is a real state-machine transition named in the acceptance criteria ("repeated start"). Add a case asserting a generated name is produced and usable for subsequent stop/remove.

3. **Seed the update-check cache to prevent a latent flake.** This PR's exact human-stderr assertions don't account for the cached update notice. #521 seeded `last_update_check`; this PR does not. Today latest release (v0.4.1) == `CARGO_PKG_VERSION` so `is_newer` is false and nothing prints — but once v0.4.2 ships, every `HumanSuccess`/`HumanFailure` case gains a stderr suffix and fails. Seed the cache as #521 did.

Minor: #521 also asserted project server state is never written under HOME; consider adding that assertion here.

Supersede/close #521 when this merges.

iskakaushik (approved):

PR #616, opened 2026-08-28
Document that --json is ignored by prometheus commands

### LINKED ISSUE #607: Document that --json is ignored by org/service prometheus commands

`cloud org prometheus <id> --json` and `cloud service prometheus <id> --json` accept the flag but silently emit raw Prometheus exposition text (exit 0, not parseable as JSON).

Emitting Prometheus text is the correct behavior (transforming it to JSON would lose the format contract) — but `--json` being silently accepted is confusing, especially for agent consumers that expect parseable JSON.

## Expected

State in the `--help` text for both commands that `--json` is ignored / output is always Prometheus text format.

sdairs:
## What

`cloud org prometheus <id> --json` and `cloud service prometheus <id> --json` silently accept `--json` but always emit raw Prometheus exposition text (exit 0, not parseable as JSON). This is correct behavior — converting Prometheus text to JSON would lose the format contract — but it was undocumented, which is confusing, especially for agent consumers that expect parseable JSON.

## Why

Docs-only change, no behavior change:

- Added an `after_help` block to both `OrgCommands::Prometheus` and `ServiceCommands::Prometheus` clap variants stating that output is always Prometheus exposition text and `--json` is silently ignored.
- Sharpened the README's "JSON output" section to name both commands explicitly rather than referring to "Prometheus commands" generically.

## Test plan

- Added clap `--help` parse tests next to each command definition asserting the new help text is present (`org_prometheus_help_documents_json_is_ignored`, `service_prometheus_help_documents_json_is_ignored`), following the existing pattern used for `service query` and clickpipes postgres help.
- `cargo fmt --all --check`
- `cargo test -p clickhousectl` (all green; one unrelated pre-existing timing-sensitive test, `wall_clock_timeout_fails_and_rolls_back_fresh_data`, flaked once on a slow run and passed on rerun/full-suite rerun — unrelated to this change)
- `cargo test -p clickhouse-cloud-api --test spec_coverage_test` (unaffected, all green)

Fixes #607.

Note: this PR is part of a stacked chain and is based on `main` (bottom of the stack).

### DIFF
--- a/README.md
+++ b/README.md
@@ -972,7 +972,7 @@ clickhousectl cloud --json service list
 clickhousectl cloud --json service get <service-id>
 ```
 
-`clickhousectl` auto-detects coding-agent contexts (Claude Code, Cursor, Codex, Gemini CLI, Goose, Devin, and any tool that sets the standard `AGENT` env var) and emits JSON to stdout automatically without setting `--json`. Protocol-oriented commands retain their natural output: Prometheus commands emit text, `cloud service query` uses a ClickHouse format such as `JSONEachRow`, and Postgres runtime configuration is JSON already.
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #618, opened 2026-08-28
Fix: cloud service delete surfaces NOT_FOUND instead of exiting 0

### LINKED ISSUE #601: `cloud service delete` on a nonexistent service prints `null` and exits 0

## Repro

```
chctl cloud service delete <already-deleted-service-id>
```

## Actual

Stdout: `null`, exit code 0.

## Expected

`NOT_FOUND` error with exit 1, consistent with `service get` on the same ID (`Error: NOT_FOUND: <id>`, exit 1). Exit 0 here misleads scripts into thinking the delete succeeded when the resource did not exist.

sdairs:
## What

`clickhousectl cloud service delete <nonexistent-id>` printed `null` and exited `0` instead of failing. `CloudClient::delete_service_if_exists` swallowed any 404 from the delete endpoint into `Ok(None)`, and the handler rendered that as `null` (with `--json`) or "Service X is already absent" (human output) with exit code 0.

## Why

Exit 0 on a delete of a nonexistent service misleads scripts/agents into believing the delete succeeded. `cloud service get` on the same missing ID already fails correctly with `Error: NOT_FOUND: ...` and exit 1 — delete should behave the same way.

## Fix

Replaced `delete_service_if_exists` (which special-cased 404 by additionally checking the organization still existed, then swallowing the service-not-found into `Ok(None)`) with a strict `delete_service` that maps errors via `convert_error_for_organization` exactly like `get_service` does, so a 404 propagates as a normal `CloudError` (exit code 1, `NOT_FOUND: ...` message, with the same organization-scoping note `get`/`list` already add for a bare `NOT_FOUND`).

## Tests

- Updated `forced_service_delete_treats_an_absent_key_and_service_as_success` (renamed to `forced_service_delete_surfaces_not_found_for_an_absent_service`) to assert exit 1 / `NOT_FOUND` error and that local query-key cleanup and the org-scoped key deletion are *not* attempted after a failed delete.
- Updated `service_delete_does_not_treat_a_missing_organization_as_an_absent_service` to reflect that delete no longer performs an extra organization-existence probe before failing.
- Added `service_delete_preserves_a_detailed_not_found_error`, mirroring the existing `service_get_preserves_a_detailed_not_found_error` coverage, to pin that a detailed API-provided NOT_FOUND message passes through verbatim.

Verified locally: `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p clickhousectl`, and `cargo test -p clickhouse-cloud-api --test spec_coverage_test` (unaffected crate, sanity only).

Fixes #601

Note: this PR is part of a stacked chain and targets `fix/609-local-init-json-paths` as its base branch, not `main`.

### DIFF
--- a/crates/clickhousectl/src/cloud/services.rs
+++ b/crates/clickhousectl/src/cloud/services.rs
@@ -1451,7 +1451,7 @@ async fn service_delete(
     }
 
     let response = client
-        .delete_service_if_exists(&org_id, service_id)
+        .delete_service(&org_id, service_id)
         .await
         .map_err(|error| service_delete_error(error, force, service_id))?;
     cleanup_service_query_key(client, &org_id, service_id, &query_key_ids).await?;
@@ -1460,8 +1460,6 @@ async fn service_delete(
     }
     if json {
         println!("{}", serde_json::to_string_pretty(&response)?);
-    } else if response.is_none() {
-        println!("Service {} is already absent", service_id);
     } else {
         println!("Service {} deletion initiated", service_id);
     }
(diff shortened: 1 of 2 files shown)


iskakaushik (approved):
LGTM

PR #619, opened 2026-08-28
Fix panic (exit 101) in `cloud service delete --force`

### LINKED ISSUE #598: Panic (exit 101) in `cloud service delete --force` when service is mid-stopping

## Repro

```
chctl cloud service create --name chctl-stress-cp --provider aws --region eu-west-1 --json
# wait until running, then:
chctl cloud service delete --force <id>
```

## Actual

Stdout printed `Stopping service … state: stopping`, then the process **panicked with exit code 101**. An immediate retry succeeded (exit 0, service deleted).

## Expected

Exit 0, or exit 1 with a clean error message — never a panic. Likely an `unwrap()` in the stop-then-delete polling path when the service is observed mid-`stopping`.

sdairs:
Fixes #598.

> Stacked PR: based on `fix/601-service-delete-not-found`, not `main`. Review only the top commit; the rest belongs to the PR below it in the chain.

## What was actually wrong

`cloud service delete --force` exited **101** after printing `state: stopping`. The issue guessed at an `unwrap()` in the stop-then-delete polling path. There isn't one:

- Clippy's `unwrap_used` / `expect_used` / `indexing_slicing` / `panic` lints over the non-test build report nothing in that path (the only hits in `cloud/` are an `Option::unwrap` guarded by a `1 =>` match arm and two enum parses outside this path).
- `service.state` is already read through `or_absent` and `classify_stop_poll_state`, which turn an absent state into a clean `CloudError`, and `stopping` into "keep waiting". Existing tests pin both.

The issue's exact trigger was not reproduced — the original run had a reader attached, and the interactive session is gone — so this is the only reachable panic the audit found, not a confirmed root cause. If #598 recurs after this lands, the cause is elsewhere and the issue should reopen.

That reachable panic is **the write itself**. `eprintln!`/`println!` panic when the write fails, and the write fails with `BrokenPipe` the moment whatever was reading the stream goes away. `--force` streams one progress line per state change for as long as the stop takes — minutes, dozens of lines on a real service — so it routinely outlives its reader (a pager the user quit, a supervising process that stopped reading). The panic then replaced a completed deletion with exit 101; the reported retry succeeded because the service was already stopped, so no polling output was produced. This repo has been bitten by the same class before (`main.rs` clap printing, `telemetry::print_first_run_notice`, `update::print_cached_update_notice` all carry the "not `eprintln!`, which panics on a closed stderr" comment).

Both new tests reproduce **exit 101 exactly** against the unfixed binary, and exit 0 with the fix.

## What changed

- `cloud::output::eprint_line` / `print_line`: write one line through the locked handle and discard a write failure, with the reasoning documented at the definition. Progress and result lines are not worth a panic — the exit code, not the line, reports the outcome.
- Every write in the delete path now goes through them: the "Stopping service …" notice, each poll progress line, the two query-key cleanup warnings in `api_keys.rs`, and the final `--json` / human result line (printed after the service is already gone).
- Named the poll interval `STOP_POLL_INTERVAL` while touching the loop.
- README: `--force` polls for minutes, streams state changes to stderr, and that progress output is best-effort while the deletion itself still completes and exits 0.

No request/response models, clap surface, or exit-code contract changed.

## Tests

- `forced_service_delete_survives_a_closed_stderr_while_stopping` — wiremock subprocess test; the service reports `running` → `stopping` → `stopped` during a `--force` delete while the child's stderr read end is closed mid-poll. Asserts exit 0 *and* that the stop-then-delete sequence still reached the API. Fails with `Some(101)` without the fix.
- `service_delete_survives_a_closed_stdout` — the stdout counterpart on the plain (fast) delete path. Also fails with `Some(101)` without the fix.
- Existing coverage unchanged: mid-`stopping` progress dedup, the absent-state error, `--force` clap parsing, and the not-found force path. Factored `spawn_service_delete` out of `invoke_service_delete` and shared `received_request_shape`.

Verified: `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo test -p clickhousectl` fully green (all 21 suites). `clickhouse-cloud-api` untouched.

## Follow-ups (not in this PR)

- The stop poll has no overall deadline: a service wedged in `stopping` makes `--force` wait forever rather than failing cleanly.
- Handlers outside this path still use `println!`/`eprintln!`, so a closed reader can still panic elsewhere (e.g. `cloud service query | head` exits 1 with `Broken pipe` via `write_all(...)?` — an error, not a panic, but not exit 0 either). Worth a separate issue if the project wants pipe-safe output as a global rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -629,6 +629,8 @@ clickhousectl cloud service delete <service-id>
 clickhousectl cloud service delete <service-id> --force
 ```
 
+`--force` stops the service and then polls it until the stop completes, which takes minutes on a real service, printing each state change to stderr (stdout stays reserved for the result). Progress output is best-effort: if whatever was reading it goes away — a pager you quit, a supervising process that stopped reading — the lines are dropped and the deletion still runs to completion and exits 0.
+
 Use `clickhousectl cloud service create --help` for the complete option list. If omitted, `--provider` defaults to `aws`, `--region` defaults to `us-east-1`, and the IP allowlist defaults to `0.0.0.0/0`; production workflows should normally set all three explicitly. When the create response includes an initial password, it is shown only once.
 
 `--query` and `--queries-file` are mutually exclusive. If neither is supplied, `cloud service query` reads SQL from stdin; `--queries-file -` also reads stdin explicitly.
--- a/crates/clickhousectl/src/cloud/api_keys.rs
+++ b/crates/clickhousectl/src/cloud/api_keys.rs
@@ -1,6 +1,6 @@
 use crate::cloud::client::{CloudClient, CloudError, Result as CloudResult};
 use crate::cloud::credentials;
-use crate::cloud::output::{or_absent, print_human};
+use crate::cloud::output::{eprint_line, or_absent, print_human};
 use crate::cloud::shared::{parse_datetime, resolve_org_id};
 use crate::cloud::types::DeleteResponse;
 use clap::Subcommand;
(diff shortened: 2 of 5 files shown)


iskakaushik (approved):
LGTM

PR #620, opened 2026-08-28
Warn when service update --remove-* matches nothing

### LINKED ISSUE #612: cloud service update --remove-ip-allow silently no-ops for absent IPs

## Repro

```
chctl cloud service update <svc> --remove-ip-allow 10.99.99.99/32   # not in the current list
```

## Actual

Exit 0, returns the full service JSON unchanged. Arguably idempotent, but it can hide typos — the caller cannot tell the removal matched nothing.

## Expected

Warn (or fail) when the requested IP/CIDR is not present in the current `ipAccessList`.

sdairs:
## Summary
- `cloud service update --remove-ip-allow <cidr>` (and the sibling `--remove-private-endpoint-id` / `--remove-tag` flags) silently exit 0 when the requested entry is not present in the service's current state, which hides typos from the caller.
- When any `--remove-*` flag is passed, `service update` now fetches the current service first and prints a stderr warning naming every requested value that matched nothing in the current IP allow list / private endpoint IDs / tags. Exit code stays `0` — the operation remains idempotent, it's just no longer silent about a no-op removal.
- No extra `GET` is issued when no `--remove-*` flag is passed (adds/renames only), so the common case is unaffected.
- Tag removal matching compares by key only (`--remove-tag env=prod` still matches an existing `env` tag with a different value), consistent with how the API matches tag removals.

## Test plan
- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p clickhousectl` (unit tests for `has_removals`/`unmatched`/`tag_key`/`unmatched_removal_warnings`, plus new wiremock subprocess tests in `cli_request_shape_test.rs` covering: warn on unmatched IP, no warning when matched, warnings for unmatched private-endpoint-id and tag together, and no extra `GET` when there are no removals)
- [x] `cargo test -p clickhouse-cloud-api --test spec_coverage_test` and `--lib`

Note: this PR is part of a stacked chain based on `fix/598-service-delete-force-panic` (not `main`).

Fixes #612

### DIFF
--- a/README.md
+++ b/README.md
@@ -573,6 +573,10 @@ clickhousectl cloud service update <service-id> \
   --add-tag env=staging \
   --transparent-data-encryption-key-id tde-key-1 \
   --enable-core-dumps false
+# `--remove-ip-allow`/`--remove-private-endpoint-id`/`--remove-tag` are
+# idempotent: removing something already absent still exits 0, but
+# clickhousectl warns on stderr for each value that matched nothing, so a
+# typo doesn't silently no-op.
 
 # Update replica scaling (vertical autoscaling — fixed replica count, variable memory)
 clickhousectl cloud service scale <service-id> \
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #623, opened 2026-08-28
Fix exit code for postgres config patch / reset-password usage errors

### LINKED ISSUE #613: cloud postgres config patch / reset-password exit 1 instead of 2 on missing args

## Repro

```
chctl cloud postgres config patch <id>          # no --set and no --file
chctl cloud postgres reset-password <id>        # neither --password nor --generate
```

## Actual

Exit code **1**. The documented exit-code contract is `2` for usage errors (clap), `1` for runtime errors.

## Expected

These handler-side "no change specified" checks should exit 2 (usage error), consistent with the clap validation elsewhere (e.g. `config replace` with no `--file` correctly exits 2).

sdairs:
## What

`cloud postgres config patch` with neither `--set` nor `--file`, and `cloud postgres reset-password` with neither `--password` nor `--generate`, exited **1** instead of the documented usage-error exit code **2**.

## Why

The "provide at least one of X or Y" checks lived in the handler as a runtime `CloudError`, so they went through the generic error path (exit 1) instead of clap's usage-error path (exit 2). `config replace` already gets this right because `--file` is a plain required field there.

## Fix

Added `ArgGroup`s (`required(true)`) over the existing `conflicts_with` pairs so clap itself enforces "at least one of":
- `PostgresCommands::ResetPassword`: group over `password` / `generate`.
- `ConfigCommands::Patch`: group over `sets` / `file`.

The handler-side checks are now unreachable through the CLI and were converted to `debug_assert!`/`unreachable!` safety nets, matching the existing pattern already used for the `conflicts_with` combination in `postgres_reset_password`.

## Test plan

- Added clap parse tests: `rejects_postgres_config_patch_without_set_or_file` and `rejects_postgres_reset_password_without_password_or_generate`, both asserting `clap::error::ErrorKind::MissingRequiredArgument` (which maps to exit code 2) and that the error message names both valid flags.
- Confirmed existing valid-combination tests (`--set`, `--file`, `--password`, `--generate`, and the mutual-exclusion rejections) still pass.
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p clickhousectl` all clean.

Fixes #613

Note: this PR is part of a stacked chain — its base is `fix/610-queries-file-multi-statement`, not `main`.

### DIFF
--- a/crates/clickhousectl/src/cloud/postgres.rs
+++ b/crates/clickhousectl/src/cloud/postgres.rs
@@ -1,7 +1,7 @@
 use crate::cloud::client::{CloudClient, CloudError, Result as CloudResult};
 use crate::cloud::output::{ABSENT, or_absent};
 use crate::cloud::shared::{parse_datetime, parse_serde_enum, parse_tags, resolve_org_id};
-use clap::Subcommand;
+use clap::{ArgGroup, Subcommand};
 use clickhouse_cloud_api::models::{
     ApiResponse, PgBouncerConfig, PgConfig, PgHaType, PgIdProperty, PgProvider, PgVersion,
     PostgresInstanceConfig, PostgresService, PostgresServiceListItem, PostgresServicePatchRequest,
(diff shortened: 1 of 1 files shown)


iskakaushik (approved):
LGTM

PR #624, opened 2026-08-28
Fix cloud postgres delete --json returning raw API envelope

### LINKED ISSUE #614: cloud postgres delete --json returns raw API envelope instead of resource object

## Repro

```
chctl cloud postgres delete <id> --json
```

## Actual

`{"status":200,"requestId":"..."}` — the raw API envelope.

## Expected

Consistent with every other `cloud postgres` subcommand, `--json` should emit the resource object (or a structured deletion confirmation), not the raw envelope.

sdairs:
## Summary
- `cloud postgres delete --json` was printing the raw API envelope (`{"status":200,"requestId":"..."}`) returned directly by the delete endpoint, instead of the Postgres resource object every other `cloud postgres` subcommand emits under `--json`.
- The delete endpoint's response body genuinely has no resource in `result` (it's `ApiResponse<serde_json::Value>`), so the fix fetches the service via `postgres_service_get` before issuing the delete, and renders that fetched object under `--json`. Human output (`Postgres service <id> deletion initiated`) was already not envelope-shaped, so it is unchanged.
- Note: `cloud service delete --json` has the same underlying envelope shape (its `DeleteResponse` wrapper only carries `status`/`requestId`), but that's out of scope for this issue, which is specifically about `cloud postgres delete`.

This PR is part of a stacked chain based on `fix/613-postgres-usage-exit-codes` (not `main`).

## Test plan
- [x] Added `postgres_delete_json_emits_the_resource_object_not_the_envelope` wiremock subprocess test in `crates/clickhousectl/tests/cli_request_shape_test.rs`, asserting the resource fields are present and `status`/`requestId` are absent from `--json` output.
- [x] `cargo fmt --all`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p clickhousectl` (all passing)

Fixes #614

### DIFF
--- a/crates/clickhousectl/src/cloud/postgres.rs
+++ b/crates/clickhousectl/src/cloud/postgres.rs
@@ -990,14 +990,27 @@ pub async fn postgres_delete(
     json: bool,
 ) -> CloudResult<()> {
     let org_id = resolve_org_id(client, org_id).await?;
+
+    // The delete endpoint itself only ever returns the raw API envelope
+    // (`ApiResponse<serde_json::Value>`, no resource in `result`), so fetch the
+    // resource before deleting it and render that instead: `--json` output must
+    // stay consistent with every other `cloud postgres` subcommand, which emits
+    // the resource object rather than `{"status":...,"requestId":...}` (#614).
     let resp = client
+        .api()
+        .postgres_service_get(&org_id, postgres_id)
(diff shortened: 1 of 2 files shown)


iskakaushik (approved):
LGTM

PR #626, opened 2026-08-28
Validate cloud postgres list --filter keys and values

### LINKED ISSUE #603: cloud postgres list --filter silently ignores unknown/empty keys

## Repro

```
chctl cloud postgres list --filter isPrimary=false   # returns ALL services, exit 0
chctl cloud postgres list --filter state=            # empty value: returns ALL, exit 0
```

## Actual

Any filter key outside `{state, region, name, provider}` (including `isPrimary`, which is shown as the `Primary` table column) silently returns the full list with exit 0 (`_ => true` at `crates/clickhousectl/src/cloud/postgres.rs:567`).

Related: the `state` filter compares `format!("{:?}", s)` (Rust Debug form) against the wire value — single-word states work, but `Unknown(...)` values or any multi-word form will not match. Compare against the serde wire value instead.

## Expected

Unknown filter keys → usage error (or documented valid-key list); empty value → empty result or usage error.

sdairs:
Fixes #603.

Stacked PR: base is `feat/615-postgres-delete-force`, not `main`.

## What was broken

`cloud postgres list --filter` matched keys with a catch-all `_ => true`, so any key outside `{state, region, name, provider}` — including `isPrimary`, which the listing shows as the `Primary` column — silently returned the **whole unfiltered list** with exit 0. A filter with an empty value (`state=`) or no `=` at all was ignored the same way. The `state` and `provider` predicates also compared `format!("{:?}", ..)` Debug forms against the wire value, so an `Unknown(..)` state could never match.

## What changed

- `--filter` is parsed at clap time by a `value_parser` into a closed `PostgresListFilter` enum. An unknown key, a missing `=`, an empty/whitespace-only value, or a non-boolean `isPrimary` is now a **usage error (exit 2)** whose message lists the valid keys — and no API request is issued.
- `isPrimary` is a supported key (`true`/`false`, plus the `yes`/`no` the `Primary` column renders).
- `state` and `provider` are compared against the serde wire value via `Display`, case-insensitively, so multi-word states and `Unknown(..)` values match correctly. `region`/`name` match exactly; keys themselves are case-insensitive.
- Per CLAUDE.md, an absent (API-omitted) response field matches no filter value, and all filters must hold for an item to be kept.
- README documents the supported keys, the matching rules and the exit-2 behaviour.

`cloud service list --filter` is **not** affected: it forwards server-side resource-tag filters (`tag:env=production`) to the API as query parameters and shares none of this code, so client-side key validation there would wrongly reject valid server-side syntax.

## Tests

- Clap parse tests next to the command definition: the accepted forms (including case-insensitive keys and `yes`/`no`), and one test per rejection asserting `ErrorKind::ValueValidation` and `exit_code() == 2`.
- Inline `apply_filter` tests: wire-value comparison for `state`/`provider` (including `Restoring_backup` and `Unknown(..)`), absent fields matching nothing, multi-filter conjunction, and the empty-filter passthrough.
- `tests/cli_request_shape_test.rs`: subprocess + wiremock tests pinning exit 2 with an `expect(0)` mock for invalid filters (no API call, no stdout), and the client-side filtering results for `isPrimary`, `state=restoring_backup`, `region` and a non-matching `name`.

Verified with `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test -p clickhousectl` (all green). The `clickhouse-cloud-api` crate is untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -674,6 +674,7 @@ Manage ClickHouse Cloud managed Postgres services. All write commands require AP
 # List / get
 clickhousectl cloud postgres list
 clickhousectl cloud postgres list --filter state=running
+clickhousectl cloud postgres list --filter region=us-east-1 --filter isPrimary=true
 clickhousectl cloud postgres get <pg-id>
 
 # Create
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #628, opened 2026-08-28
Fix clickpipe scale usage error for missing scale flags

### LINKED ISSUE #605: clickpipe scale with no flags returns Internal error instead of usage error

## Repro

```
chctl cloud clickpipe scale <svc> <pipe>   # no --replicas/--cpu-millicores/--memory-gb
```

## Actual

Exit 1, `Error: Internal error`. The server likely 400s on the empty body and the CLI wraps it generically.

## Expected

A client-side usage error (exit 2): provide at least one of `--replicas` / `--cpu-millicores` / `--memory-gb`.

sdairs:
## Summary
- `cloud clickpipe scale <svc> <pipe>` with none of `--replicas`/`--cpu-millicores`/`--memory-gb` sent an empty PATCH body, which the API 400s on and the CLI surfaced generically as `Error: Internal error` (exit 1).
- Added a required, `multiple(true)` clap `ArgGroup` over the three flags on the `Scale` command variant so the empty case is now rejected as a client-side usage error (exit 2) before any request is sent, naming the valid flags in the usage message.

Stacked PR: based on `fix/604-postgres-switchover-promote`, not `main`.

## Test plan
- [x] Clap parse tests next to the `Scale` command definition: zero flags rejected, each single flag accepted, and a multi-flag combination accepted (`crates/clickhousectl/src/cloud/clickpipes.rs`).
- [x] Updated the existing read/write classification test and the previous "no flags" parse test, which both relied on the old permissive behavior.
- [x] Added subprocess wiremock tests in `crates/clickhousectl/tests/cli_request_shape_test.rs`: no-flag invocation exits 2 without hitting the mock API; single-flag invocation sends the expected PATCH body.
- [x] `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p clickhousectl` all pass.
- [x] Updated `README.md` clickpipe scale example to note the flag requirement.

Fixes #605

### DIFF
--- a/README.md
+++ b/README.md
@@ -768,7 +768,7 @@ clickhousectl cloud clickpipe resync <service-id> <clickpipe-id>   # CDC pipes o
 # Delete a ClickPipe
 clickhousectl cloud clickpipe delete <service-id> <clickpipe-id>
 
-# Update scaling
+# Update scaling (at least one of --replicas/--cpu-millicores/--memory-gb is required)
 clickhousectl cloud clickpipe scale <service-id> <clickpipe-id> \
   --replicas 2 --cpu-millicores 250 --memory-gb 1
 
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #630, opened 2026-08-28
Omit Kafka-only ClickPipe settings for non-Kafka pipes

### LINKED ISSUE #602: clickpipe settings update broken for all non-Kafka pipes

## Repro

```
chctl cloud clickpipe settings update <svc> <object-storage-pipe> --object-storage-max-file-count 200
```

## Actual

Exit 1: `Setting 'kafka_read_committed' is only supported for Kafka ClickPipes`. This happens for **every** non-Kafka pipe regardless of flags.

Root cause: the handler unconditionally GETs settings and re-sends `kafka_read_committed` (defaulting `false` when absent) in every PUT — `crates/clickhousectl/src/cloud/clickpipes.rs:1696-1715`.

## Expected

Omit kafka-only fields when the pipe source is not Kafka.

sdairs:
## What

`clickhousectl cloud clickpipe settings update` failed for **every** non-Kafka pipe, whatever flags were passed:

```
$ chctl cloud clickpipe settings update <svc> <object-storage-pipe> --object-storage-max-file-count 200
Error: Setting 'kafka_read_committed' is only supported for Kafka ClickPipes
```

The handler unconditionally read the current settings and re-sent `kafka_read_committed` (defaulting to `false` when absent) in every PUT.

`kafka_read_committed` was a required `bool` on `ClickPipeSettingsPutRequest`, so the key could not be left off the wire at all. The spec schema carries no `required[]`, so requiredness resolved from the `"Optional"` description heuristic. The field is now `Option<bool>` with `skip_serializing_if`, plus a documented `optionality_exemptions` entry, so the drift analyzer keeps checking field presence and enum values without demanding the key on every request (same shape as the `ClickPipePostKafkaSource.authentication` exemption on the base branch).

## Behaviour after the fix

- The handler fetches the pipe first and classifies its source.
- **Kafka pipes**: unchanged — the current `kafka_read_committed` is read back from the settings endpoint and re-sent, so the PUT does not reset it.
- **Every other source** (object storage, Postgres, MySQL, MongoDB, Kinesis, BigQuery — and a response that drops `source` entirely): the key is omitted, and the settings GET is skipped.
- Audit of the other settings: they are all source-agnostic in the request and, as before, only appear on the wire when the user passed the matching flag, so the API validates applicability per source. Only `kafka_read_committed` was unconditional.
- Request construction moved into `build_clickpipe_settings_request` over a `ClickPipeSettingsValues` struct, replacing the twelve-argument handler.

## Tests

- Builder unit tests: non-Kafka omission (including an empty body when no flags are given), every flag set with a preserved Kafka value (`true` and `false`).
- Unit tests for the Kafka / object-storage / Postgres / absent-source classification.
- Wiremock subprocess tests (`tests/cli_request_shape_test.rs`): object-storage, Postgres and source-less pipe updates send no kafka field and issue only `GET pipe` + `PUT settings`; a Kafka pipe update still issues `GET pipe` + `GET settings` + `PUT settings` and preserves the value.
- Library tests: an absent `kafka_read_committed` is omitted rather than serialized as `null`; the settings PUT body is no longer strict while the create-position `ClickPipeSettings` still is.
- README note plus a README-pinning test.
- Verified: `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p clickhousectl`, `cargo test -p clickhouse-cloud-api --lib --test models_test --test client_test --test spec_coverage_test`, `cargo test -p clickhouse-openapi-analyzer`, `cargo check --workspace --all-features`.

## Stacking

Part of a stacked PR chain: based on `fix/606-kafka-no-auth`, not `main`.

Fixes #602

### DIFF
--- a/README.md
+++ b/README.md
@@ -778,6 +778,14 @@ clickhousectl cloud clickpipe settings update <service-id> <clickpipe-id> \
   --streaming-max-insert-wait-ms 10000
 ```
 
+`settings update` only sends the settings you name on the command line, and it
+first reads the pipe to find its source type: settings that the API supports for
+one source only — currently the Kafka `kafka_read_committed` setting, which the
+CLI preserves rather than exposing as a flag — are sent for Kafka pipes and
+omitted for every other source. Object-storage-, streaming- and
+ClickHouse-specific settings are validated by the API, so passing (for example)
+`--object-storage-max-file-count` to a Kafka pipe is rejected server-side.
+
 #### Creating ClickPipes
 
 Each source type has its own subcommand under `clickpipe create`:
--- a/crates/clickhouse-cloud-api/src/models/clickpipes.rs
+++ b/crates/clickhouse-cloud-api/src/models/clickpipes.rs
@@ -3392,8 +3392,15 @@ pub struct ClickPipeSettingsPutRequest {
     pub clickhouse_parallel_distributed_insert_select: Option<i64>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub clickhouse_parallel_view_processing: Option<bool>,
-    #[serde(rename = "kafka_read_committed")]
-    pub kafka_read_committed: bool,
+    /// Kafka-only: the API rejects this key outright for any other source
+    /// ("Setting 'kafka_read_committed' is only supported for Kafka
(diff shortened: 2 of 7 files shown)


iskakaushik (approved):
LGTM

PR #632, opened 2026-08-28
Guard `local remove` against running servers in any project

### LINKED ISSUE #600: `local remove` running-server guard only checks the current project

## Repro

1. Start servers on the default version from project dir A (`local server start dev` etc.).
2. From project dir B, run `chctl local remove <that-version>`.

## Actual

Exit 0 — the binary is deleted from under 3 running servers. The servers keep running (open fds), but `local client -n dev` then fails with `managed_client_binary_not_found`.

Root cause: `remove()` (`crates/clickhousectl/src/local/mod.rs:200-205`) calls `recover_current_project_servers()` then filters `list_running_servers()`, which is project-scoped. Servers in other projects are invisible. The guard works correctly when run from the same project (exit 1, `server_running`).

## Expected

`local remove --help` says "Fails if a local server is currently running on this version" — enumerate servers across all projects (the `local server list --global` machinery already exists).

sdairs:
## What

`local remove`'s running-server guard was project-scoped: it called `recover_current_project_servers()` and filtered `list_running_servers()`, both of which only see `.clickhouse/servers/` under the current working directory. A server started from another project directory was invisible, so `local remove <its-version>` exited 0 and deleted the binary from under it — the server kept running on its open fd, but `local client -n <name>` in that project then failed with `managed_client_binary_not_found` (issue #600).

The guard now enumerates running servers across **every** project using the same process-discovery machinery behind `local server list --global`, merged with this project's metadata and de-duplicated by PID:

- Both sources are consulted because each covers a gap in the other: metadata is project-scoped but survives a failed process scan, while discovery spans projects but depends on `pgrep`/`lsof`/`/proc`.
- A single discovery pass feeds both orphan recovery and the global view, so the guard does not shell out to `pgrep`/`lsof` twice (`recover_from_discovered_locked` / `global_entries`).
- A discovered process whose version cannot be read is treated as non-matching — blocking every removal on an unreadable command line would be unfixable by the user.

## Why the error and `--force` changed shape

- Refusal is still exit `1` / JSON `server_running`, but the message now names each blocking server as `'<name>' in <project-root> (PID <pid>)`, in both the human error and the structured one, and points at `clickhousectl local server list --global`. Without the project root, a blocker outside the current directory is unactionable.
- `--force` stops blockers wherever they run: through metadata in this project (so the recorded state stays consistent) and by PID elsewhere, exactly like `server stop --global`. Its confirmation line now names the project.

## Tests

- `src/local/server.rs` unit tests for the pure helpers `select_version_users` (no match, other-project match, same server via both sources reported once, merge of both sources, unreadable version, Postgres container never matching a ClickHouse version) and `describe_version_users`.
- New subprocess suite `crates/clickhousectl/tests/local_remove_global_guard_test.rs`: isolated `HOME`, three fake installed versions, two project dirs, and deterministic `pgrep`/`lsof`/`ps` stand-ins on `PATH` (same technique as `local_server_state_machine_test.rs`, so the scan can never see or kill anything else on the machine). Covers the refusal, the structured error and its `command`, `--force` stopping the other project's server before removal, the unchanged same-project path (including metadata left at `pid: 0`), and an unrelated version still removing cleanly while the other project keeps running.
- All four cross-project tests fail against the previous guard (verified by stashing the handler change).
- `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p clickhousectl` all clean.

## Docs

`README.md` and the `local remove` agent help now state that the check spans every project and that the error names the project root and PID.

## Stacked PR

Part of a stacked chain: this PR is based on `fix/599-local-remove-default-guard` (the default-version guard in the same function), not `main`.

Fixes #600

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,12 +172,12 @@ clickhousectl local list --remote           # Available for download
 
 # Remove a version
 clickhousectl local remove 26.8.1.1760
-clickhousectl local remove 26.8.1.1760 --force   # Stop running servers on this version, and remove it even if it is the default
+clickhousectl local remove 26.8.1.1760 --force   # Stop running servers on this version (in any project), and remove it even if it is the default
 ```
 
 `local use` also creates a symlink at `~/.local/bin/clickhouse` pointing to the selected version's binary, so the plain `clickhouse` command (e.g. `clickhouse local`, `clickhouse client`) is on PATH. Pass `--no-global` to skip. If a regular file already exists at that path it is left alone with a warning.
 
-`local remove` refuses to delete a version while a local server is running on it (it would leave the server pointing at a deleted binary), failing with the running server names. Stop the server first, or pass `--force` to stop the running server(s) and then remove the version.
+`local remove` refuses to delete a version while a local server is running on it (it would leave the server pointing at a deleted binary), failing with exit `1` and JSON error code `server_running`. Because versions are shared between projects, the check spans **every** project, not just the current directory: the error names each blocking server with the project root it was started from and its PID, so a server found by `clickhousectl local server list --global` is identifiable. Stop those servers first (`clickhousectl local server stop --global <name>`), or pass `--force` to stop them — in whichever project they run — and then remove the version.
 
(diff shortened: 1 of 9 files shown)


iskakaushik (approved):
LGTM

PR #634, opened 2026-08-28
Record positional argument presence in anonymous telemetry

### LINKED ISSUE #480: Record positional argument presence in anonymous telemetry
## Problem

Anonymous telemetry records command paths and non-positional flag names, but deliberately drops all positional presence. This makes materially different lifecycle invocations indistinguishable:

| Invocation | Command | Flags |
| --- | --- | --- |
| `local server stop` | `local server stop` | `[]` |
| `local server stop dev` | `local server stop` | `[]` |
| `local use latest` | `local use` | `[]` |

During the August 2026 telemetry investigation, this prevented distinguishing omitted-default friction from explicit stale names and prevented telling bare `local use` parse failures from handler failures after a version was supplied.

Relevant code: `crates/clickhousectl/src/telemetry.rs:440-481` resolves `Arg` definitions and `ValueSource::CommandLine`, then skips positionals entirely.

## Expected behavior

Record only definition-owned positional IDs whose source is the command line, never values. A bounded field such as `positionals: ["name"]` or `positionals: ["version"]` preserves the no-values/no-identifiers privacy model while making command shape observable.

Where clap defaults or generated values matter, consider a separate closed definition-owned source field such as `name_source = omitted_default | positional | name_flag | generated`. Do not record the selected name.

## Acceptance criteria

- Successful parsed invocations record only allowlisted positional definition IDs present on the command line
- Positional values remain structurally unreachable to telemetry serialization
- Defaults, environment values, generated names, passthrough arguments, and values after `--` are not misclassified as user positionals
- Hostile secret-looking positional fixtures never appear in payloads or snapshots
- Tests distinguish bare stop from named stop and supplied-version use/remove from missing-required parse outcomes
- Telemetry documentation states the privacy boundary and migration effect on dashboards

Related: #283 established the anonymous no-value contract; #342 fixed lossy parsing after positionals without recording their presence; #450 covers privacy-safe runtime failure classification.

sdairs:
## What

Anonymous telemetry recorded command paths and flag names but dropped **all** positional presence, so materially different invocations produced identical events: `local server stop` vs `local server stop dev`, and a bare `local use` parse failure vs a handler failure after a version was supplied. That is what blocked the August 2026 telemetry investigation (#480).

The payload gains a `positionals` field carrying the **definition-owned ids** of the positional slots the user filled — presence only, never a value:

```json
{"command":"local server stop","flags":[],"positionals":["name"],"outcome":"ok", ...}
{"command":"local server stop","flags":["name"],"positionals":[],"outcome":"ok", ...}
{"command":"local use","flags":[],"positionals":[],"outcome":"missing_required", ...}
{"command":"local remove","flags":[],"positionals":["version"],"outcome":"error", ...}
```

## Why it stays anonymous

Every entry is cloned from an `Arg` definition compiled into the binary — the same "definition-owned, never the user's input" anchoring the existing `suggestion` field uses — so the field's vocabulary is a closed set that structurally cannot carry argv. Three exclusions keep it honest:

- only `ValueSource::CommandLine` slots count, so clap defaults, environment-fed values, and handler-generated names stay absent. That is exactly what makes "the user named it" and "we picked one" distinguishable, so no separate `name_source` field was needed — the compatibility `--name` form already lands in `flags`, so the two naming styles are already distinct.
- passthrough slots are skipped via a **structural** marker test (`last`, `trailing_var_arg`, `allow_hyphen_values`) rather than a hand-maintained id list, so argv forwarded to `clickhouse-server` / `clickhouse-client` / `psql` — including everything after `--` — is never classified as this CLI's shape, and new passthrough slots inherit the exclusion.
- on the failed-parse path the walk already stops at `--` and never records the unmatched token; it now records only the slot the token reached.

## Tests

- Unit (`src/telemetry.rs`): bare vs positional vs `--name` stop; supplied vs missing-required `local use`/`local remove`; defaulted positionals excluded; value terminators record nothing; passthrough exclusion on both capture paths; a hostile secret-shaped positional fixture matrix (AWS key, `sk-live-…`, Postgres URI with password, S3 path, SSH key path, GitHub token) asserted absent from serialized payloads; and a structural guard walking the real command tree that every recorded id is a source-level identifier and every passthrough slot is recognized.
- End-to-end (`tests/telemetry_test.rs`): subprocess + wiremock tests pinning the bare/named/flag-named stop distinction, missing-required vs supplied version, and no hostile value on the wire; existing payload-shape assertions extended with `positionals`.
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p clickhousectl` (all green), plus `cargo check -p clickhousectl --no-default-features --all-targets` for the telemetry-compiled-out configuration.

## Docs

The first-run notice, the `telemetry` subcommand's agent help, and the README telemetry section now state the positional privacy boundary and its exclusions (notice/help wording changed from "never values or arguments" to "flag and argument names (never their values)").

## Deploy note

The ingest worker's field allowlist needs a matching deploy before the next CLI release, or `positionals` will be dropped server-side (precedent: #320 / #322, which added `outcome`/`suggestion`). Dashboards built on pre-`positionals` events keep working — the field is additive and absent from older events.

Part of a stacked PR chain: based on `fix/608-local-json-error-parity`, not `main`.

Fixes #480

### DIFF
--- a/README.md
+++ b/README.md
@@ -1202,13 +1202,20 @@ The CLI checks for updates in the background (at most once per 24 hours) and cac
 Each event contains exactly:
 
 - the command path (e.g. `local server start`)
-- the **names** of the flags passed (e.g. `json`, `org-id`) — never flag values, never positional arguments
+- the **names** of the flags passed (e.g. `json`, `org-id`) — never flag values
+- the **names** of the positional arguments passed (e.g. `name` for `local server stop dev`) — presence only, never the value
 - how the invocation ended and its exit code
 - the CLI version, OS, and architecture
 - whether it ran in CI (`CI` env var)
 - whether it ran under a detected coding agent, and if so which one (e.g. `claude-code`)
 
 There is no install ID, no device ID, and no fingerprinting of any kind. The payload is built from the clap command definitions rather than the raw command line, so leaking an argument value is structurally impossible — the code that builds the event has no access to values at all.
+
+The privacy boundary for positional arguments is exactly the same one as for flags: every recorded name is a string compiled into the binary, so the field's vocabulary is a closed set that cannot carry anything you typed. `clickhousectl local server stop analytics-prod` records `positionals: ["name"]` — the fact that a server was named, not which one. Three further exclusions keep the field honest:
+
+- only arguments you actually passed count, so a value clap filled in from a default (or from the environment), and a name the CLI generated for you, are absent — which is what makes "you named it" and "we picked one" distinguishable
+- arguments forwarded to another program are never recorded: everything after `--` for `local server start`, and the trailing arguments of `local client` and `local postgres client`, belong to `clickhouse-server`, `clickhouse-client`, and `psql`
+- when a command fails to parse, the unmatched token is still never recorded — only the slot it would have filled
 
 Nothing is sent before you have seen the notice unless you explicitly enable telemetry with `clickhousectl telemetry enable`. The first run normally prints a one-time notice to stderr, records that it was shown in `~/.clickhouse/telemetry.json`, and sends nothing. Sending starts from the following run. Explicitly enabling telemetry starts it immediately and skips the notice. The send happens in a short-lived detached process, so command latency is unaffected even when the endpoint is unreachable.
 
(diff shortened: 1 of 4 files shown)


iskakaushik (approved):
LGTM

Issue #425, opened 2026-08-18
Validate hourly Cloud backup start times before sending the request


sdairs:
## Summary

The CLI accepts any syntactically valid `HH:MM` backup start time, while the Cloud API only accepts times on the hour (`HH:00`). Invalid values make an avoidable API request and the help implies broader support than the endpoint provides.

## Reproduction

```console
clickhousectl cloud service backup-config update <service-id> \
  --backup-period-hours 24 \
  --backup-retention-period-hours 48 \
  --backup-start-time 02:30
```

The CLI accepts the arguments, then the API rejects them:

```text
Error: BAD_REQUEST: customBackupStartTime must be a valid time (HH:00)
```

`02:00` succeeds.

A related combination, a custom start time with `--backup-period-hours 12`, is also accepted locally and rejected because the period must be 24 or 48 hours when a start time is set.

## Expected behavior

Validate these constraints before making the request and describe the accepted values in `--help`:

- Start time must match `HH:00`.
- When a custom start time is present, the backup period must be 24 or 48 hours.

## Environment

Tested with `clickhousectl 0.4.1` built from PR #419 head `164d98154e77a491611c9a3b7de652d2b9caaffd` against a disposable ClickHouse Cloud service.

PR #645, opened 2026-09-01
Correct the `--backup-start-time` help and check the stored backup period

### LINKED ISSUE #642: `--backup-start-time` help says an omitted period is set to 24 hours; the API keeps the stored period and rejects the request
## Summary

`cloud service backup-config update --help` (from #438, on `main`) says:

> Backup start time in UTC, exactly on the hour (HH:00). If the period is omitted, the API sets it to 24 hours.

That is not what the API does. When `--backup-start-time` is sent without `--backup-period-hours`, the request omits `backupPeriodInHours`, and the API validates the new start time against whatever period is **already stored**. If the stored period is not 24 or 48 the update is rejected. Nothing is defaulted.

## Reproduction

Observed against a live service on 2026-09-01 (`clickhousectl 0.4.2` build from `main`):

```console
# stored period is 12
$ clickhousectl cloud service backup-config get <id> --json
{ "backupPeriodInHours": 12.0, "backupRetentionPeriodInHours": 48.0 }

# start time only, as the help text invites
$ clickhousectl cloud service backup-config update <id> --backup-start-time 03:00 --json
Error: BAD_REQUEST: customBackupPeriod must be 24 or 48 hours when customBackupStartTime is set
exit=1

# stored period unchanged, no start time stored
$ clickhousectl cloud service backup-config get <id> --json
{ "backupPeriodInHours": 12.0, "backupRetentionPeriodInHours": 48.0 }

# with stored period 24 the same command succeeds and the period stays 24
$ clickhousectl cloud service backup-config update <id> --backup-period-hours 24 --json
$ clickhousectl cloud service backup-config update <id> --backup-start-time 03:00 --json
{ "backupPeriodInHours": 24.0, "backupRetentionPeriodInHours": 48.0, "backupStartTime": "03:00" }
```

The "24 hours" in the successful case is the stored value being preserved, not a default being applied.

## Where

`crates/clickhousectl/src/cloud/backups.rs`, the doc comment on `backup_start_time` in the `Update` variant (line ~82 on `main`), and the test at line ~547 that pins the string `"API sets it to 24 hours"`.

## Expected behavior

Help text that matches the API:

> Backup start time in UTC, exactly on the hour (HH:00). Requires the backup period to be 24 or 48 hours: pass `--backup-period-hours 24|48` in the same call, or the stored period must already be one of those.

Optionally, the CLI could go further: when `--backup-start-time` is given without a period, read the current configuration first and fail client-side with a clear message if the stored period is not 24 or 48. That would also cover the mirror case noted in #564 (period-only update to a non-24/48 value while a start time is stored), which today also fails at the API.

## Acceptance criteria

- Help no longer claims a default is applied.
- The pinned help-text test is updated.
- README backup section says the same thing if it repeats the claim.

Related: #425 (original validation issue), #438 (the PR that added the text), #564 (start time cannot be cleared). Found during the 0.4.2 re-validation of ClickHouse/ClickHouse#116775.

sdairs:
## What

- Fixes the doc comment on `backup_start_time` in the `backup-config update` command. It said "If the period is omitted, the API sets it to 24 hours", which is false. It now says the start time requires the backup period to be 24 or 48 hours, either passed in the same call or already stored on the service.
- Updates the pinned help-text test so it asserts the new wording and actively asserts the old claim is gone.
- Adds a client-side pre-check: when `--backup-start-time` is given without `--backup-period-hours`, the handler reads the current backup configuration through the existing `CloudClient::get_backup_config` wrapper and fails before the PATCH if the stored `backupPeriodInHours` is present and is neither 24 nor 48. The message names the numbers and the flag to pass, and nothing else. An absent stored period is not guessed at: the update proceeds and the API decides.
- Documents the 24/48 rule in the README backup-configuration section.

The existing client-side check for an explicitly supplied incompatible period (`--backup-period-hours 12 --backup-start-time 02:00`, issue #425) is unchanged and still fires before any request.

## Why

The API keeps the stored period when an update omits one, then validates the new start time against it. On a service whose period is 12 the documented invitation to pass a start time on its own produces `BAD_REQUEST: customBackupPeriod must be 24 or 48 hours when customBackupStartTime is set` and no state change. The help text was inviting a request the API will not accept, and the resulting error names an internal field rather than a flag. Reading the configuration first turns that into an actionable local failure, and costs one GET only on the exact path that could fail this way.

## Behaviour after the fix

- Start time only, stored period 12: one GET, no PATCH, exit 1, `Error: the stored backup period is 12 hours, but --backup-start-time requires 24 or 48. Pass --backup-period-hours 24 or --backup-period-hours 48 in the same call.`
- Start time only, stored period 24 or 48: GET then PATCH, whose body carries `backupStartTime` and no `backupPeriodInHours`, so the API keeps the stored period.
- Start time plus `--backup-period-hours 24|48`: no GET, a single PATCH with both fields.
- Start time plus any other period: exit 1 with no requests at all.
- Stored period absent from the response: no local refusal, the PATCH is sent.

`--json` output is unchanged on success, and the failure exit code stays 1.

## Tests

- New unit tests for `check_stored_period_allows_start_time` covering 24 and 48, an absent stored period, an exact message assertion for 12, and a fractional value.
- The pinned help-text test now requires the new wording and forbids the old sentence.
- New subprocess and wiremock tests in `crates/clickhousectl/tests/cli_request_shape_test.rs`: stored period 12 refuses with only the GET recorded, stored period 24 issues GET then a PATCH whose body omits `backupPeriodInHours`, and an explicit period 48 issues the PATCH with both fields and no read at all. The pre-existing test for an explicit period 12 already covers the no-request case.
- Verification commands, all clean: `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo clippy -p clickhousectl --all-targets --no-default-features -- -D warnings`, `cargo test -p clickhousectl` (all binaries green, 868 unit tests plus every integration target). The API library was not touched.

## Stacking

Part of a stacked PR chain: based on `feat/450-telemetry-failure-classification`, not `main`.

Fixes #642

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,6 +637,8 @@ clickhousectl cloud service delete <service-id>
 clickhousectl cloud service delete <service-id> --force
 ```
 
+`--backup-start-time` requires the backup period to be 24 or 48 hours. Nothing is defaulted when the period is omitted: the API validates the new start time against the period already stored on the service, so either pass `--backup-period-hours 24` or `--backup-period-hours 48` in the same call, or leave the stored period at one of those. When a start time is given without a period, the CLI reads the current configuration first and fails before sending the update if the stored period is something else.
+
 `--force` stops the service and then polls it until the stop completes, which takes minutes on a real service, printing each state change to stderr (stdout stays reserved for the result). Progress output is best-effort: if whatever was reading it goes away — a pager you quit, a supervising process that stopped reading — the lines are dropped and the deletion still runs to completion and exits 0.
 
 Use `clickhousectl cloud service create --help` for the complete option list. If omitted, `--provider` defaults to `aws`, `--region` defaults to `us-east-1`, and the IP allowlist defaults to `0.0.0.0/0`; production workflows should normally set all three explicitly. When the create response includes an initial password, it is shown only once.
--- a/crates/clickhousectl/src/cloud/backups.rs
+++ b/crates/clickhousectl/src/cloud/backups.rs
@@ -78,8 +78,9 @@ pub enum BackupConfigCommands {
         #[arg(long)]
         backup_retention_period_hours: Option<u32>,
 
-        /// Backup start time in UTC, exactly on the hour (HH:00). If the period
-        /// is omitted, the API sets it to 24 hours.
+        /// Backup start time in UTC, exactly on the hour (HH:00). Requires the
+        /// backup period to be 24 or 48 hours: pass --backup-period-hours 24|48
+        /// in the same call, or the stored period must already be one of those.
         #[arg(long, value_parser = parse_backup_start_time)]
         backup_start_time: Option<String>,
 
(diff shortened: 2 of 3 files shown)


iskakaushik (approved):
LGTM

PR #646, opened 2026-09-01
Add --clear-backup-start-time so the backup start time is no longer a one-way door

### LINKED ISSUE #564: Backup start time cannot be cleared once set, making the 24/48-hour period constraint permanent
## Summary

Once a custom backup start time has been set via `cloud service backup-config update --backup-start-time`, there is no way to clear it again through the CLI:

- There is no unset/clear flag.
- `--backup-start-time ""` is rejected by flag validation (correctly, since it is not a valid time).
- `BackupConfigurationPatchRequest.backup_start_time` uses `skip_serializing_if = "Option::is_none"`, so an unset option omits `backupStartTime` from the PATCH body entirely — the CLI can never send an explicit `"backupStartTime": null`.
- The OpenAPI spec does not declare `backupStartTime` as nullable in `BackupConfigurationPatchRequest`, so clearing may also need upstream API support.

The consequence is worse than a cosmetic gap: while a start time is set, the API constrains the backup period to 24 or 48 hours (`BAD_REQUEST: customBackupPeriod must be 24 or 48 hours when customBackupStartTime is set`). Since the start time can never be removed, a service that ever sets one can never return to any other backup period (e.g. 12 hours). Setting `--backup-start-time` is a one-way door.

## Reproduction

```console
# Set a start time (period becomes 24h)
clickhousectl cloud service backup-config update <service-id> --backup-start-time 02:00

# Attempt 1: clear it — rejected by flag validation
clickhousectl cloud service backup-config update <service-id> --backup-start-time ""

# Attempt 2: there is no --clear/--unset flag, and omitting the flag omits the field from the PATCH body (no-op)

# The constraint is now permanent:
clickhousectl cloud service backup-config update <service-id> --backup-period-hours 12
# Error: BAD_REQUEST: customBackupPeriod must be 24 or 48 hours when customBackupStartTime is set
```

## Expected behavior

One of:

- A way to clear the start time, e.g. a `--clear-backup-start-time` flag that sends `"backupStartTime": null` — if the Cloud API accepts an explicit `null` despite the spec not marking the field nullable.
- If the API genuinely cannot clear it, document the permanence in `--help` for `--backup-start-time` (alongside the existing `HH:00` and 24/48-hour notes from #438), so users know it is a one-way setting before using it.

## Environment

Observed with `clickhousectl 0.4.1` against a live ClickHouse Cloud service while validating the CLI flow for the configurable-backups docs page (https://github.com/ClickHouse/ClickHouse/pull/116775). Confirmed still present on `main` (`9de8cd8`): the flag parser accepts only `HH:00`, and the patch model omits the field when unset.

Related: https://github.com/ClickHouse/clickhousectl/issues/425
Related: https://github.com/ClickHouse/clickhousectl/pull/438

sdairs:
## What

`cloud service backup-config update` gains `--clear-backup-start-time`, which removes a stored backup start time by sending an explicit `"backupStartTime": null`. It conflicts with `--backup-start-time` and can be combined with `--backup-period-hours`, so a service can be freed from the 24/48 hour restriction and given a new period in a single call.

The API library makes that expressible: `BackupConfigurationPatchRequest.backup_start_time` becomes `Option<Option<String>>`, so `None` omits the key, `Some(None)` serializes an explicit `null`, and `Some(Some(time))` sends the time string. The field keeps `skip_serializing_if`, stays free of `serde(default)`, and needed no analyzer exemption. For library consumers this is a type change on a request field: code that set `backup_start_time: Some(time)` must now write `Some(Some(time))`.

The field was `Option<String>` with `skip_serializing_if`, which has only two states: send a time, or say nothing. A PATCH that can only set a value cannot undo it, so no CLI flag could have cleared the start time without the library change first.

## Behaviour after the fix

Verified live against api.clickhouse.cloud on a disposable service created and deleted for this PR:

```
PATCH .../backupConfiguration  {"backupStartTime": null}
200 {"result":{"backupPeriodInHours":24,"backupRetentionPeriodInHours":24}}
GET  .../backupConfiguration
200 {"result":{"backupPeriodInHours":24,"backupRetentionPeriodInHours":24}}     # start time gone

PATCH .../backupConfiguration  {"backupStartTime": ""}
400 {"error":"BAD_REQUEST: customBackupStartTime must be a valid time (HH:00)"}  # so null it is

PATCH .../backupConfiguration  {"backupPeriodInHours": 12}                       # with 03:00 stored
400 {"error":"BAD_REQUEST: customBackupPeriod must be 24 or 48 hours when customBackupStartTime is set"}

PATCH .../backupConfiguration  {"backupStartTime": null, "backupPeriodInHours": 12}
200 {"result":{"backupPeriodInHours":12,"backupRetentionPeriodInHours":24}}      # clear and reperiod together
```

Repeating the null on a configuration that has no start time returns 200 and changes nothing, so clearing is idempotent. The same three steps then ran through the built binary end to end: setting `02:00`, being refused a 12 hour period, and recovering with `--clear-backup-start-time --backup-period-hours 12`.

Note that the OpenAPI spec does not declare `backupStartTime` nullable. The runtime accepts the null anyway, which is what the rustdoc on the field and the README paragraph both record.

## Tests

- Library serialization unit tests in `crates/clickhouse-cloud-api/tests/models_test.rs`: untouched start time omits the key, clear emits `null`, set emits the string.
- Clap tests next to the command: `--clear-backup-start-time` parses with a period, and the conflict with `--backup-start-time` is rejected. The help test now pins both the clear flag's description and the cross-reference from `--backup-start-time`.
- Request builder tests: clear produces `Some(None)` and serializes to `{"backupStartTime": null}`, clear travels with an incompatible period, and setting plus clearing together is an error.
- Subprocess and wiremock tests in `crates/clickhousectl/tests/cli_request_shape_test.rs` asserting the exact PATCH body for the clear, the clear plus period 12 combination including the `--json` output, and that conflicting flags exit 2 without any request.
- `cargo fmt --all`
- `cargo test -p clickhousectl` (23 test binaries, all passing)
- `cargo test -p clickhouse-cloud-api --lib --test spec_coverage_test`
- `cargo test -p clickhouse-cloud-api --test models_test --test client_test`
- `cargo test -p clickhouse-openapi-analyzer`

## Stacking

Part of a stacked PR chain: based on `fix/642-backup-start-time-help`, not `main`.

Fixes #564

### DIFF
--- a/README.md
+++ b/README.md
@@ -627,6 +627,11 @@ clickhousectl cloud service backup-config update <service-id> \
   --backup-retention-period-hours 720 \
   --backup-start-time 02:00
 
+# Remove the start time again, optionally changing the period in the same call
+clickhousectl cloud service backup-config update <service-id> \
+  --clear-backup-start-time \
+  --backup-period-hours 12
+
 # Service Prometheus configuration
 clickhousectl cloud service prometheus <service-id> --filtered-metrics true
 
@@ -638,6 +643,8 @@ clickhousectl cloud service delete <service-id> --force
 ```
 
 `--backup-start-time` requires the backup period to be 24 or 48 hours. Nothing is defaulted when the period is omitted: the API validates the new start time against the period already stored on the service, so either pass `--backup-period-hours 24` or `--backup-period-hours 48` in the same call, or leave the stored period at one of those. When a start time is given without a period, the CLI reads the current configuration first and fails before sending the update if the stored period is something else.
+
+`--clear-backup-start-time` removes a stored start time and lifts that restriction, so a service that has one can go back to any backup period. It sends an explicit `"backupStartTime": null`, which the API accepts even though the OpenAPI spec does not mark the field nullable, and it can be combined with `--backup-period-hours` to clear the start time and set an otherwise incompatible period in a single call. The two start-time flags conflict: pass one or the other.
 
 `--force` stops the service and then polls it until the stop completes, which takes minutes on a real service, printing each state change to stderr (stdout stays reserved for the result). Progress output is best-effort: if whatever was reading it goes away — a pager you quit, a supervising process that stopped reading — the lines are dropped and the deletion still runs to completion and exits 0.
 
--- a/crates/clickhouse-cloud-api/src/models/backups.rs
+++ b/crates/clickhouse-cloud-api/src/models/backups.rs
@@ -574,8 +574,18 @@ pub struct BackupConfigurationPatchRequest {
         skip_serializing_if = "Option::is_none"
     )]
     pub backup_retention_period_in_hours: Option<f64>,
+    /// Three states, because a PATCH that can only set a value cannot undo it:
+    /// `None` omits the key and leaves the stored start time alone,
+    /// `Some(None)` sends an explicit `null`, and `Some(Some(time))` sends the
+    /// time string. An explicit `null` is what clears a stored start time
+    /// (verified against api.clickhouse.cloud on 2026-09-01: the PATCH returns
+    /// 200 and the following GET no longer carries the field), which matters
+    /// because the API refuses any backup period other than 24 or 48 hours
+    /// while one is stored. The empty string is not an alternative: the API
+    /// answers it with `BAD_REQUEST: customBackupStartTime must be a valid
+    /// time (HH:00)`.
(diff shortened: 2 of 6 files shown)


iskakaushik (approved):
LGTM

PR #647, opened 2026-09-01
Refuse `cloud service query --query` when data is waiting on stdin

### LINKED ISSUE #641: `cloud service query --query` silently ignores piped stdin, so `INSERT ... FORMAT CSV < data.csv` reports OK and inserts nothing
## Summary

When `--query` is given, `cloud service query` never looks at stdin. Anything piped in is discarded with no warning. The most common way to hit this is the natural shape for a CSV load:

```console
$ clickhousectl cloud service query --id "$CH_ID" \
    --query "INSERT INTO trips FORMAT CSV" < data.csv
OK
$ echo $?
0
$ clickhousectl cloud service query --id "$CH_ID" --query "SELECT count() FROM trips"
0
```

The statement is sent with no data, the server accepts the empty insert, the CLI prints `OK` and exits 0. Zero rows land. Nothing tells the user that the file they redirected in was never read.

Found while validating the "Upload a CSV file" docs page (ClickHouse/ClickHouse#116778) on 0.4.1 and confirmed unchanged on `main` at `9aceafb` during the 0.4.2 re-validation. The docs page currently carries a warning and a workaround (`printf 'INSERT ... FORMAT CSV\n' | cat - data.csv | clickhousectl cloud service query --id "$CH_ID"`) because the CLI gives no signal.

## Where

`read_query_sql` in `crates/clickhousectl/src/cloud/services.rs` returns as soon as `inline` is `Some`, before stdin is ever inspected:

```rust
if let Some(query) = inline {
    if query.trim().is_empty() {
        return Err(CloudError::new("--query was empty"));
    }
    return Ok(query.to_string());   // stdin never consulted
}
```

The TTY check (`std::io::stdin().is_terminal()`) only runs on the neither-flag path.

#455 made `--query` and `--queries-file` mutually exclusive but did not touch stdin. `--queries-file -` reads stdin correctly and is a working alternative today.

## Expected behavior

`--query` combined with non-terminal stdin should not succeed silently. Preferred, in order:

1. **Error** (exit 2 or 1, before any network call): `--query cannot be combined with SQL or data on stdin. Pipe the statement and its data together, or use --queries-file -.` This is the agent-safe choice: an agent that redirects a file into `--query` gets a hard stop instead of a green result over an empty table.
2. Failing that, a stderr **warning** that stdin was ignored, so at least a human sees it.

Whichever is chosen, the help text should state that `--query` does not read stdin and that piped data must accompany the statement on stdin.

Note the constraint: the Query API takes a single request body, so there is no separate data channel. `--query "INSERT ... FORMAT CSV"` plus a data file can never work over this transport. The error message should say what does work rather than only what does not.

## Acceptance criteria

- `--query` with stdin that is not a terminal fails (or warns) before dispatch; `--query` with a terminal stdin is unchanged.
- `--queries-file -` and bare piped stdin keep working.
- Help for `--query` documents the stdin rule.
- Subprocess coverage in `tests/cli_request_shape_test.rs` for the redirected-stdin case, asserting no request is sent.

## Environment

clickhousectl 0.4.1 and `main` @ `9aceafb`, macOS. Transcript in the docs validation run for ClickHouse/ClickHouse#116778.

sdairs:
## What

`cloud service query --query "INSERT INTO trips FORMAT CSV" < data.csv` now fails before any network call instead of silently discarding the redirected file, sending the bare statement, printing `OK` and exiting 0 with zero rows inserted.

`read_query_sql` classifies stdin on the `--query` path and returns a `CloudError` (exit code 1, the same structured error path every other cloud runtime failure uses, and the same rendering with `--json`):

```
Error: --query cannot be combined with SQL or data on stdin. The Query API sends one
request body, so redirected data is never read. Pipe the statement and its data
together on stdin instead: printf 'INSERT INTO t FORMAT CSV\n' | cat - data.csv |
clickhousectl cloud service query --id <id>. Or read a whole statement from stdin
with --queries-file -.
```

The refusal happens while reading the SQL, before organization and service resolution, so neither the control plane nor the query host is contacted. It inherits the existing `sql_input` telemetry stage through the `at_stage` call that already wraps `read_query_sql`, so no new telemetry vocabulary was added and nothing is derived from message text.

## Why

The Query API takes a single request body, so `--query` plus a separate data channel can never work over this transport. Before this change the CLI gave no signal at all: the docs page for uploading a CSV had to carry a warning and a workaround because a natural looking command reported success over an empty table. An agent redirecting a file into `--query` now gets a hard stop instead of a green result.

## Behaviour after the fix

- `--query` with a terminal stdin: unchanged, and stdin is never read, so it can never block on the user.
- `--query` with a non-terminal but empty stdin (`</dev/null`, a CI runner, a coding agent with no tty): unchanged, the query runs.
- `--query` with data on stdin, whether already waiting or produced a moment later: exit 1 with the message above and zero requests.
- Bare piped stdin and `--queries-file -`: unchanged.

Readiness comes from `poll` on stdin with a bounded 250 ms timeout, then `fill_buf` only if poll says a read is ready, so no byte is consumed and the other SQL sources are untouched. The timeout is what makes the check honest rather than cosmetic. A zero timeout would leave the reported bug in place as a race: in `cat data.csv | clickhousectl ... --query ...` the shell starts both processes at once, so the CLI can reach the check before `cat` has written its first byte and would go on to send the empty INSERT. Nothing that already knows its answer pays the wait: a regular file, `/dev/null`, a closed pipe (`POLLHUP`) and a pipe that already holds data all return immediately. The only invocation that waits is `--query` whose stdin is an open pipe with nothing written yet, which is the hang prone harness case, and it waits 250 ms rather than forever. The residual is narrow and documented: a producer that has written nothing within those 250 ms still reads as no input.

Help text for `--query` and the `CONTEXT FOR AGENTS` block now state that `--query` does not read stdin and show the pipe together form.

## Tests

- Unit tests next to the helpers in `crates/clickhousectl/src/cloud/services.rs`: the full matrix of stdin states (terminal, non-terminal empty, non-terminal with data) crossed with `--query` present and absent, the message content, and a help text assertion.
- Pipe level unit tests that pin the timeout deterministically, against a real `pipe()` rather than a subprocess so the interleaving is not left to process startup: a producer that writes its first byte 100 ms late is caught (this test fails if the timeout is set back to 0), a pipe nobody writes to gives up within the bound instead of hanging, and data already waiting is answered immediately.
- Subprocess plus wiremock tests in `crates/clickhousectl/tests/cli_request_shape_test.rs`, driven by a `StdinPlan` describing how the test feeds the child: a redirected file with content is refused with exit 1 and zero received requests on both mocks; data written 150 ms after spawn is refused the same way; a pipe held open and never written to still sends the query (the no hang guard); `</dev/null` still sends the query; bare piped stdin SQL and `--queries-file -` still work.
- The existing `service query` subprocess tests now pin `stdin` to `Stdio::null()` so they never inherit the test runner's stdin.
- `cargo fmt --all`
- `cargo test -p clickhousectl` (all 23 test binaries green, 1189 tests)
- Manual checks against the built binary: `( sleep 0.1; cat data.csv ) | clickhousectl ... --query ...` is refused (it was silently accepted with a zero timeout), a redirected file and immediate piped data are refused, `</dev/null` adds no measurable delay, and `sleep 30 | clickhousectl ... --query ...` proceeds after 0.26 s instead of hanging.

## Stacking

Part of a stacked PR chain: based on `feat/564-backup-start-time-clear`, not `main`.

Fixes #641

### DIFF
--- a/README.md
+++ b/README.md
@@ -561,6 +561,10 @@ clickhousectl cloud service query --name my-service --queries-file query.sql   #
 clickhousectl cloud service query --name my-service --database mydb --query "SHOW TABLES"
 echo "SELECT 1+1" | clickhousectl cloud service query --name my-service
 
+# Load a CSV: the statement and its data travel together on stdin
+printf 'INSERT INTO trips FORMAT CSV\n' | cat - data.csv | \
+  clickhousectl cloud service query --id <service-id>
+
 # Replace a stale clickhousectl-owned Query API key for exactly one service
 clickhousectl cloud service repair-query-key <service-id> --org-id <org-id>
 
@@ -651,6 +655,15 @@ clickhousectl cloud service delete <service-id> --force
 Use `clickhousectl cloud service create --help` for the complete option list. If omitted, `--provider` defaults to `aws`, `--region` defaults to `us-east-1`, and the IP allowlist defaults to `0.0.0.0/0`; production workflows should normally set all three explicitly. When the create response includes an initial password, it is shown only once.
 
 `--query` and `--queries-file` are mutually exclusive. If neither is supplied, `cloud service query` reads SQL from stdin; `--queries-file -` also reads stdin explicitly.
+
+`--query` never reads stdin, so `--query "INSERT INTO trips FORMAT CSV" < data.csv` is refused (exit code `1`) before any request is sent rather than silently inserting nothing. The Query API takes a single request body, so a statement and a separate data stream cannot both be sent; pipe them together instead:
+
+```bash
+printf 'INSERT INTO trips FORMAT CSV\n' | cat - data.csv | \
+  clickhousectl cloud service query --id <service-id>
+```
+
+Only real input counts as a conflict. A redirected file, a closed pipe, `/dev/null` and a pipe that already holds data are all answered immediately, and a pipe that is open but silent is given 250 ms to produce its first byte before the CLI treats stdin as empty and runs the query. So an empty non-terminal stdin, which is what CI runners and coding agents normally have, leaves `--query` working as before, a pipe nobody writes to can never hang the command, and a producer that is merely slow to start still gets caught. The residual is narrow and deliberate: a producer that has written nothing within those 250 ms is indistinguishable from no input at all.
 
 Whatever the source, the SQL must be a single statement. The Query API runs exactly one statement per request, so a multi-statement `.sql` script is rejected by ClickHouse (error 62, `Multi-statements are not allowed`). Run statements one invocation at a time, or put a real client on PATH with `clickhousectl local use latest` and run the script through `clickhouse client` connected to the service.
 
--- a/crates/clickhousectl/src/cloud/services.rs
+++ b/crates/clickhousectl/src/cloud/services.rs
@@ -445,9 +445,14 @@ CONTEXT FOR AGENTS:
   puts the standard `clickhouse` binary on PATH; use `clickhouse client` to connect
   to the service instead.
   SQL sources: --query and --queries-file are mutually exclusive. When neither
-  is supplied, SQL is read from stdin. Default format: PrettyCompact on a TTY,
-  TabSeparated when piped. --json selects JSONEachRow and cannot be combined
-  with --format; an explicit --format takes precedence over agent auto-JSON.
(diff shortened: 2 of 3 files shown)


iskakaushik (approved):
LGTM

PR #649, opened 2026-09-02
Name the 30-second Query API limit and hint the native client on a gateway timeout

### LINKED ISSUE #644: Detect the Query API 30-second gateway timeout on `service query`: name the limit in help and hint the native-client command on failure
## Summary

The Query API gateway aborts requests after roughly 30 seconds. The statement keeps running on the server; only the HTTP response is lost. Today the CLI reports it as an anonymous server error:

```console
$ clickhousectl cloud service query --id "$CH_ID" --query "INSERT INTO pp_complete SELECT ... FROM url(...)"
Error: Query API returned HTTP 500 Internal Server Error: {"error":"Timeout error."}
exit=1   (after ~31 s)
```

Observed while loading the UK Price Paid dataset for ClickHouse/ClickHouse#116777. The INSERT finished server-side minutes later (`system.processes` showed it still running; the row count kept climbing after the CLI had exited).

Two problems:

1. The help text added by #428 says "For queries that may exceed Query API timeouts" without saying what the timeout is. A user cannot know whether their query is at risk.
2. The failure gives no guidance. A human, and more so an agent, will read "500 Internal Server Error" as transient and retry, which for an INSERT means the statement runs twice.

### Help text

Name the limit and the route around it in the `CONTEXT FOR AGENTS` block on `service query`:

> The Query API times out after about 30 seconds; the statement keeps running on the service but the result is lost. For anything longer (large INSERTs, backfills, `url()` loads), run it over the native protocol: `clickhousectl local use latest` puts `clickhouse` on PATH, then `clickhouse client --host <host> --secure --port 9440 --user default --password <password>`.

### Structured detection

Add a dedicated library variant, `Error::QueryTimeout`, alongside `Error::ServiceStopped` and `Error::ServiceIdle` in `crates/clickhouse-cloud-api/src/error.rs`, produced by the query path when the gateway returns HTTP 500 with the body `{"error":"Timeout error."}`. That follows the crate rule that a failure a caller must tell apart gets its own variant rather than a recognizable message, and it is the same shape as the existing `query_api_reports_stopped_service` detection in `client.rs`. The telemetry classification in #450 then gets it for free as a `FailureKind` from the variant, not from text.

### CLI hint

`convert_query_error` in `crates/clickhousectl/src/cloud/services.rs` already has the resolved service (from `resolve_service`) and the org, so it can render a concrete command. The service's `nativesecure` endpoint gives the host and port. Proposed output:

```
Error: the query timed out at the Query API gateway, which stops waiting after about 30 seconds.
The statement may still be running on the service; check `SELECT query_id, elapsed FROM system.processes` before running it again.

Hint: run long statements over the native protocol instead. Install the client with
  clickhousectl local use latest
then rerun the query with
  clickhouse client --host <nativesecure host> --secure --port 9440 --user default --password '<password>' --query '<your SQL>'
The password is the one shown when the service was created; `clickhousectl cloud service reset-password <service-id>` issues a new one.
```

The `system.processes` line matters: without it the hint would tell the user to rerun a statement that is still executing. For an INSERT that is a duplicate load. The hint should make clear that it is offering a different transport for the *next* attempt, not a retry.

Exit code stays 1. The CLI must not retry on this error itself; the retry/backoff in `#537` is for endpoint readiness only and should not be extended to it.

### `--json`

Emit the structured error with a stable code (for example `query_timeout`), the host, port and the suggested command as fields, matching how local structured errors carry hints (#475 / #608), so an agent can act on it without parsing prose.

## Acceptance criteria

- Library: `Error::QueryTimeout` with `Display` fixed; unit test on the 500 + `{"error":"Timeout error."}` body; other 500 bodies still map to `Error::Api`.
- CLI: help names the ~30 s limit; the timeout error renders the hint with the service's real `nativesecure` host, or falls back to `<host>` plus `service get` guidance when endpoints are absent from the response.
- Subprocess test in `tests/cli_request_shape_test.rs` with a wiremock 500 timeout body, asserting the hint text, the host substitution, exit 1, and that exactly one query request was made.
- README `service query` section mentions the limit and the native-client route.

Related: #428 (added the current help sentence), #450 (failure classification), #475 / #608 (structured JSON errors with hints).

sdairs:
## What

`cloud service query` now recognises the Query API gateway timeout as its own failure and tells the user what to do about it.

Library (`clickhouse-cloud-api`): new `Error::QueryTimeout`, produced by the query path when the gateway answers HTTP 500 with a JSON body whose `error` field is exactly `Timeout error.`. The body is parsed as JSON and the field compared in full, the same shape as the existing stopped-service detection, so a ClickHouse error that merely mentions a timeout cannot be mistaken for the gateway giving up. Every other 500 body still maps to `Error::Api` with an unchanged message.

Telemetry: the new variant maps to the existing `timeout` failure kind, from the variant and never from message text, carrying the HTTP 500 it is only ever produced from.

CLI: the `CONTEXT FOR AGENTS` block on `service query` names the roughly 30 second limit, says the statement keeps running on the service, and gives the native-protocol route. On `Error::QueryTimeout` the error itself says the gateway stops waiting after about 30 seconds, points at `SELECT query_id, elapsed FROM system.processes` before any rerun, then hints `clickhousectl local use latest` plus a `clickhouse client` command built from the service's own `nativesecure` endpoint, taken from the `GET service` response the query already made. When the response carries no complete native endpoint, the command keeps a `<host>` placeholder and the message says to read the host and port from `clickhousectl cloud service get <id> --org-id <org>`. Nothing is guessed, the user's SQL and the password are always literal placeholders, exit code stays 1, and there is no retry anywhere on this path.

`--json` gets the same failure machine-readably. `CloudError` can now carry a `CloudErrorDetail`, which reaches the top level as `Error::CloudDetailed` and is written as one object on stderr in the envelope local structured errors already use: a stable `query_timeout` code plus `host`, `port` and `command`. Absent fields are omitted rather than nulled, and a cloud failure with no structured remedy stays prose, so no code is ever invented from a message.

## Why

The gateway aborts after roughly 30 seconds and the statement keeps running on the service; only the HTTP response is lost. Reported as `Query API returned HTTP 500 Internal Server Error: {"error":"Timeout error."}`, that reads as transient, so a human and much more readily an agent retries it, and for an `INSERT` that means loading the data twice. This was hit loading the UK Price Paid dataset: the INSERT finished minutes after the CLI had exited. The old help text ("for queries that may exceed Query API timeouts") also gave no number, so a user could not tell whether their query was at risk.

A failure a caller must tell apart gets its own variant rather than a recognisable message, which is the crate's rule for `Error::Sql` and `Error::ServiceStopped`. Making it a variant is also what lets the failure classification come from the type instead of from text, and what lets the CLI attach a hint to exactly this failure without pattern matching on a 500 body.

## Behaviour after the fix

```console
$ clickhousectl cloud service query --id "$CH_ID" --query "INSERT INTO pp_complete SELECT ... FROM url(...)"
Error: the query timed out at the Query API gateway, which stops waiting after about 30 seconds.
The statement may still be running on the service; check `SELECT query_id, elapsed FROM system.processes` before running it again.

Hint: run long statements over the native protocol instead. Install the client with
  clickhousectl local use latest
then rerun the query with
  clickhouse client --host demo.gcp.clickhouse.cloud --secure --port 9440 --user default --password '<password>' --query '<your SQL>'
The password is the one shown when the service was created; `clickhousectl cloud service reset-password <service-id>` issues a new one.
exit=1
```

Under `--json`, or when a coding agent is detected:

```json
{
  "error": {
    "code": "query_timeout",
    "message": "the query timed out at the Query API gateway, ...",
    "host": "demo.gcp.clickhouse.cloud",
    "port": 9440,
    "command": "clickhouse client --host demo.gcp.clickhouse.cloud --secure --port 9440 --user default --password '<password>' --query '<your SQL>'"
  }
}
```

## Tests

- Library unit tests on the classifier: the exact gateway body becomes `QueryTimeout` with its `Display` pinned; a different message, a body that merely mentions a timeout, the right message in the wrong shape, non-JSON and an empty body all stay `Error::Api`; the same body under 206, 400, 502 and 504 stays `Error::Api`.
- Library wiremock tests in `tests/run_query_test.rs`: a 500 timeout body maps to `QueryTimeout` and exactly one request is sent; a different 500 body keeps its `Error::Api` message verbatim.
- Telemetry mapping table gains `QueryTimeout` to `timeout` with status 500.
- CLI unit tests: `nativesecure` endpoint selection, including that a half-returned endpoint (no host, no port, or no protocol) counts as absent; the timeout message and its JSON detail with a real endpoint; the `<host>` placeholder and `service get` guidance without one, with `host` and `port` omitted from the JSON; and that neither the SQL nor a credential can reach the message.
- Pinned help test updated to assert the named limit, the "statement keeps running" wording and the native-client command.
- Subprocess and wiremock tests in `tests/cli_request_shape_test.rs`: a query host returning the 500 timeout body gives exit 1, the 30 second sentence, the `system.processes` sentence and the substituted `nativesecure` host from the mocked service GET, with exactly one query request sent; a service response without endpoints asserts the `<host>` fallback and the `service get` guidance; two `--json` variants assert the code, host, port and command, and that an absent endpoint is omitted rather than nulled; a control case with a different 500 body keeps the generic error in both output modes; and `service query --help` names the limit.
- Verification commands, all passing: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo clippy -p clickhousectl --all-targets --no-default-features -- -D warnings`, `cargo test -p clickhousectl` (887 unit plus 172 request-shape plus the remaining integration binaries, 0 failures), `cargo test -p clickhouse-cloud-api --lib --test spec_coverage_test`, `cargo test -p clickhouse-cloud-api --test run_query_test`, `cargo check --workspace --all-features`, `python3 -m unittest discover -s scripts/tests -p 'test_*.py'`.
- README `service query` section documents the limit, the native-client route and the `query_timeout` JSON error shape.

## Stacking

Part of a stacked PR chain: based on `fix/643-clickpipe-settings-get-db-pipes`, not `main`.

Fixes #644

### DIFF
--- a/README.md
+++ b/README.md
@@ -665,6 +665,34 @@ printf 'INSERT INTO trips FORMAT CSV\n' | cat - data.csv | \
 
 Only real input counts as a conflict. A redirected file, a closed pipe, `/dev/null` and a pipe that already holds data are all answered immediately, and a pipe that is open but silent is given 250 ms to produce its first byte before the CLI treats stdin as empty and runs the query. So an empty non-terminal stdin, which is what CI runners and coding agents normally have, leaves `--query` working as before, a pipe nobody writes to can never hang the command, and a producer that is merely slow to start still gets caught. The residual is narrow and deliberate: a producer that has written nothing within those 250 ms is indistinguishable from no input at all.
 
+The Query API gateway stops waiting after about 30 seconds. When it does, the request fails (exit code `1`) but **the statement keeps running on the service** — only the HTTP response is lost. The error says so, points at `SELECT query_id, elapsed FROM system.processes` so a still-running statement is not started a second time, and prints the `clickhouse client` command for the service's own native endpoint. Nothing is retried automatically: re-sending a large `INSERT` would load the data twice.
+
+For anything that may run longer than that — large `INSERT`s, backfills, `url()` loads — use the native protocol from the start:
+
+```bash
+clickhousectl local use latest   # puts the standard `clickhouse` binary on PATH
+clickhouse client --host <nativesecure-host> --secure --port 9440 \
+  --user default --password '<password>' --query '<your SQL>'
+```
+
+`clickhousectl cloud service get <service-id>` lists the service endpoints, including the `nativesecure` host and port. The password is the one shown when the service was created; `clickhousectl cloud service reset-password <service-id>` issues a new one.
+
+Under `--json` (or when a coding agent is detected) that failure is emitted as one JSON object on stderr, in the same envelope local errors use:
+
+```json
+{
+  "error": {
+    "code": "query_timeout",
+    "message": "the query timed out at the Query API gateway, ...",
+    "host": "my-service.us-east-1.aws.clickhouse.cloud",
+    "port": 9440,
+    "command": "clickhouse client --host my-service.us-east-1.aws.clickhouse.cloud --secure --port 9440 --user default --password '<password>' --query '<your SQL>'"
+  }
+}
+```
+
+`code` is a stable machine-readable identifier. `host` and `port` are omitted when the API response carried no native endpoint, in which case `command` names the `<host>` placeholder instead. The suggested command never contains your SQL or your password: both are placeholders. Cloud failures that have no structured remedy stay prose on stderr (`Error: <message>`).
+
 Whatever the source, the SQL must be a single statement. The Query API runs exactly one statement per request, so a multi-statement `.sql` script is rejected by ClickHouse (error 62, `Multi-statements are not allowed`). Run statements one invocation at a time, or put a real client on PATH with `clickhousectl local use latest` and run the script through `clickhouse client` connected to the service.
 
 Private endpoint IDs supplied to `private-endpoint create --endpoint-id` and `service update --add-private-endpoint-id` are format-checked before the request is sent, because adding one registers it for the whole organization and a typo has to be unpicked from both the service and the organization. Each provider uses its own format — AWS a `vpce-` VPC endpoint ID, GCP the numeric Private Service Connect connection ID, Azure the private endpoint Resource ID or `resourceGuid` — and the provider is not known when the flag is parsed, so only provider-independent mistakes are rejected (exit code `2`): empty values, values containing whitespace, and any value carrying `vpce-` that is not exactly a well-formed AWS VPC endpoint ID (`vpce-` plus 8 or 17 lowercase hex characters) — which also catches a pasted VPC endpoint ARN or endpoint service name. Azure Resource IDs (values starting with `/`) are exempt from that check, since an Azure resource may itself be named `vpce-...`. Whether the endpoint actually exists and belongs to you is not validated by the CLI or, currently, by the Cloud API. Removal flags are never format-checked, so an already-registered bogus ID stays removable.
--- a/crates/clickhouse-cloud-api/src/client.rs
+++ b/crates/clickhouse-cloud-api/src/client.rs
@@ -88,6 +88,9 @@ fn query_api_sql_error(body: &str) -> Option<(String, String)> {
 /// only print the error see no change; callers that need to know *what kind of
 /// failure this was* read the variant instead of the message.
 fn query_api_error(status: reqwest::StatusCode, body: &str) -> Error {
+    if query_api_reports_timeout(status, body) {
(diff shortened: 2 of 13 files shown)


iskakaushik (approved):
LGTM

PR #650, opened 2026-09-02
Expose destination roles on `clickpipe create` (`--role`)

### LINKED ISSUE #568: Expose destination `roles` on `clickpipe create` (`--role`)
## Problem

`ClickPipeMutateDestination.roles` lets a pipe grant roles to the ClickHouse user that ClickPipes creates ("If omitted, the user is granted the default role"). The CLI hard-codes `roles: None` in `build_destination` (`crates/clickhousectl/src/cloud/clickpipes.rs`), so custom roles cannot be granted from the CLI on any pipe type.

This is the closest API surface to the UI's permission-role selection at pipe creation.

## Library support

Already complete: `roles: Option<Vec<String>>` on `ClickPipeMutateDestination` in `crates/clickhouse-cloud-api/src/models/clickpipes.rs`. CLI-only work.

## Proposed change

Add a repeatable `--role` flag to the `clickpipe create` subcommands and wire it through `build_destination` (including the database-pipe branch, which currently omits the field via `Default`). Omitted → field absent from the request body (today's behavior). The API reserves `clickpipes` and `clickpipes_system`; consider rejecting those client-side with a clear message rather than relaying the API error.

## Open question for product/API

The UI offers "Full access" vs "Only destination table". The API field only *adds* roles on top of the default; it is unclear whether the restricted mode is expressible through it. If not, that part is an API gap to raise upstream — this issue covers exposing what the API already supports.

## Acceptance criteria

- `--role` repeatable on all create subcommands, mapped to `destination.roles`; omitted flag leaves the field out of the JSON body.
- Clap parsing and request-shape tests.

## Context

Found while adding CLI steps to ClickHouse Cloud docs — the UI's permission-role selection had no CLI equivalent.

sdairs:
## What

Adds a repeatable `--role <ROLE>` flag to every `clickpipe create` subcommand (object-storage, kafka, kinesis, postgres, mysql, mongodb, bigquery) and wires it to `destination.roles` in the create request.

The flag lives in one place: a new flattened `DestinationRoleArgs` struct, the same pattern `KafkaSourceFields` already uses for the shared Kafka source flags, so the definition and the help text are not copied seven times.

Behaviour:

- Omitted: `roles` stays out of the JSON body, which is how ClickPipes reads "grant the destination user the default role". This is today's behaviour, unchanged.
- Given: `roles` carries the names in declaration order, trimmed, with duplicates dropped rather than sent twice.
- A blank or whitespace-only name, and the API-reserved names `clickpipes` and `clickpipes_system`, are rejected by a clap value parser. That makes them usage errors (exit code 2) formatted against the create subcommand that was invoked, raised before any credential lookup, file read, or network call. The reserved check is case-insensitive.

`build_destination` now takes the resolved roles and sets the field on both of its branches, including the database-pipe branch that previously emitted only `database` via `Default`. `roles` is not one of the four fields (`table`, `columns`, `managedTable`, `tableDefinition`) that database pipes reject, so Postgres, MySQL, MongoDB and BigQuery pipes can grant roles too.

## Why

`ClickPipeMutateDestination.roles` is the closest API surface to the UI's permission-role selection at pipe creation, and the CLI hard-coded `roles: None`, so custom roles could not be granted from the CLI on any pipe type. The API library already models the field, so this is CLI-only work.

The reserved names are rejected client-side rather than relayed from the server so the diagnostic names the flag and the two forbidden values, and so no request is made for an invocation that cannot succeed.

The issue's open question about the UI's "Full access" versus "Only destination table" modes is out of scope here: the API field only adds roles on top of the default, and this PR exposes what the API already supports.

## Behaviour after the fix

```
$ clickhousectl cloud clickpipe create postgres svc --name p --host h --pg-database d \
    --username u --password p --table-mapping public.t:t --role clickpipes_system
error: invalid value 'clickpipes_system' for '--role <ROLE>': role names reserved by ClickPipes cannot be granted: clickpipes, clickpipes_system

For more information, try '--help'.
```

## Tests

- Clap `try_parse_from` coverage next to the command definitions: repeatability and the empty default on `create kafka` and `create postgres`, plus a loop asserting `--role` parses on all seven create subcommands and that a reserved name is a `ValueValidation` error with exit code 2 on each.
- Unit tests for the validator (trimming, blank names, reserved names including a mixed-case and a padded form) and for the role resolver (absent when no flags, de-duplication preserving order).
- `build_destination` tests asserting the library struct field on both destination shapes, and the maximal `build_postgres_request` test extended with two roles.
- Subprocess plus wiremock tests in `crates/clickhousectl/tests/cli_request_shape_test.rs`: `destination.roles` order and de-duplication for a streaming create, its absence when `--role` is omitted, `destination.roles` on a database pipe alongside the four fields that must stay absent, and a reserved-name case asserting exit code 2 with no request recorded by the mock.
- README documents the flag in the `clickpipe create` section with an example.

Verified with:

- `cargo fmt --all --check`
- `cargo test -p clickhousectl` (898 bin tests, 176 request-shape tests, all other targets green)

## Stacking

Part of a stacked PR chain: based on `feat/644-query-timeout`, not `main`.

Fixes #568

### DIFF
--- a/README.md
+++ b/README.md
@@ -961,7 +961,25 @@ clickhousectl cloud clickpipe create bigquery <service-id> \
   --service-account-file ./sa-key.json \
   --staging-path gs://bucket/staging \
   --table-mapping "dataset.table:target_table"
+
+# Grant extra ClickHouse roles to the destination user (any create subcommand)
+clickhousectl cloud clickpipe create kafka <service-id> \
+  --name my-kafka-pipe \
+  --brokers 'broker:9092' --topics events \
+  --format JSONEachRow \
+  --database default --table events \
+  --column "event_id:Int64" \
+  --role analytics_reader --role analytics_writer
 ```
+
+`--role` is available on every `clickpipe create` subcommand and is repeatable.
+ClickPipes creates a ClickHouse user for the pipe; when `--role` is omitted that
+user is granted the default role only, and the request omits the field
+altogether. Each `--role` grants one more existing role on top of the default,
+so roles are added and never taken away. Repeated values are sent once in the
+order given. The names `clickpipes` and `clickpipes_system` are reserved by
+ClickPipes: the CLI rejects them as usage errors (exit code 2) before making
+any request.
 
 #### PostgreSQL ClickPipe prerequisites
 
--- a/crates/clickhousectl/src/cloud/clickpipes.rs
+++ b/crates/clickhousectl/src/cloud/clickpipes.rs
@@ -394,6 +394,20 @@ DOCUMENTATION:
     BigQuery(BigQueryCreateArgs),
 }
 
+/// Destination-permission flags shared by every `clickpipe create` subcommand.
+/// Flattened into each create args struct so `--role` and its help text have a
+/// single definition, the way `KafkaSourceFields` shares the Kafka source flags.
+#[derive(Args, Debug, Default)]
+pub struct DestinationRoleArgs {
+    /// Extra ClickHouse role to grant the ClickPipes destination user (repeatable)
+    ///
+    /// When omitted, ClickPipes grants that user the default role only. Each
+    /// --role adds a role on top of the default; nothing is taken away. The
+    /// API-reserved names `clickpipes` and `clickpipes_system` are rejected.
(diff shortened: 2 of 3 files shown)


iskakaushik (approved):
LGTM

PR #651, opened 2026-09-02
Expose Postgres CDC pipe settings on `clickpipe create postgres`

### LINKED ISSUE #565: Expose Postgres CDC pipe settings on `clickpipe create postgres`
## Problem

The Cloud API's `ClickPipePostgresPipeSettings` supports the full set of CDC tuning options, but `clickhousectl cloud clickpipe create postgres` only wires `replicationMode`, `publicationName`, and `replicationSlotName` — the handler builds the settings struct with `..Default::default()` (`crates/clickhousectl/src/cloud/clickpipes.rs`, `clickpipe_create_postgres`).

Users following the UI flow cannot set at create time:

- `syncIntervalSeconds`
- `pullBatchSize`
- `initialLoadParallelism` (parallel threads for initial load)
- `snapshotNumRowsPerPartition`
- `snapshotNumberOfParallelTables`
- `allowNullableColumns`
- `enableFailoverSlots`
- `deleteOnMerge`

This matters because the snapshot-related settings are **create-time-only** at the API level — `ClickPipePatchPostgresPipeSettings` only allows patching `syncIntervalSeconds` and `pullBatchSize` post-create, so a value not set at create can never be applied to that pipe.

`clickpipe settings update` is a different endpoint (`ClickPipeSettingsPutRequest`, streaming/object-storage only) and cannot cover these.

## Library support

Already complete: `clickhouse-cloud-api` models every field in `ClickPipePostgresPipeSettings` (`crates/clickhouse-cloud-api/src/models/clickpipes.rs`). This is CLI flag work only.

## Proposed change

Add optional flags to `PostgresCreateArgs` for each field above (e.g. `--sync-interval-seconds`, `--pull-batch-size`, `--initial-load-parallelism`, `--snapshot-rows-per-partition`, `--snapshot-parallel-tables`, `--allow-nullable-columns`, `--enable-failover-slots`, `--delete-on-merge`) and wire them into the settings struct. Omitted flags keep today's request shape (fields omitted via `skip_serializing_if`).

MySQL create (`ClickPipeMySQLPipeSettings`) has the analogous gap and could follow the same pattern in a follow-up or the same PR.

## Acceptance criteria

- Each flag maps to exactly one request field; omitted flags leave the field absent from the JSON body.
- Clap parsing tests cover defaults and explicit values per the repo's test conventions.
- Request-shape coverage via the builder tests or `tests/cli_request_shape_test.rs`.

## Context

Found while adding CLI steps to ClickHouse Cloud docs — the UI's Postgres CDC advanced settings had no CLI equivalent.

sdairs:
## What

`clickpipe create postgres` now accepts one flag per field of
`ClickPipePostgresPipeSettings`:

- `--sync-interval-seconds <SECONDS>`
- `--pull-batch-size <ROWS>`
- `--initial-load-parallelism <WORKERS>`
- `--snapshot-rows-per-partition <ROWS>` (`snapshotNumRowsPerPartition`)
- `--snapshot-parallel-tables <TABLES>` (`snapshotNumberOfParallelTables`)
- `--allow-nullable-columns <true|false>`
- `--enable-failover-slots <true|false>`
- `--delete-on-merge <true|false>`

The settings struct is now built field by field in a new
`build_postgres_pipe_settings` helper instead of `..Default::default()`, so a
settings field added to the library in future is a compile error here rather
than a silent default. Help text per flag comes from the OpenAPI field
descriptions, and a new `POSTGRES CDC SETTINGS` section of `--help` states
which settings cannot be changed after creation.

## Why

The handler wired only `replicationMode`, `publicationName` and
`replicationSlotName`, and the snapshot and initial-load settings are
create-time-only at the API: `ClickPipePatchPostgresPipeSettings` can patch
only `syncIntervalSeconds` and `pullBatchSize`, and `clickpipe settings update`
is a different endpoint that only covers streaming and object-storage pipes. A
value not set at create time could therefore never be applied to that pipe, so
users following the Console's advanced Postgres CDC settings had no CLI
equivalent at all. No API library change was needed: the library already models
every field.

## Behaviour after the fix

Numeric settings are omitted from `source.postgres.settings` when their flag is
omitted, via `skip_serializing_if`, so the request shape for an existing
invocation is unchanged.

`allowNullableColumns`, `deleteOnMerge` and `enableFailoverSlots` are required
by the schema, so they are plain `bool` in the library request type and always
serialize. Absence is not representable for them on the wire: an omitted flag
sends `false`, which is exactly what the CLI already sent through
`Default::default()`. The three flags are still tri-state at the clap layer
(`Option<bool>` taking an explicit `true|false`), so a user can pass `false`
deliberately rather than relying on the default, and the request keeps its
current shape when nothing is passed. This is documented in the flag help, in
the `--help` section, and in the README.

MySQL create has the analogous gap on `ClickPipeMySQLPipeSettings`. It is left
as a follow-up: the MySQL handler builds its request inline rather than through
an extracted builder, so the work is not mechanical in the way it was here. The
README says the settings are not yet exposed on `clickpipe create mysql`.

## Tests

- Clap: `parses_postgres_cdc_settings_flags` asserts every explicit value
  including an explicit boolean `false`;
  `parses_postgres_flags_defaults_and_repeatability` gained the all-`None`
  default assertions; `postgres_cdc_settings_booleans_require_an_explicit_value`
  pins that the booleans are not bare switches.
- Builder: `build_postgres_pipe_settings_omits_unset_settings_and_sends_explicit_false`
  asserts the library struct fields and the serialized key set;
  `build_postgres_pipe_settings_defaults_optional_booleans_to_false` pins the
  omitted-flag shape; the existing minimal and maximal
  `build_postgres_request` tests cover both ends.
- Subprocess plus wiremock in `tests/cli_request_shape_test.rs`:
  `postgres_cdc_settings_are_absent_unless_their_flags_are_passed` and
  `postgres_cdc_settings_serialize_exactly_the_flags_that_were_passed` pin the
  exact key set of `source.postgres.settings` in the POST body.
- Text pinned: `postgres_help_documents_input_tls_and_source_requirements` for
  the new flags and help section, `readme_documents_postgres_cdc_settings_flags`
  for the README table and caveats.
- Verification commands:
  - `cargo fmt --all --check` (clean)
  - `cargo clippy --workspace --all-targets -- -D warnings` (clean)
  - `cargo clippy -p clickhousectl --all-targets --no-default-features -- -D warnings` (clean)
  - `cargo test -p clickhousectl` (23 suites, all ok; 903 unit and 178
    request-shape tests)

## Stacking

Part of a stacked PR chain: based on `feat/568-clickpipe-role`, not `main`.

Fixes #565

### DIFF
--- a/README.md
+++ b/README.md
@@ -938,6 +938,18 @@ clickhousectl cloud clickpipe create postgres <service-id> \
   --publication-name clickpipes \
   --table-mapping "public.users:public_users"
 
+# From PostgreSQL, tuning the CDC and initial-load settings
+clickhousectl cloud clickpipe create postgres <service-id> \
+  --name my-tuned-pg-pipe \
+  --host db.example.com --pg-database mydb \
+  --username "$POSTGRES_USERNAME" --password "$POSTGRES_PASSWORD" \
+  --publication-name clickpipes \
+  --table-mapping "public.users:public_users" \
+  --sync-interval-seconds 30 --pull-batch-size 50000 \
+  --initial-load-parallelism 4 \
+  --snapshot-rows-per-partition 1000000 --snapshot-parallel-tables 3 \
+  --allow-nullable-columns true --delete-on-merge false
+
 # From MySQL (CDC)
 # --server-id sets the replication server ID (useful when multiple pipes read
 # from the same MySQL instance, or to avoid colliding with existing replicas)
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #652, opened 2026-09-02
Expose per-table mapping options on `clickpipe create postgres`

### LINKED ISSUE #566: Expose per-table mapping options on `clickpipe create postgres` (excluded columns, sorting keys, PARTITION BY, table engine)
## Problem

`clickhousectl cloud clickpipe create postgres` accepts table mappings only as `--table-mapping schema.table:target_table`. The handler fills `sourceSchemaName`/`sourceTable`/`targetTable` and defaults everything else (`clickpipe_create_postgres` in `crates/clickhousectl/src/cloud/clickpipes.rs`).

The API's `ClickPipePostgresPipeTableMapping` supports per-table:

- `excludedColumns` — column exclusion
- `useCustomSortingKey` + `sortingKeys` — custom ORDER BY key
- `partitionByExpr` — ClickHouse PARTITION BY expression for the created table
- `partitionKey` — snapshot-parallelism partitioning column (PG13 only benefit; unrelated to ClickHouse partitioning)
- `tableEngine` — MergeTree / ReplacingMergeTree / Null

None are reachable from the CLI, and since these shape the destination table created by ClickPipes, they cannot be fixed after creation.

## Library support

Already complete: all fields exist on `ClickPipePostgresPipeTableMapping` in `crates/clickhouse-cloud-api/src/models/clickpipes.rs`. CLI-only work.

## Proposed change

Needs a design decision on syntax, since these are per-table rather than per-pipe. Options:

1. Extend `--table-mapping` to a key=value list, e.g. `--table-mapping 'source=public.users,target=users,exclude=a;b,sorting-keys=created_at;id,engine=ReplacingMergeTree,partition-by=toYYYYMM(created_at)'`.
2. A `--table-mapping-json` flag accepting the API's mapping object verbatim (good for agents, avoids inventing a mini-grammar).
3. Both: keep the simple colon form for the common case, add the JSON form for full fidelity.

MySQL/MongoDB/BigQuery table mappings have analogous fields and should adopt the same syntax once decided.

## Acceptance criteria

- All five fields settable per table at create time; the simple `schema.table:target` form keeps working unchanged.
- `useCustomSortingKey` is set automatically when sorting keys are provided.
- Clap and request-builder tests cover minimal and maximal mappings.

## Context

Found while adding CLI steps to ClickHouse Cloud docs — the UI's per-table column exclusion, custom ordering keys, and custom PARTITION BY had no CLI equivalent.

sdairs:
## What

`clickpipe create postgres` took table mappings only as `--table-mapping schema.table:target_table`, which filled `sourceSchemaName`, `sourceTable` and `targetTable` and defaulted every other field of `ClickPipePostgresPipeTableMapping`. This adds a repeatable `--table-mapping-json <JSON>` that takes the API's table mapping object verbatim, and leaves the simple colon form exactly as it was.

The two flags are one required "at least one of" group: either alone is enough, both may be given together, and clap names both when neither is given. They combine into a single mappings list, with the `--table-mapping` values first and then the `--table-mapping-json` ones, each in the order given.

Every JSON value is parsed and validated before any request is made, so a bad mapping is a usage error (exit code 2) that never reaches the network or reads a credential file:

- The value must be JSON, must be an object, and may carry only the nine wire fields, so a typo such as `excludeColumns` fails loudly instead of being dropped on the way to the API.
- `sourceSchemaName`, `sourceTable` and `targetTable` are required and must not be empty; list entries are trimmed and must not be empty.
- `useCustomSortingKey` is set to `true` for the caller when `sortingKeys` is given, because the API ignores the keys without it. `useCustomSortingKey: false` alongside keys, and `true` with no keys, are both rejected rather than silently dropping or ignoring the keys.
- An unknown `tableEngine` is rejected locally against the library enum's wire values, since the destination table's engine cannot be changed after creation.
- Each diagnostic names the offending occurrence, for example `--table-mapping-json #2: targetTable is required and must not be empty`.

Only to add `ClickPipePostgresPipeTableMappingTableengine::VALUES`. The enum has an `Unknown(String)` catch-all, so a misspelled engine deserializes rather than failing, and the CLI needs the accepted wire values as data instead of a hand-copied list. `VALUES` is the repository's convention for exactly that, and the drift analyzer checks it stays equal to the enum's variant wire values, so a new engine in the spec cannot land without updating the const, the CLI validation and the help text together. No model shape changed.

## Behaviour after the fix

```bash
clickhousectl cloud clickpipe create postgres <service-id> \
  --name my-pg-pipe \
  --host db.example.com --pg-database mydb \
  --username "$POSTGRES_USERNAME" --password "$POSTGRES_PASSWORD" \
  --publication-name clickpipes \
  --table-mapping "public.orders:public_orders" \
  --table-mapping-json '{
    "sourceSchemaName": "public",
    "sourceTable": "users",
    "targetTable": "public_users",
    "excludedColumns": ["ssn"],
    "sortingKeys": ["created_at", "id"],
    "partitionByExpr": "toYYYYMM(created_at)",
    "tableEngine": "ReplacingMergeTree"
  }'
```

`tableEngine` defaults to `MergeTree`, which is what the simple form has always sent, so an omitted field means the same request the CLI made before. `partitionKey` partitions the initial snapshot for parallelism and is documented as unrelated to the destination table's `PARTITION BY`, which is `partitionByExpr`.

Two things were deliberately left out. There is no `@path` form for reading the JSON from a file: the repository's existing convention for file input is a separate `--*-file` flag whose value may be `-` for stdin (`--queries-file`, `--ca-certificate`, `--service-account-file`), and no `@value` convention exists to follow, so adding one here would have invented it. Duplicate mappings for the same source table are not rejected, since fanning one source table into two destinations is not obviously invalid and the API is the authority on that.

MySQL, MongoDB and BigQuery table mappings are untouched. They have analogous per-table fields and should adopt `--table-mapping-json` with the same semantics in a follow-up.

## Tests

- Clap parsing next to the command definition: JSON only, both flags together, simple only, neither (missing-argument error naming both flags), and that clap collects the raw string while validation happens in `validate_postgres_create_args`.
- Unit tests for `parse_postgres_table_mapping_json`: minimal object, maximal object with all nine fields, automatic `useCustomSortingKey`, every table engine from `VALUES`, and thirteen rejection cases (invalid JSON, non-object, unknown field singular and plural, wrong JSON type, each missing or empty required field, empty list entries, both sorting-key contradictions, unknown engine), each asserted to name its flag occurrence.
- Builder tests: both flags combined into one ordered mappings list with the library struct fields asserted directly, JSON mappings alone, and the two new cases in the defensive rejection table.
- Subprocess and wiremock (`tests/cli_request_shape_test.rs`): the POST body's `tableMappings` array reproduces the JSON fields exactly and keeps the simple form's wire shape unchanged, JSON mappings alone produce the array, and five invalid values exit 2 with no request received.
- Help and README text pinned: the `POSTGRES TABLE MAPPINGS` help section and the README section must name every wire field and every table engine, so neither can drift from the code.
- Library: `ClickPipePostgresPipeTableMappingTableengine::VALUES` round-trips each value without falling through to the catch-all.

Verification:

- `cargo fmt --all` then `cargo fmt --all --check` clean
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo clippy -p clickhousectl --all-targets --no-default-features -- -D warnings` clean
- `cargo test -p clickhousectl` 915 passed, 1 ignored, plus every integration binary green
- `cargo test -p clickhouse-cloud-api --lib --test spec_coverage_test --test models_test` 18 + 289 + 6 passed, 1 ignored
- `cargo check --workspace --all-features` clean
- `cargo test -p clickhouse-openapi-analyzer` 55 passed
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` 90 passed

## Stacking

Part of a stacked PR chain: based on `feat/565-postgres-cdc-settings`, not `main`.

Fixes #566

### DIFF
--- a/README.md
+++ b/README.md
@@ -938,6 +938,23 @@ clickhousectl cloud clickpipe create postgres <service-id> \
   --publication-name clickpipes \
   --table-mapping "public.users:public_users"
 
+# From PostgreSQL, shaping the destination tables with the JSON mapping form
+clickhousectl cloud clickpipe create postgres <service-id> \
+  --name my-shaped-pg-pipe \
+  --host db.example.com --pg-database mydb \
+  --username "$POSTGRES_USERNAME" --password "$POSTGRES_PASSWORD" \
+  --publication-name clickpipes \
+  --table-mapping "public.orders:public_orders" \
+  --table-mapping-json '{
+    "sourceSchemaName": "public",
+    "sourceTable": "users",
+    "targetTable": "public_users",
+    "excludedColumns": ["ssn"],
+    "sortingKeys": ["created_at", "id"],
+    "partitionByExpr": "toYYYYMM(created_at)",
+    "tableEngine": "ReplacingMergeTree"
+  }'
+
 # From PostgreSQL, tuning the CDC and initial-load settings
 clickhousectl cloud clickpipe create postgres <service-id> \
   --name my-tuned-pg-pipe \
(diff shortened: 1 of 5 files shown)


iskakaushik (approved):
LGTM

PR #653, opened 2026-09-02
clickpipe schema-discover: support object storage sources

### LINKED ISSUE #588: clickpipe schema-discover: support object storage sources
From the CLI↔API gap analysis (2026-08-27): `ClickPipeSchemaDiscoverySource` (`crates/clickhouse-cloud-api/src/models/clickpipes.rs:2943`) supports `kafka`, `kinesis`, `objectStorage`, and `pubsub`, but `ClickPipeSchemaDiscoverCommands` (`crates/clickhousectl/src/cloud/clickpipes.rs:236`) only has Kafka and Kinesis subcommands — the other sources are hardcoded to `None` at clickpipes.rs:1441-1452.

This issue covers the `objectStorage` variant (`clickpipe schema-discover object-storage`, mirroring the flags of `clickpipe create object-storage`). The `pubsub` variant is covered by the Pub/Sub source issue.

sdairs:
## What

`clickhousectl cloud clickpipe schema-discover <SERVICE_ID> object-storage` now
infers the schema of an S3, GCS, Azure Blob Storage (or any other supported
object store) source without creating a ClickPipe, alongside the existing
`kafka` and `kinesis` subcommands. It takes the same source connection flags as
`clickpipe create object-storage` minus the destination options: `--source-url`,
`--format`, `--storage-type`, `--compression`, `--continuous`, `--queue-url`,
`--skip-initial-load`, `--start-after`, `--delimiter`, `--iam-role`,
`--access-key-id`, `--secret-key`, `--connection-string`,
`--azure-container-name`, `--path` and `--service-account-file`.

To keep the flag set defined once, the source flags moved out of
`ObjectStorageCreateArgs` into a new `ObjectStorageSourceFields` that is
flattened back into it, exactly as `KafkaSourceFields` and
`KinesisSourceFields` are already shared between creation and discovery.
Authentication inference, the GCP service-account file read and the enum
parsing moved into a shared `build_object_storage_source` helper, and
`build_object_storage_schema_discovery_request` wraps it into the library
request struct with every other source key left absent.

## Why

`ClickPipeSchemaDiscoverySource` in the API library has carried an
`objectStorage` field all along, but `ClickPipeSchemaDiscoverCommands` only had
Kafka and Kinesis variants and hard-coded the other sources to `None`. There
was no way to discover the columns of an object-storage source to fill in
`--column` on `clickpipe create`. The library needed no changes.

## Behaviour after the fix

Discovery and creation build the identical `objectStorage` source, including the
`--iam-role` over access-key-pair over connection-string over service-account
authentication precedence and the base64-encoded GCP key file, so a source that
discovers cleanly is a source that creates cleanly. Object-storage discovery is
classified as a write like the other discovery subcommands, because the API
gateway rejects OAuth/JWT on the `schemaDiscovery` POST; it therefore requires
API key auth. Output is the same field table as Kafka and Kinesis discovery, or
the raw response with `--json`.

One incidental improvement: `clickpipe create object-storage` now builds its
source before resolving the organization ID, so an unreadable
`--service-account-file` fails before any network call, matching what
`create kafka` and `create kinesis` already did.

## Tests

- Clap: `parses_object_storage_schema_discovery_flags` pins every new flag, the
  `--storage-type s3` / `--compression auto` defaults, that only `--source-url`
  and `--format` are required, and that `--skip-initial-load` still requires
  `--queue-url` and still conflicts with `--start-after`.
- Builders: `build_object_storage_schema_discovery_request_minimal` and
  `..._maximal` assert on the library request struct fields, including that
  `kafka`, `kinesis` and `pubsub` stay absent, that `--iam-role` wins over the
  other credential flags, that the access key pair infers `IAM_USER`, and that
  the service-account file is read and base64-encoded.
- Write classification: `clickpipe_write_classification_delegates_from_cloud_commands`
  now covers `schema-discover kinesis` and `schema-discover object-storage`.
- Subprocess and wiremock: `schema_discover_object_storage_posts_source_body`
  asserts the POST body carries `source.objectStorage` with the passed fields,
  no `accessKey` the user did not pass, and no other source key.
- Existing object-storage create clap tests were updated for the flattened
  `args.source.*` path.
- Commands run: `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D
  warnings` (clean), `cargo clippy -p clickhousectl --all-targets
  --no-default-features -- -D warnings` (clean), `cargo test -p clickhousectl`
  (all suites pass: 918 unit tests, 182 request-shape tests).

## Stacking

Part of a stacked PR chain: based on `feat/566-postgres-table-mapping-json`, not `main`.

Fixes #588

### DIFF
--- a/README.md
+++ b/README.md
@@ -1140,8 +1140,8 @@ Use `clickhousectl cloud clickpipe create <source> --help` for the full list of 
 
 #### Discovering a source schema (beta)
 
-`clickpipe schema-discover` probes a Kafka or Kinesis source and returns the
-inferred fields/types without creating a pipe. It takes the same source
+`clickpipe schema-discover` probes a Kafka, Kinesis or object-storage source and
+returns the inferred fields/types without creating a pipe. It takes the same source
 connection flags as the corresponding `create` subcommand (minus the
 destination `--name`/`--database`/`--table`/`--column` options). Schema discovery requires API-key authentication:
 
@@ -1163,6 +1163,12 @@ clickhousectl cloud clickpipe schema-discover <service-id> kinesis \
   --stream-name events --region us-east-1 \
   --format JSONEachRow \
   --auth IAM_ROLE --iam-role "$KINESIS_IAM_ROLE_ARN"
+
+# Discover schema from object storage (S3, GCS, Azure Blob Storage)
+clickhousectl cloud clickpipe schema-discover <service-id> object-storage \
+  --source-url 'https://bucket.s3.us-east-1.amazonaws.com/data/*.json' \
+  --format JSONEachRow \
+  --iam-role "$S3_IAM_ROLE_ARN"
 ```
 
 Add `--json` (or run as a coding agent) for machine-readable output.
--- a/crates/clickhousectl/src/cloud/clickpipes.rs
+++ b/crates/clickhousectl/src/cloud/clickpipes.rs
@@ -203,9 +203,11 @@ the pipe itself, so read them with `clickhousectl cloud clickpipe get <service>
     /// Discover a source schema without creating a pipe (beta)
     #[command(after_help = "\\
 CONTEXT FOR AGENTS:
-  Infers the schema (column name + ClickHouse type) for a Kafka or Kinesis source
-  without creating a ClickPipe. Useful for filling in --column on `clickpipe create`.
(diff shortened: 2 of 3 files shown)


iskakaushik (approved):
LGTM

Issue #206, opened 2026-05-21
ClickPipes: live E2E test for Pub/Sub source


sdairs:
Issue #203 added Pub/Sub source models (`ClickPipePubSubSource`, `ClickPipePostPubSubSource`, `ClickPipePatchPubSubSource`) to the Rust client library. They're covered by wiremock + serde tests, but there's no live E2E test for the source yet.

We need a `clickpipe_pubsub_test` binary under `crates/clickhouse-cloud-api/tests/clickpipes/`, matching the pattern of the existing source binaries (`s3_test.rs`, `kafka_test.rs`, `kinesis_test.rs`, etc.).

This will require GCP credentials and Pub/Sub topic provisioning — we'll work out the details when we pick this up.

PR #654, opened 2026-09-02
Add Google Cloud Pub/Sub ClickPipe source support

### LINKED ISSUE #587: Add Google Cloud Pub/Sub ClickPipe source support
From the CLI↔API gap analysis (2026-08-27): Pub/Sub is the only source slot in `ClickPipePostSource` (`crates/clickhouse-cloud-api/src/models/clickpipes.rs:3048`) with no `clickpipe create` subcommand at all.

The library models it fully:
- Create: `ClickPipePostPubSubSource` — `topic`, `projectId`, `format` (JSONEachRow/Avro/Protobuf), `authentication` (SERVICE_ACCOUNT), `serviceAccountKey`, `seekType` (latest/earliest/timestamp), `seekTimestamp`, `ackDeadline`, `enableOrdering`, `filter`
- Patch: `ClickPipePatchPubSubSource` (relevant once #569 lands)
- Schema discovery: `ClickPipeSchemaDiscoverySource.pubsub`

Suggested surface: `clickpipe create pubsub` plus a `pubsub` variant of `clickpipe schema-discover`. Service-account key input should support file/stdin.

Related: #206 asks for a live E2E test for the Pub/Sub source in the library — CLI exposure and that test could land together.

sdairs:
## What

`clickpipe create pubsub` and `clickpipe schema-discover <SERVICE_ID> pubsub`, the last
source arm of `ClickPipePostSource` that had no CLI surface at all.

Both subcommands share a flattened `PubSubSourceFields` and a single
`build_pubsub_source` helper, the way Kafka, Kinesis and object storage already do, so
discovery and creation send an identical `source.pubsub` object. Requiredness follows the
library's strict request type: `--topic`, `--project-id`, `--format`, `--seek-type` and
`--service-account-file` are required flags, and `--seek-timestamp`, `--filter`,
`--enable-ordering` and `--ack-deadline` are optional and left out of the body when unset.
`--auth` is the one exception: the wire field is required but has exactly one accepted
value, so it defaults to `SERVICE_ACCOUNT` instead of making every invocation repeat it.

The destination flags, `--column` and `--role` behave exactly as on `create kafka`,
through the same `build_destination` and `build_destination_roles` helpers.

The API's own constraints are enforced before a request goes out, so a bad invocation
fails locally rather than at the API: `--seek-timestamp` is required with `--seek-type
timestamp` and refused with the other seek types, `--ack-deadline` is bounded to 10 to 600
seconds, and `--filter` to 256 characters. Every help string comes from the OpenAPI
description for the field.

## Why the service account key is a file or stdin, not an inline value

The issue and the brief suggested a `--service-account-key` flag taking a literal JSON
string, `@path` or `-`. The repo already has a convention for exactly this credential:
`--service-account-file <PATH>` on `create object-storage` and `create bigquery`, read by
`read_gcp_service_account_file` and base64-encoded into the wire field. There is no `@path`
sigil anywhere in the CLI, and an inline JSON key would land in process listings and shell
history. So Pub/Sub reuses `--service-account-file` verbatim, and the shared reader gains
one thing instead: a path of `-` reads the key from stdin, the same spelling
`service query --queries-file -` already uses, which is what makes piping a key out of a
secret manager possible without writing it to disk. That widening applies to the
object-storage and BigQuery flags too, which is coherent rather than surprising.

The reader also now refuses an empty key (empty file or empty stdin) instead of sending an
empty string. The key itself never reaches an error message, help text or telemetry: only
the path is named, and the base64 blob only ever goes into the request body.

## Behaviour after the fix

```bash
clickhousectl cloud clickpipe create pubsub <service-id> \
  --name my-pubsub-pipe \
  --topic events --project-id my-gcp-project \
  --format JSONEachRow \
  --seek-type earliest \
  --service-account-file ./sa-key.json \
  --database default --table events \
  --column "event_id:Int64"

gcloud secrets versions access latest --secret=clickpipes-sa-key |
  clickhousectl cloud clickpipe schema-discover <service-id> pubsub \
    --topic events --project-id my-gcp-project \
    --format JSONEachRow \
    --seek-type latest \
    --service-account-file -
```

Both variants are classified as writes: `create` is one, and schema discovery is
side-effect-free but the API gateway rejects OAuth on that POST, so it keeps the existing
fail-fast API-key guidance.

## Tests

- Clap parsing next to the command definitions: every flag and default on `create pubsub`,
  the required-flag set derived from the strict request type, the `--seek-type timestamp`
  and `--seek-timestamp` pairing in both directions, the `--ack-deadline` boundaries (10
  and 600 accepted, 9, 601 and 0 rejected), the 256-character `--filter` limit, and the
  full `schema-discover pubsub` flag set.
- The accepted Pub/Sub formats, auth values and seek types are pinned and every one of
  them is exercised through clap, with unknown values rejected, matching the existing
  per-source pins.
- Request builders: minimal and maximal `build_pubsub_source` and
  `build_pubsub_schema_discovery_request` cases asserting on library struct fields,
  including the base64 key, UTC normalization of an offset timestamp, and that no other
  source arm is populated. A missing and an empty key file are asserted to fail with no key
  content in the message.
- Write classification for `create pubsub` and `schema-discover pubsub`, plus `--role` on
  `create pubsub` in the existing every-subcommand test.
- Subprocess plus wiremock in `cli_request_shape_test.rs`: the POST body's `source.pubsub`
  object for create (with the key read from a file, base64-encoded, and neither the path
  nor the raw key on the wire), optional fields absent when the flags are omitted, the same
  key read from stdin with `--service-account-file -`, the schema-discovery body, and a
  rejected create where the API error is still reported but neither the key, its private
  key material nor its base64 encoding appears in stdout or stderr.
- Help and README text pinned, as the repo does for the Postgres surface.
- `cargo fmt --all --check`: clean.
- `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- `cargo clippy -p clickhousectl --all-targets --no-default-features -- -D warnings`: clean.
- `cargo test -p clickhousectl`: 23 test binaries, 0 failures (930 unit tests, 187
  request-shape tests).
- The API library is untouched, so its suites and the drift snapshot are unaffected.

#206, the live E2E library test for the Pub/Sub source, stays open: this PR adds no cloud
integration test, and none of the work here needed live Cloud access.

## Stacking

Part of a stacked PR chain: based on `feat/588-schema-discover-object-storage`, not `main`.

Fixes #587

### DIFF
--- a/README.md
+++ b/README.md
@@ -991,6 +991,33 @@ clickhousectl cloud clickpipe create bigquery <service-id> \
   --staging-path gs://bucket/staging \
   --table-mapping "dataset.table:target_table"
 
+# From Google Cloud Pub/Sub (limited preview: contact support to enable it)
+# --seek-type has no default: earliest reads the backlog, latest only new
+# messages, timestamp starts from --seek-timestamp (required for that seek
+# type, and rejected for the others).
+clickhousectl cloud clickpipe create pubsub <service-id> \
+  --name my-pubsub-pipe \
+  --topic events --project-id my-gcp-project \
+  --format JSONEachRow \
+  --seek-type earliest \
+  --service-account-file ./sa-key.json \
+  --database default --table events \
+  --column "event_id:Int64" --column "name:String"
+
+# Pub/Sub with the optional subscription tuning, reading the key from stdin
+# so it never has to be written to disk
+gcloud secrets versions access latest --secret=clickpipes-sa-key |
+  clickhousectl cloud clickpipe create pubsub <service-id> \
+    --name my-tuned-pubsub-pipe \
+    --topic events --project-id my-gcp-project \
+    --format Avro \
+    --seek-type timestamp --seek-timestamp 2026-04-10T12:00:00Z \
+    --service-account-file - \
+    --filter 'attributes.region = "eu"' \
+    --enable-ordering --ack-deadline 120 \
+    --database default --table events \
+    --column "event_id:Int64" --column "name:String"
+
 # Grant extra ClickHouse roles to the destination user (any create subcommand)
 clickhousectl cloud clickpipe create kafka <service-id> \
   --name my-kafka-pipe \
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #655, opened 2026-09-02
Add `clickpipe reverse-private-endpoint` management commands

### LINKED ISSUE #567: Add `clickpipe reverse-private-endpoint` management commands (PrivateLink support)
## Problem

The Cloud API has full CRUD for ClickPipes reverse private endpoints (`clickPipeReversePrivateEndpointGetList/Create/Get/Update/Delete`), supporting types `VPC_ENDPOINT_SERVICE`, `VPC_RESOURCE`, `MSK_MULTI_VPC`, and `GCP_PSC_SERVICE_ATTACHMENT`. This is how PrivateLink connectivity works for ClickPipes sources: create the endpoint, then reference it — by ID for Kafka pipes, or by using its hostname as `host` for Postgres/MySQL CDC pipes (per the spec's `host` field description).

The CLI has no way to manage these. The only related surface is `--reverse-private-endpoint-id` on `clickpipe create kafka`, which is useless without a way to create or list endpoints. PrivateLink-connected Postgres CDC is therefore impossible end-to-end from the CLI.

## Library support

Already complete: all five methods exist in `crates/clickhouse-cloud-api/src/client/clickpipes.rs` (`click_pipe_reverse_private_endpoint_*`), with `CreateReversePrivateEndpoint` / `UpdateReversePrivateEndpoint` / `ReversePrivateEndpoint` models. CLI-only work.

## Proposed change

Add a `clickhousectl cloud clickpipe reverse-private-endpoint` subcommand group:

- `list <service_id>`
- `get <service_id> <endpoint_id>`
- `create <service_id> --type ... --description ...` plus type-specific flags (`--vpc-endpoint-service-name`, `--vpc-resource-configuration-id`, `--vpc-resource-share-arn`, `--msk-cluster-arn`, `--msk-authentication`, `--gcp-service-attachment`, repeatable `--custom-private-dns-mapping`)
- `update <service_id> <endpoint_id> ...`
- `delete <service_id> <endpoint_id>`

Standard `--json` support; list as a `tabled` table, get via `print_human`. Read/write classification: list/get read, create/update/delete write.

Docs/help should note the Postgres CDC usage: pass the endpoint's DNS name as `--host` on `clickpipe create postgres`.

## Acceptance criteria

- Full CRUD reachable from the CLI with clap parsing tests and request-shape coverage.
- Type-specific required fields validated client-side where the spec makes them unambiguous.

## Context

Found while adding CLI steps to ClickHouse Cloud docs — the UI's AWS PrivateLink connection option for Postgres ClickPipes had no CLI equivalent. (SSH tunneling, the other UI connection option, is absent from the OpenAPI spec entirely and is being raised as an API gap separately.)

sdairs:
## What

Adds `clickhousectl cloud clickpipe reverse-private-endpoint` with the full CRUD set:

- `list <service-id>` renders a `tabled` table (ID, Type, Description, Status, DNS Names) and supports `--json`.
- `get <service-id> <endpoint-id>` renders through `print_human` and supports `--json`.
- `create <service-id> --type <TYPE> --description <D>` plus the type-specific flags: `--vpc-endpoint-service-name`, `--vpc-resource-configuration-id`, `--vpc-resource-share-arn`, `--msk-cluster-arn`, `--msk-authentication`, `--gcp-service-attachment`, and the repeatable `--custom-private-dns-mapping`.
- `update <service-id> <endpoint-id> --custom-private-dns-mapping <NAME>` sends the complete mapping list, the only field the API's PATCH accepts.
- `delete <service-id> <endpoint-id>` prints a plain confirmation line, or `{"deleted": <id>}` under `--json`.

The surface lives in a new `crates/clickhousectl/src/cloud/clickpipe_endpoints.rs` rather than in `clickpipes.rs`, which is already several thousand lines of pipe-creation surface, and nests under `clickpipe` because an endpoint is only ever useful to a pipe. It is wired per CLAUDE.md: a domain module, a private re-export from `src/cloud/cli.rs`, an exhaustive `is_write()` (list and get read, create, update and delete write), a dispatch arm in the `clickpipe` dispatcher, five thin `CloudClient` wrappers, and `build_*_request` helpers.

The group's `long_about` explains how a pipe references an endpoint: the endpoint ID for Kafka via `--reverse-private-endpoint-id`, and one of the endpoint's DNS names as `--host` for Postgres and MySQL CDC. It also says a pipe can only use an endpoint that has reached `Ready`, and that an AWS PrivateLink endpoint stays in `PendingAcceptance` until the connection request is accepted in the account that owns the source.

## Why

The Cloud API has had full CRUD for ClickPipes reverse private endpoints for a while, but the CLI had no way to reach it. `clickpipe create kafka --reverse-private-endpoint-id` could only reference an endpoint somebody else had created, and a PrivateLink-connected Postgres CDC pipe was impossible end to end from the CLI, because that pipe connects to the endpoint's DNS name rather than to an ID.

`CreateReversePrivateEndpointType` and `CreateReversePrivateEndpointMskauthentication` are now validated by the CLI as `--type` and `--msk-authentication` string arguments, so each declares the opt-in `pub const VALUES` the analyzer checks against its variant wire values. That keeps the accepted values in one place, checkable in CI, instead of a hand-copied literal list in the CLI.

## Behaviour after the fix

Client-side validation of `create` runs after parsing, because clap cannot forbid an argument based on another argument's value. It is a usage error (exit code 2) against the owning command, raised before any request and before the organization lookup, matching how `clickpipe create postgres` validates its cross-value relationships. Only the rules the OpenAPI spec states are enforced:

| `--type` | Required flags |
| --- | --- |
| `VPC_ENDPOINT_SERVICE` | `--vpc-endpoint-service-name` |
| `VPC_RESOURCE` | `--vpc-resource-configuration-id`, `--vpc-resource-share-arn` |
| `MSK_MULTI_VPC` | `--msk-cluster-arn`, `--msk-authentication` |
| `GCP_PSC_SERVICE_ATTACHMENT` | `--gcp-service-attachment` |

A flag belonging to another type is rejected by name ("`--msk-cluster-arn` applies to `--type MSK_MULTI_VPC`, not `--type VPC_ENDPOINT_SERVICE`"), and `--custom-private-dns-mapping` is refused for `MSK_MULTI_VPC` because the spec says it is not supported there. `vpcEndpointServiceName` is the one field the spec does not mark "Required for `<TYPE>` type", but it names its type exactly and has no meaning for any other type, so it is treated as `VPC_ENDPOINT_SERVICE`'s required flag. Everything else, including whether a value is well formed, is left to the API.

`delete` copies `clickpipe delete` exactly: no interactive confirmation and no `--force`, because no cloud delete in this CLI prompts today, and `service delete --force` means "stop the service first" rather than "skip a prompt". Inventing a prompt here would have made this the only command that has one.

Absence is rendered, never unwrapped: an absent list-row field shows `-`, and an endpoint whose DNS names are absent or empty renders the same placeholder for both, since neither tells the caller a name to connect to.

## Tests

- `crates/clickhousectl/src/cloud/clickpipe_endpoints.rs`: clap `try_parse_from` coverage for all five subcommands, including every type's flags, rejected `--type` and `--msk-authentication` values, `create` requiring `--type` and `--description`, and `update` requiring at least one mapping; `build_create_reverse_private_endpoint_request` minimal and maximal cases asserting on library request-struct fields; `build_update_reverse_private_endpoint_request`; the validation messages per type; read and write classification.
- `crates/clickhousectl/src/cloud/cli.rs`: the new commands added to the top-level `is_write_command()` read and write classification tests.
- `crates/clickhousectl/tests/cli_request_shape_test.rs`: subprocess plus wiremock coverage for list (JSON array, human table with placeholders, empty collection), get, create for each of the four types asserting the exact POST body, the human create output naming the status to wait for, update's PATCH body, delete in both output modes, the three client-side validation cases asserting exit code 2 and that the mock received nothing, and a help-text assertion pinning the Kafka and Postgres/MySQL usage guidance.
- `crates/clickhouse-cloud-api/tests/models_test.rs`: both `VALUES` consts and their round-trips, plus `ReversePrivateEndpoint` resolving a missing key and an explicit `null` to `None` and omitting both on serialization.
- `README.md`: a new "Reverse private endpoints (PrivateLink, Private Service Connect)" subsection under ClickPipes with one create example per type, the required-flag table, and how a pipe references the endpoint.

Verification commands, all green:

- `cargo fmt --all`
- `cargo test -p clickhouse-cloud-api --lib --test spec_coverage_test --test models_test`
- `cargo test -p clickhouse-openapi-analyzer`

No cloud resources were created: every new test runs against a local wiremock server.

## Stacking

Part of a stacked PR chain: based on `feat/587-pubsub-source`, not `main`.

Fixes #567

### DIFF
--- a/README.md
+++ b/README.md
@@ -828,6 +828,9 @@ clickhousectl cloud clickpipe scale <service-id> <clickpipe-id> \
 clickhousectl cloud clickpipe settings get <service-id> <clickpipe-id>
 clickhousectl cloud clickpipe settings update <service-id> <clickpipe-id> \
   --streaming-max-insert-wait-ms 10000
+
+# Manage reverse private endpoints for private source connectivity
+clickhousectl cloud clickpipe reverse-private-endpoint list <service-id>
 ```
 
 `settings get` and `settings update` apply to streaming (Kafka, Kinesis) and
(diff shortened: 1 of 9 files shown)


iskakaushik (approved):
LGTM

Issue #527, opened 2026-08-25
Clean up retired Query API keys during stale-key repair


sdairs:
## Problem

PR #508 repairs a rejected stored Query API credential by removing its local `service_query_keys.<service-id>` record. A rerun provisions a replacement, but the retired Cloud API key is not deleted and its UUID remains in the query endpoint `openApiKeys` list. Local metadata then tracks only the replacement.

Service deletion cleans only the currently stored replacement key, so the retired key survives after the service is deleted. Repeated repairs grow the organization key inventory and endpoint key list. If only the local secret was corrupted, the retired Cloud key can remain enabled and usable while becoming untracked.

Relevant code on PR #508:

- `crates/clickhousectl/src/cloud/services.rs:1790-1815` removes only the matching local record.
- `crates/clickhousectl/src/cloud/service_query.rs:205-243` merges the replacement UUID into the existing endpoint list.
- `crates/clickhousectl/src/cloud/services.rs:1380-1419` cleans only the currently stored key during service deletion.

## Live reproduction

Validated against a disposable ClickHouse Cloud service on 2026-08-25 with PR #508 HEAD `93f7466`:

1. Run eight concurrent first-use queries. They correctly converge on one auto-provisioned key.
2. Disable that key through the management API.
3. Run a query. The CLI removes the rejected local record and asks for a rerun.
4. Rerun eight queries concurrently. They correctly converge on one replacement.
5. Inspect the endpoint and organization key inventory. Both the retired key and replacement remain; the endpoint lists both UUIDs.
6. Force-delete the service. The replacement is deleted, but the retired key still exists in the organization.

The retired key was deleted manually after validation, and all disposable resources were removed.

## Expected behavior

Retain enough ownership metadata to remove the exact retired key and endpoint binding safely. Cleanup can be best effort, but failures must preserve an actionable recovery record instead of discarding the only resource ID.

## Acceptance criteria

- Successful repair does not leave the retired owned key in the organization or endpoint binding.
- Failed key or endpoint cleanup retains the exact organization, service, and key resource IDs for retry.
- Repeated repairs do not grow `openApiKeys` or the organization key inventory.
- A locally corrupted secret cannot turn an enabled owned key into an untracked credential.
- Service deletion cleans current and pending-retirement owned keys.
- Concurrent repair preserves the winning replacement and does not delete unrelated or user-managed keys.
- Subprocess tests cover successful cleanup, partial cleanup failures, repeated repair, and service deletion.
[...]

Issue #450, opened 2026-08-22
Add privacy-safe runtime failure classification to telemetry


sdairs:
## Problem

Anonymous telemetry classifies every non-auth `cloud service query` runtime failure as `outcome=error, exit_code=1`. It cannot distinguish SQL/input errors, service resolution, key creation, endpoint binding, propagation, rate limits, Query API responses, timeouts, or output streaming.

This blocked root-cause analysis of 6,444 query exit-1 events. A live concurrency reproduction on HEAD `93b3dd7` produced successful queries, Query API 401, and `TOO_MANY_REQUESTS` during one provisioning burst, but current telemetry cannot distinguish those paths.

This complements #234, which reviews process exit codes. Telemetry needs bounded diagnostic categories even if the public exit-code taxonomy remains coarse.

## Proposed shape

Add definition-owned, closed categories such as:

- `failure_stage`: `sql_input`, `org_resolution`, `service_resolution`, `query_request`, `key_create`, `endpoint_get`, `endpoint_upsert`, `response_stream`
- `failure_kind`: `io`, `transport`, `http_4xx`, `http_5xx`, `sql_error`, `service_stopped`, `timeout`, `rate_limited`, `other`
- HTTP status class or bounded status, retry-count bucket, provisioning state, and duration bucket

## Privacy boundary

Never send SQL, database or format values, file paths, service or organization identifiers, API response text, credentials, query hashes, user IDs, or stable session/retry identifiers. Categories must be set only at code-owned boundaries, never derived from raw messages.

## Acceptance criteria

- Runtime errors expose stable closed categories at their owning boundary.
- Telemetry serializes only allowlisted categories and buckets.
- Secret-looking fixture values and raw errors never appear in payloads.
- Tests cover representative query stages, omission rules, and exact wire fields.
- Dashboard denominators continue separating parse, auth, completed, and censored/handoff outcomes.

sdairs:
The same classification gap is now confirmed for `local client`, not only Cloud query.

All pre-handoff failures currently collapse to `outcome=error, exit_code=1`, including `ServerNotFound`, `ServerNotRunning`, corrupt/unreadable selected metadata, `NoDefaultVersion`, `VersionNotFound`, and `Error::Exec`. The payload has no runtime cause or stage.

[...]

PR #656, opened 2026-09-02
Classify a rejected stored Query API key instead of treating it as stale

### LINKED ISSUE #528: Do not auto-replace administratively disabled Query API keys
## Problem

PR #508 treats any repeated non-SQL Query API 401/403 from a stored service key as a stale local credential. It removes the local record, and a rerun automatically creates and binds a new `sql_console_admin` key.

A 401/403 does not identify why the credential was rejected. It can also mean that an administrator intentionally disabled the key, removed it from the endpoint, restricted its IP allowlist, or otherwise revoked its access. Automatic replacement can therefore undo an intentional access-control change.

Relevant code on PR #508:

- `crates/clickhousectl/src/cloud/services.rs:1780-1787` classifies all repeated non-SQL 401/403 responses as rejected stored keys.
- `crates/clickhousectl/src/cloud/services.rs:1790-1815` removes the local record.
- `crates/clickhousectl/src/cloud/service_query.rs:87-101` creates an enabled replacement with `0.0.0.0/0`.
- `crates/clickhousectl/src/cloud/service_query.rs:238-243` binds it with `sql_console_admin` and `allowedOrigins=*`.

## Live reproduction

Validated against a disposable ClickHouse Cloud service on 2026-08-25 with PR #508 HEAD `93f7466`:

1. Let `cloud service query` provision and store a service Query API key.
2. Disable that exact key through `cloud key update <id> --state disabled`.
3. Run the query again. The Query API returns 401; after confirmation, the CLI removes the local service record and reports it as stale.
4. Rerun the same query. The CLI creates an enabled replacement, grants `sql_console_admin`, binds it to the endpoint, and the query succeeds.

The test service and both keys were removed after validation.

## Expected behavior

Before destructive repair, use the stored organization and API-key resource ID to distinguish a missing key or mismatched local secret from an administratively disabled, expired, IP-restricted, or unbound key. Intentional revocation should produce an actionable error and require an explicit repair decision rather than silently restoring access on rerun.

## Acceptance criteria

- An administratively disabled key is not automatically replaced or rebound.
- An expired key is classified deliberately and has documented repair behavior.
- IP allowlist rejection and intentional endpoint unbinding do not trigger silent admin-key replacement.
- A deleted key or genuinely stale local secret still has a safe service-scoped recovery path.
- Ambiguous control-plane lookup failures preserve local ownership metadata and do not mutate endpoint access.
- Human and JSON errors identify the reason and explicit recovery action without exposing secrets.
- Subprocess tests cover disabled, expired, deleted, unbound, IP-restricted, and ambiguous lookup states.
- Add focused live Cloud integration coverage for disabled and expired keys.

Related: #454, #527, and PR #508.

### LINKED ISSUE #659: service query: a deleted stored Query API key leaves its UUID in the endpoint binding, so re-provisioning fails with a 400
### Summary

The documented recovery path for a deleted stored key does not work end to end. When the
management API reports the stored key as gone (`GET /keys/{id}` → 404), the `query_key_deleted`
verdict (#656) removes the local record and tells the user to rerun the query so it provisions a
new key. But the deleted key's UUID is still listed in the service's `serviceQueryEndpoint`
`openApiKeys` — nothing removes it — and first-use provisioning merges the existing list into its
upsert. The API rejects that upsert with

```
400 BAD_REQUEST: OpenAPI key <deleted-uuid> does not belong to the organization <org>
```

That is the same status and message shape as the key-propagation transient #658 retries on, so
after #658 the failure now also costs the full 30 s propagation window before the CLI rolls back
(deletes the new key, leaves no record) and exits 1. The user is stuck: nothing in the CLI's
output says the endpoint still references a key that no longer exists, and every rerun repeats
the 30 s wait and the same error.

### Live reproduction

Live stress run of `cloud_service_query_key_repair_survives_key_propagation_delay` on 2026-09-02
against org `CLICKHOUSECTL_CLI` (aws/eu-west-1), binary built from `fix/658-key-propagation-retry`:

1. Service provisioned; first query provisioned a key; 20 repairs succeeded.
2. The test deleted the current key via `DELETE /keys/{id}` (nothing else).
3. `cloud service query --json` → exit 1, code `query_key_deleted`, local record removed (as designed).
4. `cloud service query` → `Provisioning Query API endpoint + key...`, then
   `Waiting for the new API key to become visible to the Query API endpoint...` for 30 s, then
   `Error: BAD_REQUEST: OpenAPI key d52d90f2-4a6d-48fc-84fb-7930fa6559ca does not belong to the organization 5fae43a3-…`.
   The rollback deleted the new key; the key inventory was 34 before and 34 after the run.

The harness redacts IDs, so the UUID was not cross-checked against the deleted key's; the
persistence of the 400 across the whole window, the clean inventory afterwards, and the fact
that nothing unbinds a deleted key make the dangling binding the only consistent explanation.
The stress test added for #658 unbinds the key from the endpoint before deleting it (the way an administrator cleaning up fully would), with a comment pointing here, so the suite exercises first-use provisioning rather than this bug.

### Why it was not fixed on the #658 branch

The fix changes the semantics of the `query_key_deleted` verdict, which PR #656 (open, same
stack) deliberately pinned as making no control-plane write. That is a decision for that PR's
reviewers, not something to slip into #658.

### Proposed fix

Option A (recommended): when the stored key is confirmed deleted, unbind its exact UUID from the
endpoint before forgetting the record. `unbind_query_endpoint` in `service_query.rs` already does
exactly this (GET, drop the one UUID, PUT) and is used by the provisioning rollback. Removing a
reference to a key that provably no longer exists undoes no administrator decision, and the ID
is the CLI's own stored `api_key_id`, so it is the "exact owned ID" discipline of #527, not a
scan. If the unbind fails, keep the record (so the ID is not lost) and report the rejection as
unverifiable, pointing at `repair-query-key`, which already handles this case: its single upsert
removes every retired ID and `delete_api_key_if_exists` treats the 404 as deleted.

Option B: keep the verdict read-only and instead have it point at `repair-query-key` rather than
removing the record. Zero code in the repair path; but it changes #656's "the next query
re-provisions" story and turns a read-only verdict into a second command for the user.

Either way: update the README's "Key deleted" bullet and the JSON-shape text in #656, extend
`a_deleted_stored_query_key_removes_only_its_own_local_record` (which currently asserts the
control plane saw only GETs), and add a case where the unbind fails.

### Acceptance

- After an administrator deletes the stored key without touching the endpoint, one
  `cloud service query` reports `query_key_deleted`, and the next one provisions a new key and
  succeeds, with the endpoint binding exactly the new key plus any unrelated keys.
- The stress test's provisioning rounds can drop their explicit `unbind_key` step.

## Related

- #656 introduced the `query_key_deleted` verdict and pinned it as read-only.
- #658 added the propagation retry that now makes this failure cost the full 30 s window.
- #527 established the exact-owned-ID cleanup discipline.

sdairs:
## What

When `cloud service query` runs with a stored per-service key and the Query API answers 401/403, the CLI now reads the key's management record (by the stored organization and management key ID) and, for an enabled key, the service's Query API endpoint binding, and classifies the rejection before anything is touched:

- Key deleted (management API 404): the stored secret can never work again, but the record is kept and nothing is changed, like every other verdict. Deleting a key does not remove its UUID from the endpoint's `openApiKeys`, so the binding is stale too; the error says so and names `repair-query-key`, which replaces the key and drops that binding in its single upsert and treats the already-deleted key as retired. Rerunning the query fails identically until then. (An earlier revision of this PR removed the local record so the next query would re-provision; live testing for #658 showed that re-provisioning fails on the stale binding, see #659, so the verdict is now read-only and repair is the one recovery path.)
- Key disabled, expired, or not bound to the endpoint (or the service has no endpoint): the record is kept, nothing is created or rebound, and the error names the key ID, the reason, and the way forward (`cloud key update <id> --state enabled` for a disabled key; `repair-query-key` to replace any of them deliberately).
- Key enabled, unexpired and bound, yet still rejected: the error prints the key's IP access list (CIDRs only), names the local-secret-mismatch possibility, and points at `cloud key update --ip-allow` or `repair-query-key`. Nothing is changed.
- Lookup failed (transport, 5xx, management credentials that cannot read keys) or a legacy record without a management key ID: the rejection is reported as unverifiable, nothing is changed, and the error says how to retry.

Each verdict carries a stable code in the #644 JSON error shape (`query_key_deleted`, `query_key_disabled`, `query_key_expired`, `query_key_unbound`, `query_key_rejected`, `query_key_unverified`), plus `api_key_id` and, for the allowlist case, `ip_access_list`. Human and JSON messages never contain the stored secret. Telemetry gains a `key_get` stage; every rewrite carries `failure` across (#450): the Query API's own 401/403 for a classified verdict, the lookup failure's own class for an unverifiable one.

The explicit `repair-query-key` command is unchanged and remains the only path that replaces a disabled, expired, unbound or IP-restricted key. It already deletes the superseded key and records a failed deletion in `pending_cleanup_api_key_ids`, so no additional `retired_key_ids` field was added: that field already is the retired-key record #527 needs, and adding a second one would duplicate it.

The `service query` after-help and the README's Query API credentials section describe the new behaviour.

## Why

A Query API 401/403 for a stored key does not say why the key was rejected. The local secret may be stale, but an administrator may equally have disabled the key, let it expire, unbound it from the endpoint, or narrowed its IP access list. A path that reads every rejection as a stale credential lets a rerun undo those decisions with a fresh `sql_console_admin` key. The management record can tell the cases apart, so the CLI asks it first, only to say why and name the way forward. No verdict changes anything, locally or in the organization; the user or agent decides to run `repair-query-key`.

## Behaviour after the fix

- `cloud service query` with a disabled stored key exits 1 with `management API key <id> is disabled ... the stored credential was kept and no replacement was created`, and `.clickhouse/credentials.json` is unchanged.
- The same query with a key the organization no longer has exits 1 with `management API key <id> no longer exists in organization <org> ... It was kept and no replacement was created; the key's UUID is also still listed on the service's Query API endpoint`, names the repair command, and leaves `.clickhouse/credentials.json` unchanged.
- `--json` emits one object on stderr with the code, message, `api_key_id`, and a recovery `command` where one is safe to suggest; the unverifiable verdict pushes no command.

## Tests

- Unit tests for the two-stage classifier (deleted, disabled, expired at or before now, active, absent fields, endpoint missing, key unbound, key bound, absent `openApiKeys`), for every verdict's code, message, failure classification and exit kind, for the unverifiable and legacy errors, for the guarded compare-and-remove in `credentials.rs`, for the closed `CloudErrorCode` vocabulary, and for the `key_get` stage vocabulary.
- Subprocess plus wiremock tests in `crates/clickhousectl/tests/cli_request_shape_test.rs` for disabled (401 and 403), expired, deleted (record byte-for-byte unchanged, only GET requests), unbound (endpoint lists another key, and no endpoint), enabled and bound but still rejected, unverifiable (key GET 500, endpoint GET 503), legacy record, and the JSON shape of the disabled, rejected, deleted and unverifiable codes. Each asserts the exit code, the stderr reason and recovery text, that the control plane saw only GET requests, and that the credentials file is byte-for-byte unchanged except in the deleted case, where only that service's record is removed.
- Telemetry subprocess test in `crates/clickhousectl/tests/telemetry_test.rs`: a failed key lookup names the `key_get` stage with the lookup's class and status, under `stored_key`, and neither the stored key pair, the SQL, nor the key ID reaches the payload.
- Live Cloud coverage in `crates/clickhouse-cloud-api/tests/service_query_key_cli_test.rs`, wired into the `service` suite of `.github/workflows/cloud-integration.yml` and mapped in `scripts/classify-cloud-integration.py`. It was run three times against disposable services in the `CLICKHOUSECTL_CLI` org (aws/eu-west-1), each created and deleted by the test. Run 1 failed on the test's own assertion because the harness ran under a coding agent and the CLI auto-selected JSON output; its cleanup deleted the service but the key provisioned by the first query was not yet registered, so that one key (`1458600e-eec1-4b5d-8942-97de71554d93`, named after the disposable service) was deleted by hand and the test now registers cleanup before asserting. Run 2 verified the disabled and expired verdicts live and then hit a transient `400 BAD_REQUEST: OpenAPI key <id> does not belong to the organization` from the endpoint upsert inside `repair-query-key`; the rollback and cleanup completed. Run 3 passed end to end: `query_key_disabled` (about 30s after disabling), `query_key_expired` (the API refuses a past `expireAt`, so a near-future one is set and waited for), local record untouched, no key created, then `repair-query-key` deleted the expired key, bound only the replacement, and a query succeeded. After each run the org held exactly its 34 pre-existing keys and its one pre-existing service. The transient repair 400 is worth a follow-up issue; the test prints diagnostics and retries once so a recurrence is visible without failing the suite.
- `cargo fmt --all --check`
- `cargo test -p clickhouse-cloud-api --lib --test spec_coverage_test`
- `CLICKHOUSE_CLOUD_TEST_CLICKHOUSECTL_BIN=target/debug/clickhousectl cargo test -p clickhouse-cloud-api --test service_query_key_cli_test -- --ignored --nocapture`

## Stacking

Part of a stacked PR chain: based on `feat/567-reverse-private-endpoint`, not `main`.

Fixes #528
Fixes #659


iskakaushik (approved):
LGTM

PR #657, opened 2026-09-02
Clean up retired Query API keys during stale-key repair

### LINKED ISSUE #527: Clean up retired Query API keys during stale-key repair

sdairs:
## What

`cloud service repair-query-key` now retires the key it replaces, and a retirement that could not be completed is retried instead of being left to the user.

- The repair's single endpoint upsert removes every retired owned key from `openApiKeys` (the superseded key and any earlier retirement still pending) and binds the replacement. Nothing else about the endpoint changes.
- After the replacement is committed to `.clickhouse/credentials.json` with every retired ID listed under `pending_cleanup_api_key_ids`, the retired keys are deleted best effort. A failed delete no longer fails the repair: the ID stays on the record, the command exits 0, prints a warning naming the key, the failure, where the ID is kept and how it is retried, and reports the IDs under `pendingCleanupApiKeyIds` (and the deleted ones under `deletedApiKeyIds`) in `--json` mode.
- The next `cloud service query` for that service retries any pending retirement before running the query: quietly on success, with the same warning on failure, and never at the expense of the query. The retry re-reads the record under the provisioning lock, deletes only IDs that are on the record, never the active key, and edits the pending list only while the record still names the active key it was read with, so a concurrent repair's replacement is safe.
- The `cleanup_completed` cleanup-only mode of `repair-query-key` is gone. Repair now always means "replace", and it retires pending keys as part of that. This closes a trap in PR #508 where a persistently failing delete blocked key rotation entirely (the repair errored out before replacing anything) and where a rejected key plus a pending retirement took two repair runs.
- `cloud service delete` attempts every owned key (current and pending) even when one fails, then exits non-zero naming exactly the keys that remain and how to delete them, keeping the local record so no ID is lost. It already deleted pending keys before this PR; only the failure handling changed.
- New telemetry stage `key_delete` for the deletion boundaries (repair, retry, service delete), with the closed-vocabulary test extended and the failure carried across every rewritten message.

What was already there from PR #508 and is kept as is: `pending_cleanup_api_key_ids` on the local record, the repair upsert removing the superseded UUID, `cloud service delete` deleting pending and current keys, the provisioning lock and compare-and-remove discipline, and the refusal to repair a record that lists its own active key as pending. No new record field was added.

## Why

Issue #527: after PR #508, a failed delete of the superseded key made the whole repair exit 1 and told the user to rerun `repair-query-key`, which then ran in a cleanup-only mode. Repeated repairs could grow the organization's key inventory, and the only path back was a manual rerun. Cleanup must be best effort but must never discard the only copy of a resource ID, and it must only ever touch keys the CLI itself created, identified by the exact stored IDs.

The retry lives on the query path rather than only in the repair command because it is both cheaper and less surprising there: the pending list is empty on every ordinary run, so the retry costs nothing unless a retirement is actually outstanding, and the user does not have to run a write command a second time to finish a cleanup they already asked for. Keeping `repair-query-key` single-purpose (replace, and retire what it replaces) avoids rotating a healthy key just to finish a delete.

The "key deleted" verdict on the query path makes no write at all (#528 now keeps the record for every verdict): the pending-retirement retry still runs before the classification, and the record, with its emptied or unchanged pending list, is kept for `repair-query-key`, which retires the deleted key together with anything pending. Only the repair path unbinds keys.

## Behaviour after the fix

- `repair-query-key` succeeds, deletes the superseded key, and the endpoint binds only the replacement plus any unrelated keys. Repeated repairs leave exactly one owned key and nothing pending.
- If the delete fails: exit 0, stderr `Warning: the query key for service <id> is active, but the superseded API key could not be deleted (<key-id>: <error>). The exact key IDs remain in .clickhouse/credentials.json under service_query_keys.<id>.pending_cleanup_api_key_ids and deletion is retried automatically by the next clickhousectl cloud service query ... To delete a key now, run clickhousectl cloud key delete <key-id> --org-id <org>`, and `pendingCleanupApiKeyIds` in the JSON result.
- The next `cloud service query` deletes the pending key silently and clears the list; if it fails again the same warning is printed and the query still runs.
- `cloud service delete` deletes the current key and every pending key; on a partial failure it exits 1 naming only the keys that remain.

## Tests

- Unit: endpoint request unbinds every retired key and keeps unrelated ones; warning text names every key, the record path, the retry and the manual command; `RetirementOutcome` keeps only failed IDs pending; the repair result serializes the ID lists and omits the prose warning; credentials round trip with a pending list, a legacy record without the field reads as empty, and the compare-and-update helper edits only the record that still names the active key.
- Subprocess and wiremock (`tests/cli_request_shape_test.rs`): successful repair deletes the superseded key and the endpoint PUT omits it; DELETE 500 during repair keeps the ID pending, prints the warning, exits 0 with the IDs in JSON; a repair with a pending retirement unbinds and deletes both retired keys in one upsert and leaves nothing pending; two repairs in a row never grow the binding or the list; a query retries pending retirements quietly and clears them; a failed retry warns, keeps only the failed ID, and the query still succeeds; the retry never deletes the active key; a query with nothing pending makes no control-plane writes; a deleted key with a pending retirement keeps its record whether the retry fails or succeeds, and points at repair; `service delete` attempts every owned key and reports only the failed one, keeping the record.
- Telemetry (`tests/telemetry_test.rs`): a failed owned-key delete during `service delete` reports `failure_stage: key_delete`, `http_5xx`, `502`, with no key ID, secret or path in the payload.
- Live (`crates/clickhouse-cloud-api/tests/service_query_key_cli_test.rs`, `service` suite, already classifier-mapped): new `cloud_service_query_key_repairs_retire_keys_and_service_delete_cleans_up` provisions a disposable service, lets the CLI provision a key, runs two repairs asserting each superseded key is gone and only the replacement is bound, injects a test-owned key as a pending retirement and asserts the next query deletes it quietly, injects another and asserts `cloud service delete` removes the service, the current key and the pending key and drops the local record. The shared provisioning and repair-with-one-retry steps were extracted into helpers used by both live tests.
- Live run on 2026-09-02 against org `CLICKHOUSECTL_CLI` (aws, eu-west-1): passed in 262s. Created and deleted: service `clickhousectl-it-local-527-qr`, its query endpoint, three CLI-provisioned keys, two test-owned keys (`clickhousectl-query-clickhousectl-it-local-527-qr-retired-1/2`). Key inventory was 34 before and 34 after; no `local-527` key or service remains; the pre-existing `radar-langfuse-export-test` service and its key were not touched. The second repair hit the same transient #528 saw once (`400 BAD_REQUEST: OpenAPI key <new-id> does not belong to the organization` from the endpoint upsert right after key creation); the CLI rolled back (binding restored, new key deleted, record unchanged) and the harness's single retry passed. Two sightings in four live runs suggests a follow-up: retry the repair upsert once on that specific 400 shortly after key creation.

Commands:

- `cargo fmt --all --check`
- `cargo test -p clickhousectl` (971 unit, 212 request-shape, 49 telemetry, all local suites: 0 failures)
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` (90 tests)
- `CLICKHOUSE_CLOUD_TEST_CLICKHOUSECTL_BIN=target/debug/clickhousectl cargo test -p clickhouse-cloud-api --test service_query_key_cli_test -- --ignored --nocapture --exact cloud_service_query_key_repairs_retire_keys_and_service_delete_cleans_up`

## Stacking

Part of a stacked PR chain: based on `fix/528-disabled-key-no-auto-replace`, not `main`.

Fixes #527


iskakaushik (approved):
LGTM

PR #660, opened 2026-09-02
Wait out Query API key propagation and verify the repaired key

### LINKED ISSUE #658: repair-query-key: transient 400 'OpenAPI key <id> does not belong to the organization' from endpoint upsert right after key creation
## Summary

`clickhousectl cloud service repair-query-key` intermittently fails with

```
400 BAD_REQUEST: OpenAPI key <new-key-id> does not belong to the organization
```

from the query endpoint upsert (`PUT /v1/organizations/{org}/services/{service}/serviceQueryEndpoint`) issued immediately after the replacement key was created with `POST /v1/organizations/{org}/keys`. The key does belong to the organization: an identical re-run seconds later succeeds, and `GET /keys/{id}` returns it. This looks like a propagation delay between key creation and the endpoint service's view of the organization's keys.

Observed in 2 of 4 live runs on 2026-09-02 (org `CLICKHOUSECTL_CLI`, aws/eu-west-1) while running the CLI-driven live tests added in #656 and #657 (`crates/clickhouse-cloud-api/tests/service_query_key_cli_test.rs`). Both times the failing request was the very first call after key creation.

## Current behaviour

The repair path in `crates/clickhousectl/src/cloud/service_query.rs` (`repair_service_query_key`) does the right thing on failure: `fail_after_repair_binding` restores the original binding, deletes the just-created key, leaves the local record untouched, and exits 1 with the API error. Nothing is leaked and nothing is half-applied. But the user is told the repair failed for a reason that is not true, and has to rerun by hand.

The live test currently papers over this with a single 20 s retry inside the harness (`repair_with_one_retry`), printing diagnostics so recurrences stay visible in CI logs. That retry belongs in the CLI, not the test.

## Proposal

1. In `repair_service_query_key` (and the first-provision path that also creates a key and then upserts the endpoint, if it can hit the same race), retry the endpoint upsert a bounded number of times with a short backoff when the API answers 400 and the key was created by this same invocation. Suggested: up to 3 attempts over roughly 10 s. Do not retry other 400s and do not retry once the upsert has succeeded.
2. Recognise the condition structurally, not by sniffing message text. Per CLAUDE.md #450 the failure classification must come from the library `Error` variant and allowlisted status only. If the library needs a way to tell this 400 apart, add a typed variant or typed detail in `clickhouse-cloud-api` rather than matching on the string in the CLI.
3. Record each retry with `failure::note_retry()` so telemetry shows the run waited on propagation, the same way the `QUERY_ENDPOINT_READINESS` loop does.
4. Once the CLI retries, remove `repair_with_one_retry` from the live test so the suite fails if the condition escapes the CLI's own retry.

## Acceptance

- A wiremock subprocess test in `crates/clickhousectl/tests/cli_request_shape_test.rs`: endpoint upsert returns the 400 on the first attempt and 200 on the second; repair exits 0, exactly one key was created, the old key was retired, and no rollback DELETE of the new key was sent.
- A control test: a 400 with a different body is not retried and still rolls back as today.
- The live test no longer carries its own retry.

## Related

- #656 classifies rejected stored keys before repair.
- #657 retires superseded keys on repair.
- #508 introduced the repair path and its rollback.

sdairs:
Closes #658

## What

`cloud service repair-query-key` and first-use Query API key provisioning both create an API key and immediately upsert the service's query endpoint binding. The control plane sometimes answers that upsert with `400 BAD_REQUEST: OpenAPI key <new-id> does not belong to the organization` because the key has not yet propagated to the endpoint service (#658, seen in 2 of 4 live runs while landing #656/#657). This PR makes both paths wait that out, verifies a repaired key before the command reports success, and closes two smaller gaps the work surfaced.

- **Shared propagation wait.** One helper (`upsert_until_key_propagates` behind `CloudClient::bind_created_query_key` in `service_query.rs`) is used by both `bind_query_endpoint` and `repair_service_query_key`. It retries only the endpoint upsert, with the same body, on the library's typed `Error::Api { status: 400 }`, with exponential backoff (1 s → 8 s) for up to 30 s, prints `Waiting for the new API key to become visible to the Query API endpoint...` once on stderr, and records each retry with `failure::note_retry()`. No message text is inspected (#450): the 400's body embeds a variable key ID, so a typed variant would have to regex prose. Every 400 inside the window is therefore retried; a genuine 400 costs the window and then fails with exactly the same rollback as before, reporting the API's own last error. Non-400 failures are not retried.
- **Repair verifies the replacement.** The live stress test observed the case the README already warned about for first use: 1 of 26 queries issued immediately after a successful repair was rejected while the binding converged, and the #656 classifier would have called that `query_key_rejected`. `repair-query-key` now probes the endpoint with the new key (`SELECT 1`, the existing readiness loop) on a `running` service before reporting. The result gains a `verification` field: `verified`, `skipped` (service not `running` or unreadable, no Query API host configured for the base URL, record replaced by a concurrent repair, probe answered idle/stopped) or `failed` (probe error unrelated to readiness). The result is always printed, in both modes, before any error, so a `--json` consumer never loses the IDs of a repair that happened. Only a key the Query API kept rejecting for the whole 120 s window exits 1, with code `query_key_repair_unverified`, the key ID, and a message that says the repair stands and to run `cloud service query` (which classifies the rejection) rather than rerun repair, which would rotate a key that may only be slow.
- **A rolled-back key whose deletion fails is recorded.** When a failed repair restores the binding but the DELETE of the just-created key fails, the ID is appended to the existing record's `pending_cleanup_api_key_ids` (new guarded compare-and-append in `credentials.rs`, never the active key) so the next query's retry deletes it, and the error says so. Previously the ID lived only in the error message.
- **Library:** `Client::configured_query_host()` exposes whether a Query API host is actually configured (override, env, or derivable from the base URL). The CLI uses it to skip the probe rather than send a test query to production when the control plane is a mock or unusual host.
- Notices in the provisioning and repair paths use `eprint_line`, so a closed stderr cannot panic between key creation and rollback.
- The live harness's own `repair_with_one_retry` is gone; the suite now fails if the race escapes the CLI.

## Live results (org `CLICKHOUSECTL_CLI`, aws/eu-west-1, 2026-09-02)

New live test `cloud_service_query_key_repair_survives_key_propagation_delay` runs repairs back to back, each followed immediately by a query, then cycles first-use provisioning (iterations via `CLICKHOUSE_CLOUD_TEST_REPAIR_STRESS_ITERATIONS` / `..._PROVISION_STRESS_ITERATIONS`, defaults 6 / 2).

| Run | Repairs | Waited on propagation | Query right after repair rejected | Provisioning rounds | Key inventory |
|---|---|---|---|---|---|
| measurement, retry only, N=20 | 20/20 | 1 (recovered on the 1 s retry) | 0 | n/a (#659) | 34 → 34 |
| retry only, all three tests | 6/6 | 0 | **1** (accepted ~10 s later) | 3/3 | 34 → 34 |
| retry + verification | 12/12 | 1 (1 s) | 0 | 3/3 | 34 → 34 |
| final code, all three tests | 6/6 | 0 | 0 | 3/3 | 34 → 34 |
| rebased on read-only #656, all three tests | 6/6 | 1 (7.6 s) | 0 | 2/2 deleted→repair→query, 1/1 first-use | 34 → 34, IDs identical |

Both observed propagation waits resolved on the first 1 s retry; the 30 s window is kept for headroom (earlier sightings only proved 20 s worked). Every run left the organization with its 34 pre-existing keys and only the pre-existing `radar-langfuse-export-test` service; nothing was cleaned by hand.

The measurement run also found a pre-existing bug, filed as #659: after a stored key was deleted, the #656 `query_key_deleted` verdict forgot the local record so the next query would re-provision, but the dead UUID stays in the endpoint's `openApiKeys` and the merged upsert is rejected with the same 400 for the whole window. Al's decision: that verdict is now read-only and `repair-query-key` is the recovery path (fixed in #656, this branch rebased on top). The stress test proves the flow end to end: delete the key without touching the endpoint, assert the read-only `query_key_deleted` verdict with the record and endpoint untouched, run `repair-query-key`, assert only the replacement is bound and the next query succeeds first time (`CLICKHOUSE_CLOUD_TEST_RECOVERY_STRESS_ITERATIONS`, default 2). One pure first-use provisioning round (unbind, delete, forget the record) keeps the provisioning path's propagation retry exercised (`..._FIRST_USE_STRESS_ITERATIONS`, default 1).

## Tests

- Unit (`service_query.rs`): 400-only predicate over every other status and variant; retry then success; non-400 not retried; deadline returns the last 400; backoff capped (deterministic with `start_paused`). Unit (`credentials.rs`): guarded compare-and-append. Unit (`services.rs`): `verify_repaired_query_key` against wiremock with a millisecond readiness policy: timeout → generic kind, code, command, ≥2 probes; 401→200 → verified; no configured host and record-gone guard → skipped with zero requests.
- Subprocess + wiremock (`tests/cli_request_shape_test.rs`): repair upsert 400 then 200 → exit 0, one `POST /keys`, byte-identical retried bodies, old key retired, no DELETE of the new key, notice once, `verification: verified`; a 400 that outlives the window (real 30 s) → rollback exactly as before, last 400 reported, credentials byte-identical; first-use provisioning 400 then 200; rollback DELETE failure records the ID as pending; verification outcomes: verified, state 404 → skipped, stopped/idle → skipped, unconfigured host → skipped with every request on the mock, probe 502 → failed with exit 0 and result on stdout, `Service is stopped` mid-probe → skipped, timeout → exit 1 with result on stdout and the structured error on stderr, no rollback (this one waits the real 120 s window and is `#[ignore]`d; run once by hand: pass). Every pre-existing success-path repair test now mounts a running service and a 200 probe and asserts the probe was received.
- Telemetry (`tests/telemetry_test.rs`): a repair that retried once reports `failure_stage: endpoint_upsert`, `retry_bucket: "1"`, with no key ID, secret or message text in the payload.
- Live: the two existing tests in `service_query_key_cli_test.rs` (without the harness retry) plus the new stress test, all green on the final code (367 s).

Commands: `cargo fmt --all --check`; `cargo clippy --workspace --all-targets -- -D warnings`; `cargo clippy -p clickhousectl --all-targets --no-default-features -- -D warnings`; `cargo check --workspace --all-features`; `cargo test -p clickhousectl`; `cargo test -p clickhouse-cloud-api --lib --test spec_coverage_test`; `python3 -m unittest discover -s scripts/tests -p 'test_*.py'`; `CLICKHOUSE_CLOUD_TEST_CLICKHOUSECTL_BIN=target/debug/clickhousectl cargo test -p clickhouse-cloud-api --test service_query_key_cli_test -- --ignored --nocapture`.

## Stacking

Part of a stacked PR chain: based on `fix/527-retired-key-cleanup` (#657), not `main`.

Fixes #658


iskakaushik (approved):
LGTM

Issue #606, opened 2026-08-28
clickpipe create kafka / schema-discover kafka reject no-auth brokers with implicit PLAIN error


sdairs:
Discovered during v0.4.2 post-merge stress testing.

## Repro

```
chctl cloud clickpipe create kafka <svc> ...   # no --auth flag
chctl cloud clickpipe schema-discover kafka <svc> ...
```

## Actual

Client-side failure: `PLAIN requires --username and --password` even though the user never asked for PLAIN. This blocks the legitimate "broker with no auth" case.

## Expected

Default to no auth, or only error when auth flags are partially set.

PR #669, opened 2026-09-02
Stop requiring --username/--password for IAM_ROLE Postgres ClickPipes

### LINKED ISSUE #662: `clickpipe create postgres --auth IAM_ROLE` still requires `--username`/`--password`
## Summary

`cloud clickpipe create postgres` rejects `--auth IAM_ROLE` unless basic-auth credentials are also supplied, even though IAM-role auth has no username/password:

```console
$ clickhousectl cloud clickpipe create postgres "$SERVICE_ID" \
    --name p --host h --pg-database db \
    --auth IAM_ROLE --iam-role arn:aws:iam::123:role/clickpipes \
    --table-mapping public.users:users
error: the following required arguments were not provided:
  --username <USERNAME>
  --password <PASSWORD>
```

`--iam-role` itself is already conditionally required (`required_if_eq("auth", "IAM_ROLE")`) and the reverse guard works (`--iam-role` with `--auth basic` exits 2 with a clear message). Only `--username`/`--password` are unconditionally required. Observed on the 0.4.2 binary during release smoke testing.

## Where

Two layers both force credentials:

1. Clap: `PostgresCreateArgs` declares `username` and `password` as plain required `String` args with no auth-aware conditionality — `crates/clickhousectl/src/cloud/clickpipes.rs` around lines 904–910. (`iam_role` at ~959 shows the intended pattern.)
2. Model: `ClickPipeMutatePostgresSource.credentials` is a non-`Option` `PLAIN` — `crates/clickhouse-cloud-api/src/models/clickpipes.rs` around line 2377 — so `build_postgres_request` always sends `PLAIN { username, password }` (`clickpipes.rs` ~2936–2939) and could not omit it today.

The vendored OpenAPI snapshot does not list `credentials` in `required` for `ClickPipeMutatePostgresSource`, and documents `iamRole` as "required for IAM_ROLE authentication" — i.e. credentials should be omittable for IAM_ROLE. The struct already uses exactly this workaround pattern for `iamRole` and `caCertificate` (both `Option` with explanatory comments because the server rejects the bare-String default).

## Expected behavior

- `--username`/`--password` required only for `--auth basic` (e.g. `required_unless_eq("auth", "IAM_ROLE")`).
- With `--auth IAM_ROLE`, `credentials` is omitted from the request (`Option<PLAIN>` + `skip_serializing_if`), `iamRole` is sent.
- Passing `--username`/`--password` alongside `--auth IAM_ROLE` is rejected rather than silently sent (mirrors the existing `--iam-role` + basic guard).

Caveat to verify during implementation: server acceptance of an omitted `credentials` object for `iamRole`-auth pipes could not be live-tested here (needs an RDS/Aurora source fixture). If the server turns out to require a placeholder, keep the flags required for IAM_ROLE and document why instead.

## Acceptance criteria

- IAM_ROLE create without credentials parses and sends a request with no `credentials` object (asserted in a `build_postgres_request` unit test and, if feasible, a wiremock case in `tests/cli_request_shape_test.rs`).
- Basic auth unchanged: missing credentials still a usage error.
- `--username`/`--password` + IAM_ROLE → exit 2 with a message mirroring the existing `--iam-role`/basic guard.
- Clap parsing tests for all three combinations next to the existing `postgres_iam_role_auth_requires_role_arn` tests.

sdairs:
## What

`clickhousectl cloud clickpipe create postgres ... --auth IAM_ROLE --iam-role arn:...` failed before it ever reached the API, with clap reporting "the following required arguments were not provided: --username --password". IAM role authentication has no username or password, so an RDS or Aurora Postgres pipe could not be created at all.

Two layers forced the pair, and both are fixed:

- `PostgresCreateArgs.username` and `.password` are now `Option<String>` joined to each other with clap `requires`, so the pair is all-or-nothing and neither flag is conditionally required. Passing either one with `--auth IAM_ROLE` is now rejected instead of being silently dropped, mirroring the existing guard that rejects `--iam-role` with basic auth.
- `ClickPipeMutatePostgresSource.credentials` in the API library is now `Option<PLAIN>` with `skip_serializing_if`, so `build_postgres_request` can leave the `credentials` object off the wire entirely for IAM_ROLE instead of sending an empty username and password pair.

The CLI cannot omit a field the library models as non-`Option`. `credentials` was a bare `PLAIN`, so every Postgres pipe request carried a username and password whatever the auth mode said.

The vendored OpenAPI snapshot has no `required[]` array at all on `ClickPipeMutatePostgresSource`, so the analyzer resolves requiredness from the description convention and `credentials` (which has no description) reads as required. That is why the field needs an `optionality_exemptions` entry, sitting next to the `caCertificate`, `iamRole` and `tlsHost` entries that are in exactly the same position for exactly the same reason: an empty value fails server-side validation for sources that do not use the field. The spec documents `iamRole` as "required for IAM_ROLE authentication", and the sibling `ClickPipeMutateMySQLSource` schema does carry a `required[]` array which omits `credentials`, so MySQL already models it as `Option<PLAIN>`. This brings Postgres to the same shape. The exemption was verified to be load-bearing: removing it produces `FieldOptionalityMismatch [/components/schemas/ClickPipeMutatePostgresSource/properties/credentials]`.

## Behaviour after the fix

- `--auth IAM_ROLE --iam-role arn:...` with no credentials succeeds and sends a body with `iamRole` and no `credentials` key under `source.postgres`.
- `--auth basic` (the default) still requires both `--username` and `--password`, still a clap usage error listing the missing flag.
- `--auth IAM_ROLE` with `--username`/`--password` exits 2 with "--username and --password cannot be used with --auth IAM_ROLE; use --auth basic", the same shape as the existing "--iam-role cannot be used with --auth basic; use --auth IAM_ROLE".
- Basic auth request bodies are unchanged.

Live acceptance of an omitted `credentials` object for an IAM_ROLE pipe was not tested, because it needs an RDS or Aurora fixture with an attached IAM role. The spec says the field is optional and MySQL already treats it that way. If the control plane turns out to require a placeholder object, that is a spec defect to raise upstream rather than something to model around here.

## Tests

- Library: `serialize_clickpipe_mutate_postgres_source_omits_absent_credentials` in `crates/clickhouse-cloud-api/tests/models_test.rs` asserts that `credentials: None` omits the key while `authentication` and `iamRole` still serialize, and that basic auth still sends the pair verbatim.
- Request builder: the maximal `build_postgres_request` test now uses IAM_ROLE with no credentials and asserts `credentials == None` plus `iam_role == Some(..)`; the minimal basic-auth test asserts the pair is unchanged.
- Clap: `postgres_iam_role_auth_does_not_require_username_or_password` covers IAM_ROLE parsing without credentials and the rejection of the pair; `postgres_basic_auth_still_requires_username_and_password` covers each flag being omitted under basic auth as a `MissingRequiredArgument` error.
- Help and README text pins updated in `postgres_help_documents_input_tls_and_source_requirements` and `readme_documents_postgres_tls_and_cdc_prerequisites`.
- Subprocess and wiremock in `crates/clickhousectl/tests/cli_request_shape_test.rs`: `postgres_iam_role_create_omits_the_credentials_object`, `postgres_basic_auth_create_sends_the_credentials_object`, and `postgres_credentials_with_iam_role_auth_are_rejected` (exit 2).

The credential flags are paired with clap `requires` rather than made conditionally required, matching the Kafka precedent from issue #606. `required_unless_present = "iam_role"` made `--auth IAM_ROLE` on its own demand `--username`, `--password` and `--iam-role` all at once, and following that advice then hit the "cannot be used with --auth IAM_ROLE" rejection. With the pairing, `--auth IAM_ROLE` alone names only `--iam-role`, half a pair names its missing half, and the basic-auth rule is owned by the `validate_postgres_create_args` check that was previously unreachable. Three tests cover this: `postgres_iam_role_only_error_names_the_role_flag_and_not_the_pair` asserts the IAM_ROLE-only usage error names `--iam-role` and mentions neither credential flag, `postgres_basic_auth_without_any_credentials_is_a_validation_error` asserts that omitting both flags under basic auth reaches the "--auth basic requires --username <USERNAME> and --password <PASSWORD>" message (previously uncovered), and `postgres_basic_auth_still_requires_username_and_password` keeps the half-pair cases as clap `MissingRequiredArgument` errors naming the omitted flag. `build_postgres_request` now matches exhaustively on the parsed `ClickPipeMutatePostgresSourceAuthentication` instead of comparing `--auth` as a string, so a new authentication mode in the library enum is a compile error here rather than a pipe created with no credentials.

`ClickPipePatchPostgresSource.credentials` is still a non-`Option` `PLAIN`; it has no construction site in the CLI, so it is left alone.

Verified with:

```
cargo fmt --all --check
cargo clippy --workspace --all-targets -- -D warnings
cargo clippy -p clickhousectl --all-targets --no-default-features -- -D warnings
cargo test -p clickhousectl
cargo test -p clickhouse-cloud-api --lib --test spec_coverage_test
cargo test -p clickhouse-cloud-api --test models_test
cargo check --workspace --all-features
python3 -m unittest discover -s scripts/tests -p 'test_*.py'
```

The same clap shape exists on `clickpipe create mysql`; that CLI-only follow-up is tracked as #670.

## Stacking

Part of a stacked PR chain: based on `fix/663-postgres-update-name`, not `main`.

Fixes #662

### DIFF
--- a/README.md
+++ b/README.md
@@ -1102,9 +1102,10 @@ PostgreSQL ClickPipes require at least one table mapping, from either
 (see [PostgreSQL table mappings](#postgresql-table-mappings)). Ports must be in
 `1..=65535`. `--auth IAM_ROLE` requires `--iam-role`; the CLI rejects
 `--iam-role` with basic auth rather than silently ignoring it.
-`--username` and `--password` are basic-auth only: they are required for the
-default `--auth basic`, and rejected with `--auth IAM_ROLE`, where the role ARN
-is the whole credential and no `credentials` object is sent.
+`--username` and `--password` are basic-auth only and must be given together:
+they are required for the default `--auth basic`, and rejected with
+`--auth IAM_ROLE`, where the role ARN is the whole credential and
+no `credentials` object is sent.
 `--replication-slot-name` is valid only with `--replication-mode cdc_only`.
 
 #### PostgreSQL table mappings
--- a/README.md
+++ b/README.md
@@ -948,6 +948,16 @@ clickhousectl cloud clickpipe create postgres <service-id> \
   --table-mapping "public.users:public_users" \
   --table-mapping "public.orders:public_orders"
 
+# From an RDS or Aurora PostgreSQL with IAM role authentication (CDC)
+# IAM_ROLE auth takes no --username/--password: the role ARN is the credential
+clickhousectl cloud clickpipe create postgres <service-id> \
+  --name my-rds-pg-pipe \
+  --host db.abcdefg.us-east-1.rds.amazonaws.com --pg-database mydb \
+  --postgres-type rdspostgres \
+  --auth IAM_ROLE --iam-role "$POSTGRES_IAM_ROLE_ARN" \
+  --publication-name clickpipes \
+  --table-mapping "public.users:public_users"
+
 # From PostgreSQL with a private or self-signed CA (CDC)
 clickhousectl cloud clickpipe create postgres <service-id> \
   --name my-private-pg-pipe \
@@ -1092,6 +1102,9 @@ PostgreSQL ClickPipes require at least one table mapping, from either
 (see [PostgreSQL table mappings](#postgresql-table-mappings)). Ports must be in
 `1..=65535`. `--auth IAM_ROLE` requires `--iam-role`; the CLI rejects
 `--iam-role` with basic auth rather than silently ignoring it.
+`--username` and `--password` are basic-auth only: they are required for the
+default `--auth basic`, and rejected with `--auth IAM_ROLE`, where the role ARN
+is the whole credential and no `credentials` object is sent.
 `--replication-slot-name` is valid only with `--replication-mode cdc_only`.
 
 #### PostgreSQL table mappings
--- a/crates/clickhouse-cloud-api/src/models/clickpipes.rs
+++ b/crates/clickhouse-cloud-api/src/models/clickpipes.rs
@@ -2374,7 +2374,14 @@ pub struct ClickPipeMutatePostgresSource {
     // `Option<String>` so callers can omit it.
     #[serde(rename = "caCertificate", skip_serializing_if = "Option::is_none")]
     pub ca_certificate: Option<String>,
-    pub credentials: PLAIN,
+    // credentials is the basic-auth username/password pair. The schema carries
+    // no `required[]`, so requiredness resolves from the description heuristic
+    // and the field reads as required, but IAM_ROLE authentication has no
+    // username or password: the role ARN in `iamRole` is the whole credential.
+    // Modeled as `Option<PLAIN>` so an IAM_ROLE request can omit the object
+    // entirely instead of sending an empty `{"username":"","password":""}`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub credentials: Option<PLAIN>,
     pub database: String,
     #[serde(rename = "disableTls")]
     pub disable_tls: bool,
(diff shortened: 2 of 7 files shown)


iskakaushik (approved):
LGTM

PR #671, opened 2026-09-02
Summarise PEM certificates in human output instead of dumping the body

### LINKED ISSUE #665: ClickPipe human output dumps the full `caCertificate` PEM (~1.5 KB); elide to a summary in non-JSON output
## Summary

Human-readable ClickPipe output dumps the full source CA certificate PEM (~1.5 KB) inline. Observed on the 0.4.2 binary during ClickPipe CDC smoke testing: `clickpipe list`, `get`, `stop`, `start`, and `resync` all render `source.postgres.caCertificate` as the complete PEM block in non-JSON output.

That is certificate noise in an operational view — nothing a human can act on at a glance — while the full PEM remains correctly available via `--json`.

## Expected behavior

In human output, elide the PEM body and show a summary instead, e.g. `caCertificate: <present, 1 certificate, SHA-256 fingerprint ab:cd:...>` (or simply `<present>`/`<absent>`). `--json` keeps emitting the field verbatim — `print_human` and JSON already share serialization, so this is a rendering concern only.

For reference, response models drive human output through `print_human` (`src/cloud/clickpipes.rs`), so a targeted pretty-printer or a wrapper type for the certificate field keeps `--json` untouched.

## Acceptance criteria

- Human output of `clickpipe list/get/stop/start/resync` does not print raw PEM for `caCertificate` when present; prints nothing/absent marker when it is not.
- `--json` output unchanged (full PEM still present).
- A formatting unit test pinning the elided human representation.

sdairs:
## What

`clickhousectl cloud clickpipe get` renders a Postgres pipe's `source.postgres.caCertificate` through `print_human`, which printed the whole PEM block: roughly 1.5 KB of base64 that pushed every other field of the pipe off the screen. Human output now replaces each well-formed PEM block with a one-line summary of that block:

```
source:
  postgres:
    host: db.example.com
    caCertificate: <PEM CERTIFICATE, SHA-256 fingerprint 5A:6D:67:FD:14:1B:1E:61:4A:F4:E2:7D:F1:F8:67:E2:75:85:DF:92:E3:66:31:85:75:AB:2C:C3:F4:8C:9A:D8>
    database: postgres
```

The summary is keyed on the value's format, not on the field name. `print_human` turns a JSON string scalar into text in exactly one place (`scalar_string`), so the work lives there and covers object fields, nested objects, arrays of strings and array bullets, and client certificates and private keys as well as CA certificates, with no per-field allowlist to keep in sync as the API grows new certificate fields.

Each block is summarised on its own terms, in place, and the text around the blocks is kept exactly as it was. A certificate-family block (`CERTIFICATE`, `TRUSTED CERTIFICATE`, `CERTIFICATE REQUEST`, `NEW CERTIFICATE REQUEST`, `X509 CRL`) is identified by the SHA-256 of its own base64-decoded DER as colon-separated uppercase hex, which is the form `openssl x509 -fingerprint -sha256` prints, so every block of a chain can be checked against the file the user holds. Every other label renders as `<PEM EC PRIVATE KEY, 121 bytes>` and is never fingerprinted: a stable identifier of secret material left in scrollback is not something any caller needs. A certificate whose base64 does not decode, an encrypted key with RFC 1421 headers or a truncated paste, falls back to the same byte count over that block's body. The body itself is never part of the result on any path.

When the value is nothing but blocks and whitespace, which is the common case of a single certificate or a chain, the per-block summaries are joined with commas on one line. Any other value keeps its own text and has each block replaced where it stood.

The block label is the only part of a block that reaches the summary line, so it is validated against RFC 7468 section 3's `label` production: non-empty, at most 64 characters, printable ASCII or space, and no leading or trailing space or hyphen. A frame carrying a newline, a control character, an ANSI escape, non-ASCII bytes or a kilobyte of text where a label belongs is not a block at all, and the scan steps past that `-----BEGIN ` and keeps looking rather than stopping, so one malformed frame cannot hide the blocks after it.

## Why

`--json` is untouched and keeps the certificate verbatim, so nothing is lost: an agent or script that needs the bytes reads them from JSON, and the human reading `clickpipe get` gets a screen they can scan. `cloud postgres certs get` is deliberately unchanged, since printing the PEM to stdout is that command's entire purpose and it does not go through `print_human`.

Summarising in place, rather than replacing the whole value, is what makes the behaviour honest about real inputs. A bundle with header comments before the first block (curl's `cacert.pem`), an `openssl x509 -text` dump, a byte-order mark, or prose between two blocks is common, and any of those either escaped elision entirely or lost its surrounding text under a whole-value approach. Per-block labels also mean a certificate followed by a key is described as what it is instead of claiming to be two certificates, and the byte-count fallback now counts the block it could not identify instead of the entire string.

`clickpipe list`, `stop`, `start` and `resync` render name, id and state only, so they never showed a certificate; the fix is in the one human path that did.

`sha2` is a new direct dependency of `clickhousectl` (it was already in `Cargo.lock` transitively). `base64` was already a direct dependency. No `pem` crate: the framing parse is small and the label rules are the part that needed care.

## Behaviour after the fix

- Human output for the test fixture: `caCertificate: <PEM CERTIFICATE, SHA-256 fingerprint 5A:6D:67:FD:14:1B:1E:61:4A:F4:E2:7D:F1:F8:67:E2:75:85:DF:92:E3:66:31:85:75:AB:2C:C3:F4:8C:9A:D8>`
- `--json` output: the full PEM, byte for byte, as the API returned it.
- A three-certificate chain renders as one line with three summaries joined by commas.
- A certificate followed by a private key renders as `<PEM CERTIFICATE, SHA-256 fingerprint ...>, <PEM EC PRIVATE KEY, 10 bytes>`, with no digest of the key.
- A bundle with header comments keeps the comments and replaces only the block.
- A non-PEM string, framing that never closes, an `END` label that does not match its `BEGIN`, an unlabelled frame, and a frame whose label is not a label all render exactly as before.
- README documents this in the shared JSON and output section, since it applies to every human detail view rather than only to pipes, and states that `cloud postgres certs get` still prints raw PEM on purpose. AGENTS.md notes it on the `print_human` step of "Adding a command", including that a hand-written `println!` or a `tabled` cell bypasses it.

## Tests

- Unit tests in `crates/clickhousectl/src/cloud/output.rs` over a committed self-signed EC certificate fixture: the summary matches the fingerprint `openssl x509 -noout -fingerprint -sha256` reports for it; a whitespace-separated three-certificate chain is one line with three summaries; a mixed certificate plus key bundle summarises the key by size and is asserted not to contain the SHA-256 of the key's body; header text before a block and prose between two blocks are preserved with the blocks substituted in place; a byte-order mark survives; CRLF framing is still summarised; an encrypted key with `Proc-Type` and `DEK-Info` headers falls back to a byte count with no header text in the summary; an undecodable certificate body falls back to a byte count over that block; non-PEM strings, unterminated framing, a mismatched `END` label and an unlabelled frame are left alone; the body and the framing never appear in a summary.
- Label validation tests: a label containing a newline, a label containing an ESC byte, and a 200-character label each produce no summary, and the rendered output is asserted to equal the raw string with no `<PEM` anywhere, so nothing is synthesised and nothing of the value is injected into a summary line.
- Render-level unit tests: a nested `caCertificate` renders the summary line with its siblings intact and no line contains `-----BEGIN`; certificates inside a scalar array and under an array bullet are summarised too.
- Subprocess plus wiremock tests in `crates/clickhousectl/tests/cli_request_shape_test.rs`: `clickpipe get` against a mocked Postgres pipe carrying a PEM `caCertificate` prints the summary and no `-----BEGIN` in human mode, and returns the certificate verbatim under `--json`.
- Verification commands:
  - `cargo fmt --all`

## Stacking

Part of a stacked PR chain: based on `fix/662-clickpipe-postgres-iam-role-credentials`, not `main`.

Fixes #665

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,7 +47,7 @@ Local clap definitions live in `src/local/cli.rs`. Cloud clap definitions, handl
        print_human(&data)?;
    }
    ```
-   List views stay as `tabled` tables, and short action confirmations (e.g. "Service X starting") stay as plain `println!`.
+   List views stay as `tabled` tables, and short action confirmations (e.g. "Service X starting") stay as plain `println!`. Human output through `print_human` also summarises a PEM-framed string (a certificate, a key) instead of printing its body, so a hand-written `println!` line or a `tabled` cell bypasses that and dumps the whole PEM.
 
    Every field of a library response type is `Option` (see Request and response models), so never `unwrap()`/`expect()` one. Render absence with `crate::cloud::output::or_absent` (`-`) or `ABSENT` in `tabled` cells and plain output, and have `--filter`-style predicates treat an absent field as non-matching. `print_human` and `--json` serialize the model and need no per-field work.
 8. Add `Cli::try_parse_from` coverage next to the domain command definition for the new command's body-related flags, asserting parsed values.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1021,6 +1021,7 @@ dependencies = [
  "rpassword",
  "serde",
  "serde_json",
+ "sha2 0.11.0",
  "tabled",
  "tar",
  "tempfile",
--- a/README.md
+++ b/README.md
@@ -1331,12 +1331,6 @@ clickhousectl cloud clickpipe schema-discover <service-id> pubsub \
 
 Add `--json` (or run as a coding agent) for machine-readable output.
 
-Human-readable output summarises any PEM certificate or key a pipe returns, such
-as `source.postgres.caCertificate`, as a one-line
-`<PEM: 1 CERTIFICATE block(s), SHA-256 fingerprint AB:CD:...>` (the fingerprint
-`openssl x509 -fingerprint -sha256` prints for the first block) instead of
-printing the ~1.5 KB body; `--json` still returns the certificate verbatim.
-
 ### Members
 
 Role IDs used by member, invitation, and API-key commands currently come from the ClickHouse Cloud Console or API.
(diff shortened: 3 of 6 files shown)


iskakaushik (approved):
LGTM

PR #672, opened 2026-09-02
Report a well-formed but unknown id as not found

### LINKED ISSUE #666: Well-formed but nonexistent UUIDs surface as `BAD_REQUEST: Invalid ... id` instead of a not-found message
## Summary

Looking up a well-formed but nonexistent resource ID surfaces as `Invalid <thing> id` rather than a not-found-style message:

```console
$ clickhousectl cloud postgres get 00000000-0000-0000-0000-000000000000
Error: BAD_REQUEST: Invalid Postgres service id string:"00000000-0000-0000-0000-000000000000"  (exit 1)

$ clickhousectl cloud org get 00000000-0000-0000-0000-000000000000
$ clickhousectl cloud service get 00000000-0000-0000-0000-000000000000
# same shape: BAD_REQUEST "Invalid ..." rather than "not found"
```

The server answers these with HTTP 400 (so the `BAD_REQUEST` classification is faithful), but "Invalid" reads as "you typed that wrong" — misleading for a syntactically perfect UUID that simply belongs to no resource. Users scripting existence checks (`get` → branch on not-found) cannot distinguish malformed input from absent resources, and `activity get bogus-id` is even less helpful: `Error: BAD_REQUEST length is too small: 8 < 20`.

Observed on the 0.4.2 binary in the release smoke test (org, service, postgres get paths).

## Where

The CLI relays the API's 400 body verbatim through `CloudError` (`src/cloud/client.rs` error conversion). The refinement belongs there or in the get handlers: a 400 whose body matches the API's `Invalid <resource> id <...>` pattern for a syntactically valid UUID the CLI just sent can be re-presented as not-found.

## Expected behavior

For a well-formed UUID that the API rejects with its "Invalid ... id" 400, report it as a not-found-style error, e.g. `Error: no such Postgres service: <id> (organization <org>)` — while genuinely malformed IDs (non-UUID, wrong length) keep an "invalid id" message. The failure kind for telemetry should stay faithful to the transport (the server did say 400) — this is about the user-facing message, not the classification.

## Acceptance criteria

- `postgres get`, `service get`, `org get` with a well-formed-but-nonexistent UUID produce a not-found-style message (exit 1).
- Malformed IDs still produce an invalid-input message.
- `--json` error output carries the same refined message.
- Unit test for the message mapping.

sdairs:
## What

`cloud postgres get 00000000-0000-0000-0000-000000000000` printed `Error: BAD_REQUEST: Invalid Postgres service id string:"0000..."` and exited 1. `cloud service get` and `cloud org get` behaved the same way. The API answers HTTP 400 for a syntactically valid UUID that resolves to nothing, so a missing resource was reported as a malformed request.

Every command that reads or deletes one of those resources by identifier now re-presents that failure as a not-found, keeping the server's own text as the trailing detail. The refinement is applied by `CloudClient::convert_error_for_lookup` in `src/cloud/client.rs`, and the call sites are the three get wrappers (`get_service`, `get_postgres_service`, `get_organization`), the read behind `service delete --force` (`get_service_if_exists`), and `delete_service`.

A `ResourceLookup` describes the identifiers the CLI put into the request path. It is built only through `ResourceLookup::in_org(kind, id, org_id)` and `ResourceLookup::organization(id)`, with private fields, so the kind and the organization id cannot disagree: an organization lookup is always scoped to itself, and every other kind is always scoped to the organization it was read from. `ResourceLookup::rejected_well_formed_ids` is the single implementation of the judgement, read both by `convert_error_for_lookup` and by `get_service_if_exists`.

## Why

There is no message to read here. CLAUDE.md forbids deriving a category from response prose, and the library `Error` contract says a distinguishable failure gets a variant rather than a recognisable message. The API's text is prose with the id interpolated into it, so it could never become a typed variant either.

The refinement is therefore structural and lives in the CLI. A by-identifier request has exactly one class of user-controlled input: the identifiers the CLI itself formatted into the URL path. When the API answers 400 and every one of those parses with `uuid::Uuid::parse_str`, "the identifier is invalid" is not what happened. The two structural inputs are the status the library reported and whether those ids parse. Nothing reads the response body.

That reasoning covers a delete by identifier exactly as it covers a read, which is why `delete_service` is included: it sends no body and takes no other user input. It also covers the read behind `--force`, where the API's 400 for a well-formed id is the same answer as its 404, so treating it as an absent service is what keeps `service delete --force` consistent with `service get` on the same id. Nothing else was widened, and `service_delete_error`, which still sniffs a `CONFLICT:` prefix, is tracked separately.

Conversion still runs through `convert_error_with_organization`, so the telemetry classification (#450) is the faithful `http_4xx` for status 400 and is carried across the rewrite with `..error`. Only the user-facing message changes, and no telemetry vocabulary is derived from the id or the message. Every other status is left alone, including the existing organization-scope enrichment on a bare 404.

`CloudErrorDetail` already carries a `code`, so a code fitted without contortion: `resource_not_found`, added to `CloudErrorCode` and to the existing stable-codes test. Its `command` is the matching `list` for the scope, which carries an organization id at most and nothing the message did not already name. Both the scope parenthetical in the message and that command key on the resource kind, not on the two identifiers being string-equal.

Two smaller inconsistencies the refinement exposed are fixed with it. `postgres get` was one of the pre-modularization paths that called `client.api()` directly, so it moves onto the `get_postgres_service` wrapper; to keep that move behaviour-preserving, the module's `unwrap_api` helper now delegates to `CloudClient::unwrap_response`, so every Postgres path reports an empty body the way the rest of the cloud surface does. And the repair verification's skip note reads `could not read service <id>: <error>` rather than nesting the refined message inside parentheses.

## Behaviour after the fix

A well-formed identifier no resource has, on a read:

```
Error: No such Postgres service: 00000000-0000-0000-0000-000000000000 (organization 00000000-0000-4000-8000-000000000001). The API rejected the identifier: BAD_REQUEST: Invalid Postgres service id string:"00000000-0000-0000-0000-000000000000"
```

The same identifier on a delete, with and without `--force`. With `--force` the read no longer aborts the command, so the delete is still attempted and its own answer is what the user sees:

```
Error: No such service: 00000000-0000-0000-0000-000000000000 (organization 00000000-0000-4000-8000-000000000001). The API rejected the identifier: BAD_REQUEST: Invalid service id string:"00000000-0000-0000-0000-000000000000"
```

A malformed identifier is untouched, in human and in `--json` mode, because there the API's "invalid" answer is both true and the more useful thing to say:

```
Error: BAD_REQUEST: Invalid Postgres service id string:"not-a-uuid"
```

`--json` mode carries the same message as human mode under the new stable cloud code, plus the command that lists what does exist in the same scope:

```json
{
  "error": {
    "code": "resource_not_found",
    "message": "No such organization: 00000000-0000-0000-0000-000000000000. The API rejected the identifier: BAD_REQUEST: Invalid organization id string:\"00000000-0000-0000-0000-000000000000\"",
    "command": "clickhousectl cloud org list"
  }
}
```

## Tests

- Unit tests next to the existing `convert_error_*` tests in `src/cloud/client.rs`: the not-found message for each of the three kinds with the server text appended, the nil UUID counted as well formed, the organization never naming an organization as its own scope however the lookup was built, the `failure` classification still `http_4xx` at status 400, the structured detail and its command, a malformed resource id left alone, a malformed organization id left alone even when the resource id parses, and 404, 500 and a non-API error unchanged.
- Subprocess and wiremock tests in `crates/clickhousectl/tests/cli_request_shape_test.rs` against a mocked 400 `Invalid ... id` body: exact stderr and exit 1 for human `postgres get`, `service get` and `org get` with a nil UUID; the `resource_not_found` code and command in `--json` mode for all three; the raw server message preserved for `not-a-uuid` on all three in human mode and in `--json` mode, where the failure stays prose with no `code`; `service delete --force` on a nil UUID whose read answers 400 still issuing the delete and reporting the refined message, mirroring the existing 404 test's request shape; and `service delete` without `--force`, where the delete's own 400 is refined and no read is made at all.
- The repair verification test that pinned the old skip note is updated to the new wording.
- No telemetry test: none of these paths records a `FailureStage`, so no failure fields reach a payload today and there is nothing for the existing harness to assert. Giving them a stage would be a separate change. The classification itself is pinned by the unit test.
- Verified with `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo clippy -p clickhousectl --all-targets --no-default-features -- -D warnings`, and `cargo test -p clickhousectl`. The API library is unchanged, so its suites were not affected.

## Stacking

Part of a stacked PR chain: based on `fix/665-elide-pem-in-human-output`, not `main`.

Fixes #666

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,6 +479,8 @@ clickhousectl cloud --debug service list
 
 Manage ClickHouse, Postgres, and other ClickHouse Cloud resources via the API.
 
+Any command that reads or deletes a cloud resource by identifier reports an identifier no resource has as not found. The API answers `400 BAD_REQUEST: Invalid ... id` for a syntactically valid UUID that resolves to nothing, so the CLI re-presents that as `No such service: <id> (organization <org-id>)` and keeps the server's own text as a trailing detail. In `--json` mode the failure carries the stable code `resource_not_found` and the `list` command for the same scope. An identifier that is not a well-formed UUID keeps the API's `invalid` message unchanged, because there it is both true and the more useful answer.
+
 ### Organizations
 
 ```bash
@@ -1471,8 +1473,6 @@ The schema and meanings of existing codes are stable. New optional fields or cod
 | `postgres_error` | A Postgres validation or state error; the message carries its recovery guidance |
 | `io_error` | A local filesystem, metadata, or serialization operation failed |
 | `local_error` | A redacted fallback for failures whose text cannot be rendered safely |
-
-Cloud failures use their own codes rather than this local schema. A by-id read (`cloud service get`, `cloud postgres get`, `cloud org get`) that names a well-formed UUID no resource has is reported as not found instead of as the API's `BAD_REQUEST: Invalid ... id`, keeping the server's own text as a trailing detail, and in `--json` mode it carries the stable code `resource_not_found` plus the matching `list` command for the same scope. A malformed identifier keeps the API's `invalid` message unchanged, because there that is both true and the more useful answer.
 
 ### Exit codes
 
--- a/README.md
+++ b/README.md
@@ -1472,6 +1472,8 @@ The schema and meanings of existing codes are stable. New optional fields or cod
 | `io_error` | A local filesystem, metadata, or serialization operation failed |
 | `local_error` | A redacted fallback for failures whose text cannot be rendered safely |
 
+Cloud failures use their own codes rather than this local schema. A by-id read (`cloud service get`, `cloud postgres get`, `cloud org get`) that names a well-formed UUID no resource has is reported as not found instead of as the API's `BAD_REQUEST: Invalid ... id`, keeping the server's own text as a trailing detail, and in `--json` mode it carries the stable code `resource_not_found` plus the matching `list` command for the same scope. A malformed identifier keeps the API's `invalid` message unchanged, because there that is both true and the more useful answer.
+
 ### Exit codes
 
 Usage errors and cancelled actions use distinct exit codes.
--- a/crates/clickhousectl/src/cloud/client.rs
+++ b/crates/clickhousectl/src/cloud/client.rs
@@ -373,14 +373,14 @@ impl AuthSource {
     }
 }
 
-// ── by-identifier reads (issue #666) ───────────────────────────────────────
+// ── by-identifier requests (issue #666) ───────────────────────────────────
 //
 // The API answers HTTP 400 `Invalid <thing> id string:"<id>"` for a
 // syntactically valid UUID that resolves to nothing, so `cloud postgres get
 // 00000000-0000-0000-0000-000000000000` reported a bad request rather than a
 // missing resource. The fix is structural, not textual: nothing here reads the
 // response prose (the message interpolates the id, so it could never become a
-// typed library variant either). A GET-by-identifier request has exactly one
+// typed library variant either). A by-identifier request has exactly one
 // class of user-controlled input — the identifiers the CLI itself formatted
 // into the URL path — so when every one of those is a well-formed UUID, "the
 // identifier is invalid" is not what happened.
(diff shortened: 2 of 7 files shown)


iskakaushik (approved):
LGTM

PR #674, opened 2026-09-02
Report the watchdog PID in both views of a running server

### LINKED ISSUE #664: `local server list` and `local server list --global` report different PIDs for the same server (watchdog vs server child)
## Summary

`local server list` (project-scoped) and `local server list --global` report different PIDs for the same running server. Reproduced on the 0.4.2 binary on macOS:

```console
$ clickhousectl local server start pidcheck
$ clickhousectl local server list          # project view
"pid": 95193
$ clickhousectl local server list --global
"pid": 95194

$ ps -o pid,ppid,comm -p 95193 -p 95194
  PID  PPID COMM
95193     1 clickhouse-watchdog
95194 95193 ~/.clickhouse/versions/26.9.1.217/clickhouse server
```

The project view reports the **watchdog** PID; the global view reports the **server child** PID.

## Where

- `local server start` records `child.id()` of the spawned process — the watchdog — into the project's server metadata (`start_server` in `crates/clickhousectl/src/local/mod.rs`; `ServerInfo.pid` in `crates/clickhousectl/src/local/server.rs`). Project-scoped `list` reads that metadata.
- `--global` uses process discovery (`list_all_servers_global()` → `discovery::discover_clickhouse_processes()` in `crates/clickhousectl/src/local/server.rs` ~line 1032), which matches the actual `clickhouse server` child process.

## Why it matters

- The two views disagree on a field a user acts on directly. `kill <pid-from-global-list>` kills the server child, and the watchdog restarts it — the opposite of the user's intent. The watchdog PID from the project view is the one whose termination actually stops the server.
- `stop` itself works regardless (it signals the stored PID), so this is a reporting inconsistency, but any user or script operating on the reported PID gets different semantics depending on which list they read.

## Expected behavior

Both views report the same PID for the same server. The stored/watchdog PID is the natural choice: it is what `stop` signals, it is what start recorded, and terminating it stops the server for good. (Alternative: report both, e.g. `pid` + `watchdog_pid`/`server_pid`, if the discovery view needs the child for its own matching.)

## Acceptance criteria

- After `local server start`, project `list` and `list --global` show identical `pid` values for that server (test with a real started server or a discovery fixture).
- `--global` discovery still matches servers started from other projects.

sdairs:
## What

`local server list` and `local server list --global` reported different PIDs for the same server, and stopping the one the CLI records could leave the other process running. Process discovery now resolves each discovered server's parent and reports the parent's PID when that parent is the ClickHouse watchdog, and the stop path terminates the watchdog together with the server it supervises.

- `crates/clickhousectl/src/local/discovery.rs`: after `pgrep -x clickhouse` finds the server PIDs, the cwd filter runs first, then `scan_process_rows` reads `pid ppid args` for the surviving PIDs in one batched `ps` call and their distinct parents in a second one. `DiscoveredProcess.pid` becomes the parent PID when that parent is a watchdog. Those same rows supply each server's command line, so the whole scan is two flat subprocess calls with no per-PID lookup, and server metadata (project root, name, ports, version) still comes from the server process, whose command line is the one that carries the binary path and the port flags.
- Parsing and classification stay pure functions over `ps` output text: `parse_ps_process_rows`, `is_watchdog_command` and `resolve_watchdog_pids`.
- `crates/clickhousectl/src/local/server.rs`: `kill_process` sends SIGTERM, waits 2.5 s, and when the process is still alive reads its direct children with one `pgrep -P` call, SIGKILLs the parent, SIGKILLs those children, and returns success only once every one of them is gone. Otherwise it names the survivor in the error.
- No change to `server start`, to the stored metadata, or to the JSON field set.

## Why

`start` records `child.id()`, and the process the CLI spawns is the watchdog, so the project-scoped view was already reporting the watchdog. Discovery went the other way: the watchdog rewrites its `argv[0]` to `clickhouse-watchdog`, so `pgrep -x clickhouse` matches only the server child it supervises. The watchdog owns the lifetime of the pair, forwarding SIGHUP, SIGINT, SIGQUIT and SIGTERM to the server and exiting with it, so it is the watchdog that has to be signalled. Restarting the server is opt-in through `CLICKHOUSE_WATCHDOG_RESTART`, so the reason to prefer the watchdog is ownership of the pair, not a restart the earlier rationale claimed. `--global` was handing users a PID with the opposite meaning to the one `stop` uses.

Every consumer of a discovered PID wanted the supervising PID:

- `list_all_servers_global` (the `--global` list) now agrees with the project-scoped list.
- `stop_server_global` and `stop_all_servers_global` call `kill_server_by_pid` on the discovered PID. They previously signalled the supervised child while its watchdog stayed alive; they now signal the watchdog.
- `recover_from_discovered_locked` (orphaned-server recovery) writes the discovered PID into the metadata file, so a recovered server carries the same PID `start` would have written, and a later `stop` terminates it.
- `select_version_users` (the cross-project `local remove` guard) de-duplicates metadata entries against discovered entries. `--force` removal reaches `ensure_stopped_by_pid`, which now signals the watchdog.
- Port selection uses `TcpListener` binding, not PIDs, so it is unaffected.

Reporting the watchdog exposed a second, worse problem in the shared kill helper. The watchdog forwards SIGTERM, but a SIGKILL cannot be caught, so escalating against a watchdog that ignores SIGTERM (which is exactly what `CLICKHOUSE_WATCHDOG_NO_FORWARD=1` produces) killed the watchdog and left the server running, reparented to init, still holding the ports and the data directory, while `stop`, `stop --global`, `stop-all --global` and `remove --force` all reported success. The escalation branch therefore reads the children before the parent dies, because once the parent is gone the kernel has reparented them and the relation is lost, and it verifies the whole set. A process group kill is not an alternative: `Command::spawn` leaves the watchdog in clickhousectl's own process group, so a group signal would hit the CLI itself. Putting the fix in `kill_process` rather than threading a second PID out of discovery means the project-scoped `stop`, which works from metadata and never sees a discovery row, is covered by the same code, and it is the design that can be driven end to end by a test.

`args` is used rather than `comm` because on Linux `comm` is the thread name the watchdog sets with `prctl`, which has changed between ClickHouse versions, while the rewritten `argv[0]` that `args` reports is identical on Linux and macOS. The accepted name is any prefix of `clickhouse-watchdog` longer than `clickhouse` itself, because the rewrite is truncated to the length of the spawn path and the only bar the name has to clear is the server's own basename. That direction of the comparison also rejects `clickhouse-client` and `clickhouse-watchdog-wrapper`.

Cost stays flat: two `ps` calls whatever the number of servers on the machine, the same reason `get_process_cwds` batches `lsof`, plus one `pgrep -P` call only on the escalation path.

Residuals, stated rather than hidden. The child lookup is one level deep, which is the shape ClickHouse creates (watchdog plus one server); a deeper tree of grandchildren would not be walked. It is also a subprocess call, so on a machine without a usable `pgrep` the escalation degrades to signalling the parent alone, as before. And in the escalation branch only, any direct child of the signalled PID is SIGKILLed, which for a server started with `CLICKHOUSE_WATCHDOG_ENABLE=0` means a child of the server process itself would be included.

## Behaviour after the fix

Both views report the watchdog, and it is the process `ps` shows as `clickhouse-watchdog`:

```
$ clickhousectl local server start --http-port 18500 --tcp-port 19500
{
  "name": "default",
  "pid": 45386,
  "http_port": 18500,
  "tcp_port": 19500,
  "version": "26.9.1.217"
}

$ clickhousectl local server list          # project-scoped pid
45386
$ clickhousectl local server list --global # same server, --global pid
45386

$ ps -o pid=,ppid=,args= -p 45386
45386     1 clickhouse-watchdog
$ ps -o pid=,ppid=,args= -p $(pgrep -P 45386)
45387 45386 /Users/al/.clickhouse/versions/26.9.1.217/clickhouse server -- --path=./ --http_port=18500 --tcp_port=19500

$ clickhousectl local server stop          # 1.3 s, SIGTERM forwarded
{ "name": "default", "already_stopped": false, "selection": "implicit" }
$ ps -o pid=,ppid=,args= -p 45386,45387
                                           # both gone
```

`server start` does not clear the environment, so the watchdog's own switches pass through to it. With SIGTERM forwarding turned off, `stop` escalates and still leaves nothing behind:

```
$ CLICKHOUSE_WATCHDOG_NO_FORWARD=1 clickhousectl local server start --http-port 18500 --tcp-port 19500
{ "name": "default", "pid": 45805, ... }
$ ps -o pid=,ppid=,args= -p 45805 $(pgrep -P 45805)
45805     1 clickhouse-watchdog
45806 45805 /Users/al/.clickhouse/versions/26.9.1.217/clickhouse server -- --path=./ --http_port=18500 --tcp_port=19500

$ time clickhousectl local server stop
{ "name": "default", "already_stopped": false, "selection": "implicit" }
2.899 total                                # the SIGTERM was ignored, so it escalated
$ ps -o pid=,ppid=,args= -p 45805,45806
                                           # both gone
```

On a binary without the escalation fix the same sequence reports the same success and leaves the server behind:

```
$ clickhousectl local server stop
{ "name": "default", "already_stopped": false, "selection": "implicit" }
$ ps -o pid=,ppid=,args= -p 41507,41508
41508     1 /Users/al/.clickhouse/versions/26.9.1.217/clickhouse server -- --path=./ --http_port=18500 --tcp_port=19500
```

A server started with the watchdog disabled (`CLICKHOUSE_WATCHDOG_ENABLE=0`) has no watchdog parent, so it keeps reporting its own PID.

## Tests

- Unit tests next to the parser in `discovery.rs`: macOS-shaped and Linux-shaped `ps -o pid=,ppid=,args=` output, where the watchdog row is the rewritten name plus the padding of the zeroed `argv` region and carries no arguments, as the real one prints it; a child whose parent is the watchdog resolves to the parent; a child whose parent is a shell resolves to itself; a missing parent row and a missing row of its own resolve to itself; each scanned PID resolves independently; a truncated prefix counts as the watchdog while `clickhouse`, `clickhouse local`, `clickhouse-client` and `clickhouse-watchdog-wrapper` do not; and `inspect_process` builds the same result from a scanned row, degrading to the cwd-derived fields when the row is missing.
- Unit tests next to `kill_process`: a shell that ignores SIGTERM and supervises a real child is stopped by the escalation and both processes are gone afterwards, and a process with no children yields no supervised PIDs. The first fails on the unfixed code.
- Unit tests next to `select_version_users`: one server reported under two PIDs is listed once, and two servers of the same name in different projects are both listed.
- Subprocess test `crates/clickhousectl/tests/local_server_watchdog_pid_test.rs`, using the isolated `HOME` plus `pgrep`/`lsof`/`ps` stand-in technique of `local_server_state_machine_test.rs`, with a fake ClickHouse that really forks a supervised child: the project-scoped and `--global` lists agree on the watchdog PID and neither reports the child; orphan recovery rebuilds metadata with the watchdog PID; a server with no watchdog reports its own PID; a project-scoped `stop` of a forwarding watchdog removes both processes; and a project-scoped `stop` of a watchdog that ignores SIGTERM exits 0 with `already_stopped: false` only because the CLI verified that the SIGKILLed pair is gone. The list-agreement tests and the escalation test fail on the unfixed code. The stand-in `pgrep` answers `-P` through the real `pgrep`, so the escalation resolves an actual process tree.
- The `local_remove_global_guard_test.rs` and `local_server_state_machine_test.rs` stand-ins were updated to the single batched `ps -o pid=,ppid=,args=` shape that discovery now uses, and their `pgrep` answers `-P` with nothing, matching fake servers that `exec` a sleep and have no children.
- Verification commands, all passing:
  - `cargo fmt --all` and `cargo fmt --all --check`
  - `cargo test -p clickhousectl` (1029 unit tests plus every integration binary, 0 failures)
  - `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` (90 tests)
  - Live check on macOS with a real ClickHouse 26.9.1.217 in a scratch project, both legs as pasted above.

## Stacking

Part of a stacked PR chain: based on `fix/666-well-formed-unknown-id-not-found`, not `main`.

Fixes #664


iskakaushik (approved):
LGTM

PR #675, opened 2026-09-02
Stop requiring `--username`/`--password` for IAM_ROLE MySQL ClickPipes

### LINKED ISSUE #670: `clickpipe create mysql --auth IAM_ROLE` still requires `--username`/`--password`
## Summary

Same defect as #662, on the MySQL source: `cloud clickpipe create mysql --auth IAM_ROLE --iam-role arn:...` fails with a clap usage error demanding `--username` and `--password`, even though IAM-role auth has no username or password.

## Where

`MySqlCreateArgs.username` / `.password` in `crates/clickhousectl/src/cloud/clickpipes.rs` are plain required `String` args with no auth-aware conditionality. The library side is already correct: `ClickPipeMutateMySQLSource.credentials` is `Option<PLAIN>` (the schema carries a `required[]` array that omits `credentials`), so this is a CLI-only fix.

## Expected behavior

Mirror PR #669 (issue #662): `--username`/`--password` required only for basic auth (`required_unless_present = "iam_role"`, because clap's `required_if_eq` does not fire on `--auth`'s default value), rejected with `--auth IAM_ROLE`, and `credentials` omitted from the request body for IAM_ROLE.

## Acceptance criteria

- IAM_ROLE create without credentials parses and sends no `credentials` object (request-builder unit test and a wiremock case in `tests/cli_request_shape_test.rs`).
- Basic auth unchanged: missing credentials still a usage error.
- `--username`/`--password` + IAM_ROLE exits 2 with a message mirroring the Postgres guard.

Found while fixing #662 during the 0.4.2 release smoke-test follow-ups.

sdairs:
## What

`clickhousectl cloud clickpipe create mysql --auth IAM_ROLE --iam-role arn:...` failed clap validation with `the following required arguments were not provided: --username --password`, so an RDS or Aurora MySQL pipe using IAM role authentication could not be created at all. This is the MySQL twin of #662 and takes the same shape as the fix that landed there.

- `--username` and `--password` are now `Option<String>` joined to each other with clap `requires`, so the pair is all-or-nothing and half a pair is still a usage error naming the missing half.
- The "basic auth needs the pair at all" rule moves to a new `validate_mysql_create_args` check, because it depends on `--auth`'s value, which clap cannot express.
- `--iam-role` gains `required_if_eq("auth", "IAM_ROLE")`, matching the Postgres flag, so `--auth IAM_ROLE` on its own asks for the role ARN only.
- MySQL gets the reverse guard it lacked entirely: `--iam-role` with basic auth is rejected instead of being silently ignored.
- `build_mysql_request` is extracted from the handler so the request shape is unit-testable, and it matches the parsed `ClickPipeMutateMySQLSourceAuthentication` instead of comparing `--auth` as a string, so a new authentication mode in the library enum is a compile error here rather than a pipe created with no credentials.
- The post-parse validation hook is generalised to `clickpipe_create_validation_error`, returning the source subcommand alongside the message so each usage error is still reported against the command that owns the flags. Postgres behaviour is unchanged.

## Why

IAM role authentication has no username or password: the role ARN in `iamRole` is the whole credential. The old code demanded the pair and then sent `credentials: {"username": ..., "password": ...}` unconditionally, so there was no way to express an IAM_ROLE MySQL source.

No API library change was needed. `ClickPipeMutateMySQLSource.credentials` is already `Option<PLAIN>` with `skip_serializing_if`, so an IAM_ROLE request omits the object entirely rather than sending an empty pair.

## Behaviour after the fix

| Invocation | Result |
| --- | --- |
| `--auth IAM_ROLE` alone | usage error naming `--iam-role` only, exit 2 |
| `--auth IAM_ROLE --iam-role arn:x` | accepted, no `credentials` key on the wire |
| `--auth IAM_ROLE --iam-role arn:x --username u --password p` | `--username and --password cannot be used with --auth IAM_ROLE; use --auth basic`, exit 2 |
| no auth flags, no credentials | `--auth basic requires --username <USERNAME> and --password <PASSWORD>`, exit 2 |
| `--username u` only | clap `MissingRequiredArgument` naming `--password`, exit 2 |
| `--iam-role arn:x` with default basic auth | `--iam-role cannot be used with --auth basic; use --auth IAM_ROLE`, exit 2 |

## Tests

- Clap coverage next to `MySqlCreateArgs`: IAM_ROLE without credentials parses with both fields `None`; the IAM_ROLE-only error names `--iam-role` and neither credential flag; basic auth with no credentials yields the basic-auth validation message; half a pair is a clap `MissingRequiredArgument` naming the omitted flag; the pair with IAM_ROLE yields the rejection; `--iam-role` with basic auth yields the new guard. `assert_mysql_value` now gives each auth mode only its own flags, and the maximal parse test no longer pairs credentials with IAM_ROLE.
- `build_mysql_request` unit tests: minimal basic auth sends `credentials.username`/`password`, maximal IAM_ROLE sends `credentials == None` with the role ARN, certificate file contents, server ID and roles, and a defensive case list covering all four validation messages plus a malformed table mapping.
- Subprocess and wiremock coverage in `crates/clickhousectl/tests/cli_request_shape_test.rs`: `mysql_iam_role_create_omits_the_credentials_object`, `mysql_basic_auth_create_sends_the_credentials_object`, `mysql_credentials_with_iam_role_auth_are_rejected` (exit 2, usage line naming `clickpipe create mysql`) and `mysql_iam_role_with_basic_auth_is_rejected`.
- README pin test `readme_documents_mysql_iam_role_authentication` covers the new section and the IAM_ROLE example.
- `create mysql --help` now carries a `MYSQL INPUT RULES:` block for the auth rules, since the credential pair left the usage line when it became `Option`; `mysql_help_documents_input_rules` pins it, the Postgres README window now ends at the MySQL section so its shared wording cannot satisfy the Postgres assertions, and `every_clickpipe_create_validation_source_resolves_to_a_subcommand` checks every source name the post-parse hook can return against the real command tree.
- Verification commands, all clean:
  - `cargo fmt --all`
- Not verified live: acceptance of an omitted `credentials` object for an IAM_ROLE MySQL pipe needs an RDS or Aurora MySQL fixture. The schema's `required[]` omits `credentials`, and the Postgres equivalent behaves this way, so this is the same bet #662 made.

## Stacking

Part of a stacked PR chain: based on `fix/664-server-list-pid-consistency`, not `main`.

Fixes #670

### DIFF
--- a/README.md
+++ b/README.md
@@ -1009,6 +1009,15 @@ clickhousectl cloud clickpipe create mysql <service-id> \
   --table-mapping "mydb.users:mydb_users" \
   --server-id 4242
 
+# From an RDS or Aurora MySQL with IAM role authentication (CDC)
+# IAM_ROLE auth takes no --username/--password: the role ARN is the credential
+clickhousectl cloud clickpipe create mysql <service-id> \
+  --name my-rds-mysql-pipe \
+  --host mysql.abcdefg.us-east-1.rds.amazonaws.com \
+  --mysql-type rdsmysql \
+  --auth IAM_ROLE --iam-role "$MYSQL_IAM_ROLE_ARN" \
+  --table-mapping "mydb.users:mydb_users"
+
 # From MongoDB (CDC)
 clickhousectl cloud clickpipe create mongodb <service-id> \
   --name my-mongo-pipe \
@@ -1213,6 +1222,19 @@ must be between 10 and 600, and `--filter` takes a Pub/Sub CEL subscription
 filter of at most 256 characters. Both are checked before the request is sent.
 
 Use `clickhousectl cloud clickpipe create <source> --help` for the full list of options per source type.
+
+#### MySQL ClickPipe authentication
+
+`clickpipe create mysql` authenticates with either a username and password
+(the default `--auth basic`) or an AWS IAM role on an RDS or Aurora MySQL
+source (`--auth IAM_ROLE`).
+
+`--auth IAM_ROLE` requires `--iam-role`; the CLI rejects `--iam-role` with
+basic auth rather than silently ignoring it. `--username` and `--password` are
+basic-auth only and must be given together: they are required for the default
+`--auth basic`, and rejected with `--auth IAM_ROLE`, where the role ARN is the
+whole credential and no `credentials` object is sent. Each rule is a usage
+error (exit code 2) before any request is made.
 
 #### Reverse private endpoints (PrivateLink, Private Service Connect)
 
--- a/crates/clickhousectl/src/cloud/cli.rs
+++ b/crates/clickhousectl/src/cloud/cli.rs
@@ -62,11 +62,15 @@ impl CloudArgs {
             )
     }
 
-    pub fn postgres_clickpipe_validation_error(&self) -> Option<String> {
+    /// The `clickpipe create <source>` validation message and the source
+    /// subcommand it belongs to, if the flags cannot describe the chosen
+    /// `--auth`. Covers both database sources whose credential flags depend on
+    /// `--auth`'s value: `postgres` and `mysql`.
+    pub fn clickpipe_create_validation_error(&self) -> Option<(&'static str, String)> {
         let CloudCommands::ClickPipe { command } = &self.command else {
             return None;
         };
-        command.postgres_create_validation_error()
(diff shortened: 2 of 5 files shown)


iskakaushik (approved):
LGTM

PR #703, opened 2026-09-05
fix(cli): preserve query endpoint keys when adding access

### LINKED ISSUE #596: `service query-endpoint create` silently drops API keys already authorized on the endpoint
From the CLI↔API gap analysis (2026-08-27). `instance_query_endpoint_upsert` has full-replace semantics, and `service query-endpoint create` (`build_query_endpoint_create_request`, `crates/clickhousectl/src/cloud/services.rs:1225-1240`) sends exactly what's on the command line without merging — so running it against an endpoint that already has authorized keys silently de-authorizes them.

The auto-provisioning path used by `service query` (`crates/clickhousectl/src/cloud/service_query.rs:188-194`) *does* read-merge existing keys, so the two paths are inconsistent.

Related sharp edges on the same command worth fixing together:
- Omitting `--role` sends `roles: []` rather than erroring.
- `--allowed-origins` silently defaults to `"*"` (services.rs:1235-1238).

sdairs:
`cloud service query-endpoint create` could silently revoke every existing API key because the endpoint API replaces its configuration. The command now reads the endpoint, adds and deduplicates requested keys, and preserves browser origins unless `--allowed-origins` explicitly changes them. Failed or incomplete reads stop the update before any write.

`--role` is required and explicitly replaces the endpoint-wide role list. First creation requires `--allowed-origins` (including an explicit `'*'` for unrestricted browser origins). `--replace-open-api-keys` deliberately replaces the authorized-key list and requires at least one `--open-api-key`. README and command help document these semantics and the need to avoid concurrent endpoint updates.

Validation: request-builder and clap tests cover required roles, additive/replacement keys, deduplication, explicit and omitted origins, and sparse responses. Subprocess wiremock tests verify read-before-write order, Basic auth, JSON output, first creation, updates, replacement, and no writes on failed/incomplete reads or invalid arguments. Checks:

- `cargo fmt --all`

Closes #596

### DIFF
--- a/README.md
+++ b/README.md
@@ -678,6 +678,11 @@ clickhousectl cloud service delete <service-id>
 clickhousectl cloud service delete <service-id> --force
 ```
 
+`query-endpoint create` adds and deduplicates API keys while preserving existing browser origins. `--role` is required and replaces the endpoint-wide roles for **all** authorized keys; roles are not assigned per key.
+Pass `--allowed-origins` on first creation or to change browser access (`'*'` explicitly allows every origin). Use `--replace-open-api-keys` with `--open-api-key` to deliberately replace the entire authorized-key list.
+The command reads the existing configuration before updating; a failed or incomplete read prevents changes to unknown fields. Avoid concurrent changes to the same query endpoint.
+
+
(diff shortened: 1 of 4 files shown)


iskakaushik (approved):
LGTM

PR #706, opened 2026-09-05
feat(cli): add explicit API key expiry clearing

### LINKED ISSUE #698: Allow clearing an API key expiry without changing other key settings
`cloud key update --expires-at` can set an expiration but cannot remove one. Omitting the flag means no change, and the flag parser requires an RFC 3339 timestamp, so neither an empty value nor null can express "never expires".

The API explicitly supports `expireAt: null` to clear expiry. `ApiKeyPatchRequest.expire_at` is currently `Option<DateTime<Utc>>` with `skip_serializing_if = Option::is_none` (`crates/clickhouse-cloud-api/src/models/api_keys.rs:111-115`), so the library cannot emit the clear operation either. The CLI builder at `crates/clickhousectl/src/cloud/api_keys.rs:303` collapses absence into omission. The existing `BackupConfigurationPatchRequest.backup_start_time` already demonstrates the needed omitted/value/null distinction.

## Proposed command

```sh
chctl cloud key update <key-id> --clear-expiry
```

`--clear-expiry` conflicts with `--expires-at`. No expiry flag preserves the existing expiry. The clear flag sends exactly `"expireAt": null` and does not modify roles, IP allowlist, state or name unless their own flags are present. Retain normal write/API-key authentication rules.

## Implementation sequence

1. **Library PR first:** provide a request representation that distinguishes omit, timestamp and explicit null (following the existing nested-Option serialization pattern used for backup start-time). Keep response `ApiKey.expire_at` as tolerant `Option<DateTime<Utc>>`; update conversions/construction sites only as required. Add model and wiremock tests for all three serialized states and a focused analyzer regression if necessary to keep request nullable handling correctly inventoried.
2. **CLI PR second:** add `--clear-expiry` to key update, clap conflict and builder option; set explicit null only for the clear flag. Keep absent fields omitted, preserve error conversion and normal JSON output.
3. Add README examples for setting/removing an expiry and explain that omission leaves the current expiry unchanged.

## Tests and acceptance criteria

- Model tests distinguish `{}`, `{"expireAt":"..."}`, and `{"expireAt":null}`; the clear representation also survives deserialization if supported for request write-back.
- Clap tests assert the clear flag parses, both expiry flags conflict with a usage error, and existing timestamps still parse.
- Minimal/maximal builder tests verify clear alone, omitted expiry, explicit timestamp, and clear alongside unrelated fields.
- Subprocess/wiremock verifies explicit null on the real PATCH request and omission when no expiry change was requested; also verify OAuth write rejection occurs before HTTP.
- No breaking change to create's existing "omit for never" behavior.

Related #597 covers clearing list-valued roles, allowlists and tags; this is a separate nullable-scalar/model gap and should share conventions without being lost inside that list-only scope.

All CLI work follows the existing domain pattern: clap definitions and exhaustive read/write classification, CloudClient wrappers with structured error conversion, request builders, optional-safe human output and automatic/explicit JSON. Add README examples. New source/test paths must be registered with both CI classifiers. Run cargo fmt --all; CLI clippy/tests plus no-default-features check/clippy; touched library/analyzer tests and clippy plus Python tests for library work. Library changes land in their own PR before CLI adoption.

sdairs:
`cloud key update <key-id> --clear-expiry` now removes an existing API key expiry using an explicit `expireAt: null` PATCH. It conflicts with `--expires-at`; omitting both leaves expiry unchanged. Roles, name, state and IP allowlist change only when their own flags are supplied.

Use the nullable request representation from the preceding library PR. Preserve existing API-key write authentication, error conversion and JSON output. Add clap conflict/default tests, minimal/maximal request builders, exact subprocess PATCH assertions for omission/set/clear, and OAuth rejection before HTTP. Document setting and removing expiry in README.

Combined-stack validation also passed all seven PgBouncer and expiry subprocess tests and CLI Clippy.

Closes #698.

Validation:

- `cargo fmt --all`
- `cargo test -p clickhousectl -- --test-threads=1` — 1,433 passed: 1,026 unit, 244 cloud subprocess, 113 local, 50 telemetry (2 existing ignores)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1515,8 +1515,12 @@ clickhousectl cloud key create --name ci-key \
 clickhousectl cloud key update <key-id> \
   --name renamed-key \
   --state disabled
+clickhousectl cloud key update <key-id> --expires-at 2030-12-31T23:59:59Z
+clickhousectl cloud key update <key-id> --clear-expiry
 clickhousectl cloud key delete <key-id>
 ```
+
+On update, omitting both expiry flags keeps the current expiry. `--clear-expiry` removes it and conflicts with `--expires-at`; other key settings change only when their flags are supplied.
 
 ### Activity
 
--- a/crates/clickhousectl/src/cloud/api_keys.rs
+++ b/crates/clickhousectl/src/cloud/api_keys.rs
@@ -82,9 +82,13 @@ pub enum KeyCommands {
         #[arg(long)]
         role_id: Vec<String>,
 
-        /// New expiry as RFC 3339 (e.g. 2025-12-31T23:59:59Z)
-        #[arg(long, value_parser = parse_datetime)]
+        /// New expiry as RFC 3339; conflicts with --clear-expiry
+        #[arg(long, value_parser = parse_datetime, conflicts_with = "clear_expiry")]
         expires_at: Option<String>,
+
+        /// Remove the expiry; conflicts with --expires-at
+        #[arg(long, conflicts_with = "expires_at")]
+        clear_expiry: bool,
 
         /// Key state (enabled or disabled)
         #[arg(long)]
@@ -157,6 +161,7 @@ pub async fn run(client: &CloudClient, command: KeyCommands, json: bool) -> Clou
             name,
             role_id,
             expires_at,
+            clear_expiry,
             state,
             ip_allow,
             org_id,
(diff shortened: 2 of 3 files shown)


iskakaushik (approved):
LGTM

PR #707, opened 2026-09-05
feat(cli): expose ClickPipe settings with verified update semantics

### LINKED ISSUE #682: cloud clickpipe settings update: verify PUT replace semantics; three settings are never sent

`build_clickpipe_settings_request` hardcodes `clickhouse_max_download_threads`, `clickhouse_min_insert_block_size_bytes` and `clickhouse_parallel_distributed_insert_select` to `None`, so they are never sent. A handler comment in `clickpipes.rs` says an omitted field in the PUT resets it (that is why `kafka_read_committed` is read back and resent), but the spec only says "Send key-value pairs" and nothing verifies replace semantics.

If PUT really replaces, every `settings update` resets those three settings. Needs a live PUT against a test pipe to settle, then either expose the three flags or correct the comment/help. Help currently says: only the settings named on the command line are sent; run `settings get` first and re-pass everything you want to keep.

sdairs:
`cloud clickpipe settings update` hardcoded three supported settings to `None`, and its help told users to resend every value without verified replacement semantics. Expose `--clickhouse-max-download-threads`, `--clickhouse-min-insert-block-size-bytes`, and `--clickhouse-parallel-distributed-insert-select` with the API's numeric bounds, plus explicit `--kafka-read-committed true|false` control.

Object-storage updates send the requested fields and preserve omitted values. Kafka updates retain the existing read-before-write behavior when isolation is omitted, but now refuse missing/null current state instead of inventing `false`. An explicit Kafka value skips that settings read and requires a confirmed Kafka source. Empty updates fail in clap before HTTP. README and command help describe the resulting behavior, and detail output uses the shared human formatter.

Live verification on 2026-09-05 used a disposable S3 ClickPipe and service created for this issue. After setting all three new fields plus two existing fields to nondefault values, a PUT changing only the polling interval preserved all five; a separate GET confirmed them. Explicit zeros and `false` also persisted, and an empty body returned HTTP 400. These observations establish merge behavior for the tested object-storage settings only; Kafka omission semantics were not live-verified. Both owned resources were deleted, service deletion was independently confirmed by HTTP 404, and the temporary creation response was removed. No existing user resources were modified.

Validation:

- `cargo fmt --all` and `git diff --check`.
- `cargo clippy -p clickhousectl -- -D warnings`.
- `cargo check -p clickhousectl --no-default-features`.
- `cargo clippy -p clickhousectl --all-targets --no-default-features -- -D warnings`.
- Settings unit tests: 13 passed; settings wiremock subprocess tests: 11 passed.
- `cargo test -p clickhousectl -- --test-threads=1`: 1432 passed, 0 failed, 2 existing ignored tests across 24 test binaries.

Closes #682.

### DIFF
--- a/README.md
+++ b/README.md
@@ -912,7 +912,14 @@ clickhousectl cloud clickpipe settings update <service-id> <clickpipe-id> \
   --object-storage-max-insert-bytes 268435456 \
   --object-storage-use-cluster-function true \
   --clickhouse-max-threads 16 --clickhouse-max-insert-threads 4 \
+  --clickhouse-max-download-threads 8 \
+  --clickhouse-min-insert-block-size-bytes 20971520 \
+  --clickhouse-parallel-distributed-insert-select 1 \
   --clickhouse-parallel-view-processing false
+
+# Change Kafka consumer isolation explicitly
+clickhousectl cloud clickpipe settings update <service-id> <clickpipe-id> \
+  --kafka-read-committed true
 
 # Manage reverse private endpoints for private source connectivity
 clickhousectl cloud clickpipe reverse-private-endpoint list <service-id>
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #708, opened 2026-09-05
feat(api): expose checked Postgres config enum values

### LINKED ISSUE #591: `postgres config --set` silently drops misspelled/unknown keys and reports success
From the CLI↔API gap analysis (2026-08-27). `PgConfig` deserialization does not use `deny_unknown_fields` and the CLI does no key validation (`parse_pg_config_overrides`, `crates/clickhousectl/src/cloud/postgres.rs:484`), so:

```
chctl cloud postgres config patch <ID> --set max_conections=500   # typo
Configuration patched
```

actually sends `{"pgConfig":{},"pgBouncerConfig":{}}` — the typo'd key is silently discarded and the user believes the change applied.

The model has all 31 typed GUC fields (`crates/clickhouse-cloud-api/src/models/postgres.rs:1208-1271`), so the CLI can validate `--set` keys against the known field names and reject unknowns. Related sharp edges to consider in the same fix:

- Enum-valued GUCs (`default_transaction_isolation`, `ssl_min_protocol_version`, `wal_compression`) pass through unvalidated because of their `Unknown(String)` catch-alls — contrast with `--ha-type`/`--pg-version`, which use `PossibleValuesParser`.
- On the `--file` paths, an omitted `pgConfig` key is silently sent as `{}` (`instance_config_from_json`, postgres.rs:467-471) rather than rejected.
- `config replace` is file-only (no `--set`), and `config get` ignores the `--json` flag (always prints JSON).

sdairs:
Expose `VALUES` on the three closed Postgres configuration enums so CLI validation can use the API library's accepted values. The existing analyzer checks these constants against each enum and the OpenAPI schema, preventing a copied CLI allowlist from falling behind.

This prerequisite adds round-trip coverage for all values. The following CLI layer validates config keys, enum values and file sections before networking.

Validation: formatting, library/analyzer Clippy with warnings denied, full library/analyzer tests including vendored drift coverage, and all 91 Python tests passed. Integrated CLI Clippy also passed.

Refs #591.

### DIFF
--- a/crates/clickhouse-cloud-api/src/models/postgres.rs
+++ b/crates/clickhouse-cloud-api/src/models/postgres.rs
@@ -732,6 +732,12 @@ impl std::fmt::Display for PgConfigDefaultTransactionIsolation {
     }
 }
 
+impl PgConfigDefaultTransactionIsolation {
+    /// Wire values accepted by the API, excluding the catch-all.
+    pub const VALUES: &'static [&'static str] =
+        &["read committed", "repeatable read", "serializable"];
+}
+
 /// Inline enum for `pgConfig.ssl_min_protocol_version`.
 #[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
 pub enum PgConfigSslMinProtocolVersion {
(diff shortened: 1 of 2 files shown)


iskakaushik (approved):
LGTM

PR #709, opened 2026-09-05
fix(cli): reject invalid Postgres configuration before networking

### LINKED ISSUE #591: `postgres config --set` silently drops misspelled/unknown keys and reports success

sdairs:
Postgres configuration inputs could silently discard misspelled GUC names, `null` values, and unsupported values for closed enum settings, then report a successful update. This change validates the API's 31 supported `pgConfig` keys and the allowed values for `default_transaction_isolation`, `ssl_min_protocol_version`, and `wal_compression` before organization discovery or an API request.

The validation covers `config patch --set`, standalone PgConfig files used by create, read-replica create, and restore, and sectioned files used by config patch/replace. Sectioned files must explicitly contain object-valued `pgConfig` and `pgBouncerConfig` sections, so a missing section is no longer silently converted to `{}`. Explicit empty sections and `false`/zero values remain preserved, and PgBouncer parameter names remain open-ended string maps.

The API prerequisite #708 supplies analyzer-checked enum value constants used by this validation.

Validation:

- `cargo fmt --all`
- `cargo test -p clickhousectl cloud::postgres::tests::`
- focused real-binary Wiremock tests for all six PgConfig input paths

Combined stack validation: all 267 Cloud subprocess tests passed (one existing ignored), and CLI Clippy passed.

Closes #591.


iskakaushik (approved):
LGTM

PR #714, opened 2026-09-05
feat(cli): explicitly clear Cloud roles, allowlists, tags and DNS mappings

### LINKED ISSUE #597: List-valued fields cannot be cleared (api-key roles/IP list, member roles, postgres tags)
From the CLI↔API gap analysis (2026-08-27). Several update commands collapse an empty list to `None` (no change), so a list can be replaced but never emptied:

- `api-key update`: empty `--role-id` and empty `--ip-allow` both become `None` (`crates/clickhousectl/src/cloud/api_keys.rs:305-320`) — cannot clear role assignments or the IP allowlist.
- `member update`: empty `--role-id` mapped to `None` (`crates/clickhousectl/src/cloud/organizations.rs:398`) — cannot clear all roles.
- `postgres update`: tags are diff-only (`--add-tag`/`--remove-tag`); no wholesale set/clear, and the read-modify-write merge fails hard if the GET response omits `tags` (`merge_response_tags`, `crates/clickhousectl/src/cloud/postgres.rs:652`).

Suggested shape: explicit `--clear-roles` / `--clear-ip-allow` / `--clear-tags` flags (an explicit flag avoids making an accidental empty value destructive).

sdairs:
Several Cloud update commands could replace list-valued fields but could not explicitly empty them because an empty CLI value was serialized as an omitted field. Add explicit clear flags for API key roles and IP allowlists, member roles, Postgres tags, and reverse private endpoint custom DNS mappings while preserving omission as no change.

Postgres tag clearing now sends an empty list directly without fetching current tags. Add/remove tag updates retain their safe read/merge behavior and refuse to PATCH when the API omits the current tag snapshot. Reverse private endpoint updates require either replacement mappings or `--clear-custom-private-dns-mappings`, and reject both together.

Closes #597.

Validation:

- `cargo fmt --all --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -824,6 +824,10 @@ clickhousectl cloud postgres update <pg-id> \
   --size c6gd.4xlarge \
   --ha-type sync \
   --add-tag env=prod --remove-tag legacy
+clickhousectl cloud postgres update <pg-id> --clear-tags
+
+# --clear-tags replaces the tag list with an empty list and conflicts with
+# --add-tag and --remove-tag; omitting all three leaves tags unchanged.
 
 # Delete (works from any state, including running; no stop needed first)
 clickhousectl cloud postgres delete <pg-id>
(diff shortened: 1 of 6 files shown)


iskakaushik (approved):
LGTM

PR #722, opened 2026-09-05
feat(api): expose BigQuery replication mode values

### LINKED ISSUE #592: BigQuery ClickPipe create sends zero-valued settings on every request
From the CLI↔API gap analysis (2026-08-27). `ClickPipeBigQuerySettings` fields are non-`Option` in the library, so the CLI's `..Default::default()` at `crates/clickhousectl/src/cloud/clickpipes.rs:2063-2066` doesn't omit them — it **serializes** `allowNullableColumns: false`, `initialLoadParallelism: 0`, `snapshotNumRowsPerPartition: 0`, `snapshotNumberOfParallelTables: 0` on every `clickpipe create bigquery`.

Whether the server treats `0` as "use default" or literally is unverified, but the CLI is asserting values the user never chose either way.

Two parts:
1. Bug: stop sending unchosen zero values (likely a library-side model question too — should these fields be `Option`? If the spec requires them, the CLI needs real defaults).
2. Enhancement: add flags for the settings fields, as none are currently settable. Also `settings.replicationMode` is hardcoded to `"snapshot"` via `parse_enum("snapshot")` (clickpipes.rs:2064) — harmless today (only variant) but should follow the enum.

sdairs:
BigQuery ClickPipe replication mode is a typed Cloud API enum, but it did not expose the analyzer-checked wire-value list needed by CLI validation. Add `ClickPipeBigQueryPipeSettingsReplicationmode::VALUES` with the currently supported `snapshot` value so downstream callers can reject unsupported modes before HTTP without duplicating the enum vocabulary.

This is the API prerequisite for #592.

## Validation

- `cargo fmt --all`
- `cargo test -p clickhouse-cloud-api -p clickhouse-openapi-analyzer`
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` (91 passed)

### DIFF
--- a/crates/clickhouse-cloud-api/src/models/clickpipes.rs
+++ b/crates/clickhouse-cloud-api/src/models/clickpipes.rs
@@ -67,6 +67,11 @@ impl std::fmt::Display for ClickPipeBigQueryPipeSettingsReplicationmode {
     }
 }
 
+impl ClickPipeBigQueryPipeSettingsReplicationmode {
+    /// Wire values accepted by the API, excluding the catch-all.
+    pub const VALUES: &'static [&'static str] = &["snapshot"];
+}
+
 /// Inline enum for `ClickPipeBigQueryPipeTableMapping.tableEngine`.
 #[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


iskakaushik (approved):
LGTM

PR #725, opened 2026-09-05
feat(clickpipes): expose BigQuery snapshot tuning

### LINKED ISSUE #592: BigQuery ClickPipe create sends zero-valued settings on every request

sdairs:
BigQuery ClickPipe creation could not set any source snapshot tuning, even though the typed API request supports nullability, initial-load parallelism, rows per partition, and parallel-table controls. Add `--allow-nullable-columns`, `--initial-load-parallelism`, `--snapshot-rows-per-partition`, and `--snapshot-parallel-tables`, plus an explicit `--replication-mode` whose accepted values come from the API model.

Omitted tuning flags remain absent from the request so ClickPipes applies its service defaults; explicit values, including `false` and fractional API `number` values, are preserved. Non-finite numeric values are rejected as usage errors before reading the service-account file or sending HTTP, while the CLI adds no range or integer constraint absent from the API schema. The README now includes a tuned BigQuery snapshot example and documents omission behavior. Existing destination database, destination role, service-account file, staging path, and table mapping behavior remains intact.

Closes #592.

## Validation

- `cargo fmt --all`
- `cargo test -p clickhousectl` (1,065 unit passed/1 ignored; 304 Cloud request-shape passed/1 ignored; all local subprocess and 50 telemetry tests passed)
- focused BigQuery unit tests (5 passed), covering parsing and defensive request-builder validation
- focused BigQuery request-shape tests, covering omitted and explicit settings on the wire plus rejection before key-file access or HTTP

### DIFF
--- a/README.md
+++ b/README.md
@@ -1319,6 +1319,11 @@ clickhousectl cloud clickpipe create bigquery <service-id> \
   --name my-bq-pipe \
   --service-account-file ./sa-key.json \
   --staging-path gs://bucket/staging \
+  --replication-mode snapshot \
+  --allow-nullable-columns true \
+  --initial-load-parallelism 4 \
+  --snapshot-rows-per-partition 1000000 \
+  --snapshot-parallel-tables 3 \
   --table-mapping "dataset.table:target_table"
 
 # From Google Cloud Pub/Sub (limited preview: contact support to enable it)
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #738, opened 2026-09-05
fix(service): normalize IP removal warning matches

### LINKED ISSUE #589: Support descriptions on IP allowlist entries (`service` and `api-key`)
From the CLI↔API gap analysis (2026-08-27): `IpAccessListEntry.description` is hardcoded to `None` in both copies of `parse_ip_access_entries` — `crates/clickhousectl/src/cloud/services.rs:852` (used by `service create --ip-allow` and `service update --add-ip-allow`) and `crates/clickhousectl/src/cloud/api_keys.rs:228` (used by `api-key create/update --ip-allow`).

Users cannot label allowlist entries; the only description ever sent is the auto-generated "Allow all (created by clickhousectl)" default when `--ip-allow` is omitted on `service create`.

Suggested shape: `--ip-allow 'CIDR[:description]'` (or a `SOURCE=DESC` key=value form, matching the existing `--tag` conventions). While here, consider deduplicating the two identical parsers into `src/cloud/shared.rs`.

sdairs:
`cloud service update --remove-ip-allow` validated and normalized `SOURCE[=DESCRIPTION]` values for the PATCH request, but warning detection compared the original CLI text with the current service sources. A described or padded removal therefore succeeded while incorrectly reporting that no entry was removed.

Warning detection now consumes the already-validated request entries and compares their parsed `source` fields. Descriptions and surrounding source whitespace no longer cause false warnings, invalid entries still fail before any API request, and unmatched removals continue to warn through the existing output channel.

Tests cover described and padded IPv4/IPv6 matches, normalized unmatched warnings, the real CLI PATCH body and stderr behavior, and pre-network invalid-input rejection.

Validation: formatting, default clippy, no-default check and all-target clippy passed; five focused unit tests and both relevant real-binary mock tests passed.

References #589.
Follow-up correction for PR #721.

### DIFF
--- a/crates/clickhousectl/src/cloud/services.rs
+++ b/crates/clickhousectl/src/cloud/services.rs
@@ -1537,9 +1537,14 @@ fn unmatched<'a>(requested: &'a [String], current: &HashSet<&str>) -> Vec<&'a st
         .collect()
 }
 
-/// Human-readable warnings for every `--remove-*` value on `options` that
-/// does not match anything in `current`.
-fn unmatched_removal_warnings(options: &ServiceUpdateOptions, current: &Service) -> Vec<String> {
+/// Human-readable warnings for every `--remove-*` value that does not match
+/// anything in `current`. IP entries come from the validated request so the
+/// comparison uses the same parsed source that the API receives.
+fn unmatched_removal_warnings(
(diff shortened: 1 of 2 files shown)


iskakaushik (approved):
LGTM

PR #742, opened 2026-09-05
fix(api): preserve omitted organization role fields on update

### LINKED ISSUE #577: Expose organization custom roles (list/get/create/update/delete)
From the CLI↔API gap analysis (2026-08-27): `organization_roles_get_list`, `organization_role_post`, `organization_role_get`, `organization_role_patch`, `organization_role_delete` (`crates/clickhouse-cloud-api/src/client/organizations.rs`) have no CLI surface. These are **GA**.

This is a discoverability hole as well as a feature gap: `member update --role-id`, `org invite --role-id`, and `api-key create --role-id` all take role IDs that the CLI currently gives users no way to look up.

Suggested surface: `cloud org role list|get|create|update|delete`.

sdairs:
`RoleUpdateRequest` required `name`, `actors`, and `policies` even though the role PATCH endpoint documents every field as optional. Callers therefore had to resend complete role state for a one-field update, risking accidental replacement of actors or policies.

This makes all three request fields optional and omits absent fields during serialization. Because the schema is named `RoleUpdateRequest` rather than `*PatchRequest`, the analyzer receives a narrow, documented optionality exemption for these exact fields. The organization lifecycle integration now performs a name-only PATCH and verifies that actors and policies remain unchanged.

Validation:

- `cargo fmt --all`
- `cargo test -p clickhouse-cloud-api -p clickhouse-openapi-analyzer`
- `python3 scripts/check-openapi-drift.py --dry-run` (0 actionable drift)

Prerequisite for #577.

### DIFF
--- a/crates/clickhouse-cloud-api/src/models/rbac.rs
+++ b/crates/clickhouse-cloud-api/src/models/rbac.rs
@@ -173,7 +173,10 @@ pub struct RoleCreateRequest {
 /// `RoleUpdateRequest` from the ClickHouse Cloud API.
 #[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
 pub struct RoleUpdateRequest {
-    pub actors: Vec<String>,
-    pub name: String,
-    pub policies: Vec<RBACPolicyCreateRequest>,
+    #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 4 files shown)


iskakaushik (approved):
LGTM

PR #756, opened 2026-09-06
docs: introduce ClickHouse product family

sdairs:
Closes #755.

The README now opens with a centered official ClickHouse full logo in light and dark variants, adds linked Slack, X, and YouTube badges, directs new users to the 30-day ClickHouse Cloud trial with $300 in credits, and positions ClickHouse, ClickHouse Managed Postgres, ClickStack, and ClickPipes through the workflows `clickhousectl` supports. Product descriptions, links, and the trial offer were checked against the current ClickHouse brand and product pages and aligned with ClickHouse/examples#409.

Stacked on #754.

Validation:

- `cargo fmt --all`
- `cargo clippy -p clickhousectl -- -D warnings`
- `cargo check -p clickhousectl --no-default-features`
- `cargo clippy -p clickhousectl --all-targets --no-default-features -- -D warnings`
- `cargo test -p clickhousectl` (one environment-dependent test fails because a local OrbStack process already occupies ports 8123 and 9000; the isolated retry reports the same port notice)

PR #765, opened 2026-09-07
docs: clarify partial local server configs

### LINKED ISSUE #764: Clarify partial local server configs in help and README
Users looking for a default config to edit can miss that `local server start --config` already accepts a partial overlay. The flag describes a custom config file, while the inheritance explanation appears only in `server configs --help` and after the README example.

Clarify the existing experience without adding commands or changing config loading:

- Explain partial overrides where users discover `--config`, including the global named-file store.
- Show that one overlay can contain server settings and users, profiles, and quotas; a separate `users.xml` is unnecessary with embedded defaults.
- Explain how an existing `config.d` fragment can be selected directly.
- Distinguish the editable source from its runtime copy and generated preprocessed output.
- Document stop/start with the same `--config` to apply edits, and removal of the managed overlay when the flag is omitted.
- Keep help concise and put storage layout and merge details in the README.

sdairs:
Stacked on #756 (`docs/755-readme-products`).

`local server start --config` was described as a custom config file at the point of use, making it easy to assume a complete default config was required. The flag and empty config-list guidance now explain that files contain only changed settings and inherit the remaining defaults.

The README demonstrates server and profile settings in one overlay, explains how to use an existing `config.d` fragment, and documents source edits, restart selection, generated output, and file lookup. Help keeps the short operational guidance; config loading and JSON output are unchanged.

Closes #764.

Validation (run before restacking; patch unchanged):

- `cargo fmt --all` and `git diff --check`
- `cargo test -p clickhousectl`: one existing port-dependent stderr assertion fails (details below); all other executed tests pass, with one ignored test.
- `cargo test -p clickhousectl --test local_version_error_test --test telemetry_test` covers the binaries the full run stopped before reaching.
- Manually reviewed rendered server/start/configs help against the repository help standard.

Local test limitation: `config_and_argument_failures_carry_their_full_human_detail_in_json` fails because OrbStack already listens on 8123 and 9000. The CLI emits `Note: default ports in use, auto-assigned HTTP:8124 TCP:9001` before the missing-config error, while the unchanged test expects only the error. A focused rerun reproduces it. The change does not touch port selection or error rendering.

Restacking validation: `git range-diff` confirms the same single-commit patch; `cargo fmt --all` and `git diff --check` pass on the new base.

### DIFF
--- a/README.md
+++ b/README.md
@@ -489,7 +489,9 @@ Version removal and server-data removal are separate operations:
 
 #### Custom config files
 
-Drop ClickHouse config files into `~/.clickhouse/configs/` and apply one by name when starting a server:
+`clickhousectl local server start` uses ClickHouse's embedded defaults without creating an editable main config file or loading `/etc/clickhouse-server/config.xml`. To customize the server, supply a **partial config** containing only the settings you want to change; everything else inherits the defaults.
+
+Create a file in the global `~/.clickhouse/configs/` directory and select it by name. This example enables the query log and sets the default user's `max_threads` query setting:
 
 ```bash
 mkdir -p ~/.clickhouse/configs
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #767, opened 2026-09-08
fix(cloud): tolerate closed stderr for notices

### LINKED ISSUE #677: Credential-precedence notice panics (exit 101) on a closed stderr before any cloud command runs
## Summary

`cloud::run` prints the credential-precedence `note:` (env `CLICKHOUSE_CLOUD_API_KEY`/`_SECRET` present but shadowed by file credentials) with `eprintln!` at `crates/clickhousectl/src/cloud/mod.rs:75`. `eprintln!` panics when the write fails, and it fails with `BrokenPipe` as soon as whatever was reading stderr goes away. Because this line runs before the command is dispatched, **every** `cloud` subcommand exits 101 without doing anything when a user has both env and file credentials and stderr is closed. The two `[debug]` lines just below it (lines 79-80) have the same shape, gated on `--debug`.

Same bug class as #598 (forced delete stop-poll progress) and the retirement-warning fix in PR #657. The fix there was `crate::cloud::output::eprint_line`, which discards a write failure: the exit code, not the line, reports the outcome.

## Reproduce

Shell with `CLICKHOUSE_CLOUD_API_KEY`/`CLICKHOUSE_CLOUD_API_SECRET` exported and a `.clickhouse/credentials.json` in the project, then close stderr before the process writes:

```
clickhousectl cloud service list --json 2>&-
# exit 101, nothing sent to the API
```

## Found while

Writing the closed-stderr regression tests for PR #657: the fixture has both env and file credentials, so the tests only exercised the retirement warning after `env_remove`-ing the env credentials, because the notice panicked first.

## Proposed fix

- Route the notice and the two `[debug]` lines through `eprint_line`.
- Add a subprocess test in `tests/cli_request_shape_test.rs` modelled on `forced_service_delete_survives_a_closed_stderr_while_stopping`: env + file credentials, stderr dropped before wait, assert exit 0 and that the request reached the mock.
- Worth a grep for the remaining `eprintln!` in `src/cloud/services.rs` (readiness "Waiting for..." line, `OK` acknowledgement, idle wake notice) and deciding whether the whole cloud surface should stop using `eprintln!` for informational lines.

sdairs:
fix(cloud): tolerate closed stderr for notices

Cloud resource commands could panic with exit 101 before dispatch when credential-precedence or debug notices encountered a closed stderr. Route informational Cloud stderr lines through the existing best-effort writer so the command continues and its exit status reflects the operation; this also covers the remaining auth and service-query status notices in the same bug class.

Add a subprocess regression with environment credentials shadowed by project credentials and `--debug`: it closes the stderr reader before waiting, asserts exit 0, and verifies the organization request reached the mock API. Document that Cloud stderr notices are best-effort.

Validation:
- `cargo fmt --all`
- `cargo test -p clickhousectl --test cli_request_shape_test closed_stderr` (4 passed)

Closes #677

### DIFF
--- a/README.md
+++ b/README.md
@@ -672,6 +672,8 @@ inactive and identifies the source that outranked them.
 
 Pass `--debug` to a Cloud resource command to print the resolved credential source (and the API URL) to stderr before the command runs. This works with and without `--json`.
 
+Cloud notices on stderr are best-effort: if their reader goes away, the operation still runs and reports its outcome through the exit status.
+
 ```bash
 clickhousectl cloud --debug service list
 # [debug] auth source: credentials file (.clickhouse/credentials.json)
--- a/crates/clickhousectl/src/cloud/auth.rs
+++ b/crates/clickhousectl/src/cloud/auth.rs
@@ -1,4 +1,5 @@
 use crate::cloud::credentials;
+use crate::cloud::output::eprint_line;
 use crate::cloud::{
     AuthSource, dotenv_env_provenance, env_cred_presence, resolve_active_auth_source,
 };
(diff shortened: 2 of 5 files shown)


iskakaushik (approved):
LGTM

PR #770, opened 2026-09-08
fix(org): require complete private endpoint removal identities

### LINKED ISSUE #679: cloud org update --remove-private-endpoint silently defaults provider/region

`parse_org_private_endpoint_remove` (`crates/clickhousectl/src/cloud/organizations.rs`, ~line 312) initialises `cloud_provider` and `region` from `Default::default()`. Those defaults are `gcp` and `ap-northeast-1` (`crates/clickhouse-cloud-api/src/models/organization_private_endpoints.rs`), and both fields are non-`Option`, so omitting either segment of `--remove-private-endpoint` patches the wrong provider/region instead of failing.

Help now warns that both segments are required; the parser should reject a partial value as a usage error.

sdairs:
## Summary

- require `cloud-provider` and `region` in every `cloud org update --remove-private-endpoint` value
- reject absent, empty, and whitespace-only identity fields during clap parsing, before credentials or HTTP work
- remove the generated provider and region defaults from the internal parser and document the required fields

## Validation

- `CARGO_TARGET_DIR=/Users/al/ch/ai/clickhousectl/target cargo test -p clickhousectl cloud::organizations::tests` (33 passed)
- `CARGO_TARGET_DIR=/Users/al/ch/ai/clickhousectl/target cargo test -p clickhousectl --test cli_request_shape_test organization_update_rejects_incomplete_endpoint_removals_before_http -- --exact` (1 passed)

Closes #679

### DIFF
--- a/README.md
+++ b/README.md
@@ -730,6 +730,10 @@ BYOC create, update, and delete require API key authentication. Update requires
 `--display-name`, so it cannot send an empty/no-op patch. The API has no separate
 BYOC list command; `cloud org get` returns the organization's `byocConfig` entries.
 
+Each `cloud org update --remove-private-endpoint` value must include both
+`cloud-provider` and `region`; incomplete endpoint identities are rejected before
+any request is sent.
+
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #771, opened 2026-09-08
fix(cloud): classify service delete conflicts structurally

### LINKED ISSUE #673: `service_delete_error` sniffs the CONFLICT message prefix and drops the failure classification
## Summary

`service_delete_error` in `crates/clickhousectl/src/cloud/services.rs` (around line 1593) decides behaviour from response prose (`message.starts_with("CONFLICT:")`) and rebuilds the error with `CloudError::new(...)`, which drops the `failure: Option<ApiFailure>` attached at the `CloudClient` conversion boundary.

Both are things CLAUDE.md rules out: never derive a category from message text, and a boundary that rewrites a message must carry `failure` across (`..error` in a struct literal or `with_failure`).

## Expected behavior

- Branch on the library error structurally (`Error::Api { status: 409, .. }`), not on the `CONFLICT:` prefix.
- Preserve `failure` (and `kind`) when the message is rewritten, so telemetry for a refused delete keeps its `http_4xx`/409 classification.
- Unit test pinning that the rewritten error still carries the classification.

## Where found

Spotted during review of PR #672 (issue #666), which introduced `convert_error_for_lookup` next to `convert_error_for_organization`; the same `..error` pattern used there applies here. Pre-existing, not introduced by that PR.

sdairs:
Fixes #673.

Service deletion now recognizes conflicts from the typed HTTP 409 status before converting the library error, so the guidance no longer depends on API response prose. The rewritten error keeps its converted kind, structured details, and failure classification, includes the original API response, and only suggests force deletion conditionally when the service is running.

Validation:
- `cargo fmt --all`
- `CARGO_TARGET_DIR=/private/tmp/chctl-050-673-target cargo test -p clickhousectl service_delete_error`
- `CARGO_TARGET_DIR=/private/tmp/chctl-050-673-target cargo test -p clickhousectl service_delete_conflict_preserves_converted_error_metadata`
- `CARGO_TARGET_DIR=/private/tmp/chctl-050-673-target cargo test -p clickhousectl --test cli_request_shape_test service_delete_conflict_suggests_force_without_inspecting_prose`

### DIFF
--- a/README.md
+++ b/README.md
@@ -929,6 +929,9 @@ clickhousectl cloud service delete <service-id>
 clickhousectl cloud service delete <service-id> --force
 ```
 
+If deletion returns a conflict, the CLI includes the API response and suggests `--force` only
+conditionally: use it when the service is running; resolve other reported conflicts before retrying.
+
 IP allowlist flags accept `IP_OR_CIDR` or `IP_OR_CIDR=DESCRIPTION`. The `=`
 delimiter is safe with IPv6; quote entries whose descriptions contain spaces.
 
--- a/crates/clickhousectl/src/cloud/services.rs
+++ b/crates/clickhousectl/src/cloud/services.rs
@@ -2572,15 +2572,30 @@ impl StopPollProgress {
     }
 }
 
-fn service_delete_error(error: CloudError, force: bool, service_id: &str) -> CloudError {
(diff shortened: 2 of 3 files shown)


iskakaushik (approved):
LGTM

PR #772, opened 2026-09-08
fix(api): preserve fractional Postgres slow-query durations

### LINKED ISSUE #758: Postgres slow-query list/detail fails on fractional duration values
Postgres `slow-queries list` and detail requests fail when live responses contain fractional durations. This blocks the new slow-query feature even though the API returned valid JSON.

Against an instance with slow-query patterns in the selected time range:

```sh
clickhousectl cloud postgres slow-queries list "$PG_SERVICE_ID" \
  --from-date 2026-09-05T00:00:00.000Z \
  --to-date 2026-09-06T16:00:00.000Z --limit 10 --json
```

Exit 1:

```text
JSON error: invalid type: floating point `1190.8419405320817`, expected i64
```

The number varied across repeated live requests. A direct read-only API response contained fractional `avgDurationUs`; detail lookup also failed on a fractional duration. No SQL text or account identifiers are needed to reproduce the type mismatch.

The [Rust model](https://github.com/ClickHouse/clickhousectl/blob/eac19cf1d7ed74fcee2c02c8f6cf79968073d001/crates/clickhouse-cloud-api/src/models/postgres.rs#L1177) uses `Option<i64>` for these values. The checked-in OpenAPI also declares integers, so static drift checks pass while live data cannot deserialize.

### Resolution and acceptance

- Reconcile the live response, OpenAPI schema and library numeric types; preserve fractional duration values rather than silently truncating them.
- Add representative fractional average/percentile response fixtures for list and detail, and verify CLI JSON and human rendering.
- Keep genuinely integral fields integral; do not widen every numeric field indiscriminately.

Priority: high. Related feature: #585 / PR #734; API drift work: #522. Found in the 2026-09-06 live release audit of PR #756 at `eac19cf1d7ed74fcee2c02c8f6cf79968073d001` (release build, package version 0.4.2).

sdairs:
Postgres slow-query list and detail responses fail to deserialize when the API returns fractional microseconds, such as `avgDurationUs: 1190.8419405320817`. Represent the six aggregate `*DurationUs` measurements and execution `durationUs` as `Option<f64>` so callers retain these values. Call, error, row, block, byte and other integral counters keep their existing types; missing/null durations remain absent.

The public OpenAPI fetched on 2026-09-08 still declares these durations as integers. Leave that snapshot untouched and document seven exact response-only runtime exceptions, backed by the live evidence in #758. The analyzer shares its integer/float inventory with exception validation and reports stale entries when the upstream type, Rust field/type, or response reachability changes. Request-side fields cannot use these exceptions.

Validation:
- `cargo fmt --all`
- `cargo test -p clickhouse-cloud-api -p clickhouse-openapi-analyzer` (offline suites; explicitly ignored live suites remain ignored)
- Python script suite: 91 tests pass (per-process `commit.gpgsign=false` for temporary Git fixtures).
- Model fixtures cover fractional list/detail aggregates and executions, integer inputs, missing/null responses and counts above 2^53.
- Analyzer fixtures cover narrow matching and stale entries after schema/field/type/reachability changes, including request-only and shared request/response types. Existing integer policy and snapshot coverage tests pass.
- Analysis against the freshly fetched public spec has only the pre-existing unrelated `organization_prometheus_discovery_get` beta marker finding; no duration exception is stale.

Refs #758. CLI JSON/human regression coverage is a separate stacked PR.

### DIFF
--- a/crates/clickhouse-cloud-api/AGENTS.md
+++ b/crates/clickhouse-cloud-api/AGENTS.md
@@ -79,7 +79,8 @@ reachability from return types across the client module tree, so a newly wired o
 - `models_carry_no_serde_default` — via the analyzer's `model_fields_with_serde_default()`.
 - `scim_models_are_outside_the_response_tree` — the 40 `Scim*` schemas have no spec path and no `Client` method,
   so they are legitimately strict; the test fails if one becomes response-reachable.
-- `integer_schema_fields_are_not_typed_as_float` — integer schemas do not use floating-point Rust fields.
+- `integer_schema_fields_are_not_typed_as_float` — integer schemas do not use floating-point Rust fields, except
+  verified response-only runtime divergences in `fractional_response_exemptions`.
 
 Scope enforcement to the response tree, never to "every model type": operation-unreferenced and request-only
 schemas resolve in request position, so making them all-`Option` reports genuine `FieldOptionalityMismatch` drift.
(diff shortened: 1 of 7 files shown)


iskakaushik (approved):
LGTM

PR #774, opened 2026-09-08
test(local): isolate ports and make readiness checks robust

### LINKED ISSUE #687: Flaky local tests: structured-errors port race and readiness wall-clock
Observed while running `cargo test -p clickhousectl` in parallel worktrees during the #678 docs audit (no code changed):

- `tests/local_structured_errors_test.rs::version_port_and_startup_failures_have_typed_safe_shapes` fails on clean HEAD in a loaded environment; it races on a real port and leaves an untracked `crates/clickhousectl/.clickhouse/` directory behind (a `.metadata.lock` nearly got committed).
- `tests/local_postgres_readiness_test.rs::wall_clock_timeout_fails_and_rolls_back_fresh_data` asserts `elapsed < 4s` and fails under build load.

Both pass on an idle machine. Consider a free-port helper, a temp `.clickhouse` dir, and a looser or mocked clock. Also consider adding `crates/clickhousectl/.clickhouse/` to `.gitignore`.

sdairs:
The local structured-error fixtures selected ephemeral ports and released them before launching the CLI, allowing another test process to claim them. Fake ClickHouse children now use automatic port selection; the occupied-port case retains its listener through the assertion.

The Postgres readiness test measured time spent waiting for the shared fixture lock as part of its four-second deadline. It now checks the timeout error and the fake Docker start/inspect/removal sequence, including exactly one rollback and removal of fresh data and metadata. A separate 30-second watchdog starts after the fixture lock is acquired and kills/reaps a hung child while preserving captured diagnostics.

Temporary project and home directories, plus the existing `/crates/clickhousectl/.clickhouse/` ignore entry, already cover the reported state leakage. No production behavior changes.

Validation:
- `cargo fmt --all`
- Both focused integration binaries: 22 tests passed.
- Three concurrent copies of each focused binary: all six runs passed (66 tests).
- Focused Clippy with `-D warnings` passed.
- `git diff --check` and nested fixture `git check-ignore` passed.

Closes #687.

### DIFF
--- a/crates/clickhousectl/tests/local_postgres_readiness_test.rs
+++ b/crates/clickhousectl/tests/local_postgres_readiness_test.rs
@@ -6,7 +6,7 @@ use std::io::{ErrorKind, Read, Write};
 use std::net::Shutdown;
 use std::os::unix::net::{UnixListener, UnixStream};
 use std::path::{Path, PathBuf};
-use std::process::{Command, Output};
+use std::process::{Command, Output, Stdio};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::thread::{self, JoinHandle};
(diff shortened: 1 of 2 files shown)


iskakaushik (approved):
LGTM

PR #775, opened 2026-09-08
fix(cli): expose and validate canonical closed-set values

### LINKED ISSUE #684: Runtime-validated enum flags exit 1 instead of a clap usage error

Several flags with a closed value set are validated after parsing, so an invalid value exits 1 (error) instead of 2 (usage), and clap cannot render `[possible values: ...]` for them:
- `cloud postgres create --provider` (`PgProvider::VALUES`); compare `--pg-version`/`--ha-type` which use clap parsers and exit 2
- `cloud key create/update --state` (enabled/disabled)
- `skills --agent` (17 agent names, now listed in the CONTEXT FOR AGENTS block instead)

Moving them to `value_parser` is a small behaviour change (exit code 1 -> 2), so it was kept out of the docs PR.

sdairs:
Managed Postgres providers, API key states and Skills agent names are validated during clap parsing, returning usage exit 2 before networking.

Every ClickPipe `--auth` flag advertises canonical possible values while delegating validation to its existing parser, preserving accepted modes and aliases. Credential inference and requirements are unchanged.

Parser and subprocess tests cover closed-set values and usage errors before HTTP. All eleven ClickPipe auth declarations expose validated canonical choices.

Refs #841 (choice visibility only; keep its remaining auth behavior scope open).

Fixes #684

Final combined revision `82b374816a917cbac3a68194659beaaa5c7c9087` passed formatting, default CLI Clippy/tests, telemetry-disabled check/all-target Clippy, API/analyzer all-target Clippy/tests, and Python/classifier checks (1,822 CLI, 639 library/doc, and 91 Python tests). Required Cloud checks on the final PR heads remain pending; historical runs are not substituted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1259,6 +1259,8 @@ generate-backup-bucket-json | clickhousectl cloud backup bucket create <service-
 
 ### ClickPipes
 
+Every ClickPipe `--auth` flag lists its accepted canonical modes in `--help`.
+
 Manage ClickPipes for ingesting data into ClickHouse Cloud from external sources. See the [ClickPipes documentation](https://clickhouse.com/docs/integrations/clickpipes).
 
 ```bash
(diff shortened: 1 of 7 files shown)


iskakaushik (approved):
LGTM

PR #776, opened 2026-09-08
fix(clickpipes): reject source port zero

### LINKED ISSUE #685: MySQL ClickPipe --port accepts 0

`cloud clickpipe create mysql --port` is a bare `u16`, so `--port 0` parses. The Postgres source range-checks `1..=65535` (twice). Add the same `value_parser` to MySQL and any other source ports that lack it.

sdairs:
`cloud clickpipe create mysql --port 0` now fails during clap parsing with usage exit code 2. The MySQL source uses the same 1–65535 port range as PostgreSQL; its default remains 3306. Auditing the other ClickPipe source flags found no other standalone port flag missing validation.

Adds parser coverage for the default, both valid boundaries, zero, and overflow, and documents the source port range in README.

Validation: `cargo fmt --all`; `cargo test -p clickhousectl --bin clickhousectl cloud::clickpipes::tests`.

Closes #685.

Typed update JSON follow-up: #813 addresses the QA comment on #685 by validating MySQL and Postgres update ports before HTTP, including file/stdin input and partial-update omission coverage. It is appended to the same PR stack (#769); this PR continues to cover the create-flag path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2220,6 +2220,10 @@ checked before the request is sent.
 
 Use `clickhousectl cloud clickpipe create <source> --help` for the full list of options per source type.
 
+PostgreSQL and MySQL source `--port` values must be between 1 and 65535
+(defaults: 5432 and 3306, respectively). Values outside this range are usage
+errors (exit code 2) before any request is made.
+
 #### MySQL ClickPipe authentication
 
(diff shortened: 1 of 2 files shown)


iskakaushik (approved):
LGTM

Issue #757, opened 2026-09-06
v0.5.0 release scope and acceptance checklist


sdairs:
## Release objective

Ship chctl 0.5.0 with the Cloud CLI coverage planned in #699, plus the remaining correctness, local-development, output-contract, and CLI consistency work below.

This is the overall release scope tracker. #699 remains the detailed Cloud coverage plan and issue-to-PR matrix; individual issues remain the implementation owners. A submitted PR is not completed release work: track completion when the change is merged and validated.

Scope agreed on 2026-09-06: include all items from the remaining-issue review except #129 (locally built binaries) and #206 (live Pub/Sub E2E). The previously optional and deferred items listed below are now included in 0.5.0.

## Existing Cloud scope

- [ ] Complete and merge #699's 37 implementation issues and API baseline #522, including the separate library prerequisites and follow-up fixes. The recorded 53-layer stack runs from #700 through #754; use #699's issue-to-PR matrix rather than duplicating it here.
- [ ] Land the live CDC fixture repair #747 / #750 as part of that stack.
- [ ] Land README branding and product overview #755 / #756.

### Correctness and release reliability

- [ ] #677 — Prevent closed-stderr credential/debug notices from panicking before Cloud command dispatch; add subprocess coverage proving the command still runs.
- [ ] #679 — Reject private-endpoint removal inputs missing provider or region instead of substituting defaults; fail before sending a request.
- [ ] #253 — Make the native-argument boundary explicit for both local clients. Require `--` for passthrough so unknown options cannot swallow later wrapper flags or change connection selection. Document compatibility changes and test both clients.
- [ ] #684 — Move closed-set flag validation to clap so invalid values produce usage exit 2 and discoverable possible values before networking.
- [ ] #685 — Reject MySQL ClickPipe port zero and audit other source-port parsers for the same omission.
- [ ] #673 — Classify service-delete failures structurally and preserve failure metadata when rewriting their messages.
- [ ] #676 — Bound generated integration resource names while preserving uniqueness; make manual Cloud integration dispatch reliable. Record the CLI name-length validation decision separately from any broader library constraint work.
- [ ] #687 — Isolate local test ports and project state and remove fragile wall-clock assumptions so the release test suite is reliable under load.
- [ ] #680 — Return the correct running-Postgres removal error with an actionable stop command. Track the issue's additional resume-JSON suggestion separately rather than letting it obscure the removal fix.

### Output contracts and CLI consistency

[...]

PR #778, opened 2026-09-08
fix(postgres): honor JSON mode for config output

### LINKED ISSUE #683: cloud postgres config get ignores --json

`postgres_config_get` (`crates/clickhousectl/src/cloud/postgres.rs` ~line 1207) takes `_json` and always prints JSON, while `--json` is still listed on the screen. Either honour the flag with a human view via `print_human`, or hide the flag on this command. Help now states that the command always prints JSON.

sdairs:
`cloud postgres config get` previously always printed JSON, ignoring its output mode. It now uses the shared human renderer by default and preserves the configuration JSON when `--json` is passed or a coding agent is detected. Help and README examples use `config get --json` when saving a document for `config replace`.

Validation: `cargo fmt --all`; `cargo test -p clickhousectl --test cli_request_shape_test config_get` (2 passed, using loopback wiremock fixtures). The regression checks human fields, explicit and agent JSON values, authenticated GET requests, and absence of query parameters or bodies. Existing sparse-response JSON coverage passes. Full combined-release checks are tracked in #757.

Closes #683.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1143,12 +1143,13 @@ clickhousectl cloud postgres delete <pg-id>
 clickhousectl cloud postgres certs get <pg-id>                   # raw PEM to stdout
 clickhousectl cloud postgres certs get <pg-id> --output ca.pem   # file (mode 0600 on unix)
 
-# Runtime configuration (`config get` always prints JSON; --json changes nothing)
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #780, opened 2026-09-08
fix(api): send and decode service settings as objects

### LINKED ISSUE #759: Service settings updates send a string where the live API requires an object
`cloud service settings set` is rejected by the live API because the client sends `settings` as a JSON string and the endpoint requires an object. This reproduced on both an existing service and a newly provisioned disposable service.

Use a disposable service with a supported `compatibility` setting:

```sh
clickhousectl cloud service settings set "$SERVICE_ID" \
  --setting 'compatibility="26.2"' --json
```

Numeric assignment `--setting compatibility=26.2` and a `--settings-file` containing `{"compatibility":"26.2"}` also failed. All returned exit 1:

```text
BAD_REQUEST: settings must be a non-empty object
```

Settings list/schema/get succeeded, and read-back stayed unchanged after the rejected writes.

The captured request shape is:

```json
{"settings":"{\"compatibility\":\"26.2\"}"}
```

The [CLI builder](https://github.com/ClickHouse/clickhousectl/blob/eac19cf1d7ed74fcee2c02c8f6cf79968073d001/crates/clickhousectl/src/cloud/services.rs#L1719) serializes the map to a string; the [request model](https://github.com/ClickHouse/clickhousectl/blob/eac19cf1d7ed74fcee2c02c8f6cf79968073d001/crates/clickhouse-cloud-api/src/models/services.rs#L1706) and checked-in OpenAPI agree on `String`. Existing wire tests assert that same string representation, so they do not establish compatibility with the live server.

### Resolution and acceptance

- Reconcile the endpoint and OpenAPI/library contract. The live error indicates an object is required; confirm the accepted body before changing the client.
- Verify both repeated `--setting` and file inputs against the corrected wire contract.
- Verify a real supported setting update and read-back on a disposable service, plus malformed-input behavior.
- Update the API library and CLI exposure in their respective changes, following repository practice.

Priority: high. Related feature: #574 / PR #746; API drift work: #522. Found in the 2026-09-06 live release audit of PR #756 at `eac19cf1d7ed74fcee2c02c8f6cf79968073d001` (release build, package version 0.4.2).

sdairs:
Service settings updates currently encode the `settings` object as a string, which the live Cloud API rejects with `settings must be a non-empty object`. Model the request and update response as maps of setting names to JSON values; preserve value types and reject invalid legacy encoded input before HTTP.

The live endpoint was verified to accept and return objects, while the published OpenAPI still declares both fields as strings. Keep the downloaded spec unchanged and document the verified discrepancy (#759). The default request uses a map; explicit `ServiceClickhouseSettingsPatchRequest<String>` callers are normalized to the object wire format. Two mechanical CLI type annotations keep this API prerequisite buildable until the separate CLI adaptation adopts the map directly. Existing CLI wire fixtures and response expectations are updated here because the API serializer changes their transport immediately.

Validation: library clippy with all targets; 168 client tests, 315 model tests, and six snapshot/spec tests passed (one live-spec test ignored); CLI `cargo check` passed at this prerequisite. Both existing settings-set wire tests also pass on the API prerequisite alone (`cargo test -p clickhousectl --test cli_request_shape_test service_settings_set`). Live PATCH of `compatibility="26.2"` returned HTTP 200 with an object response, and GET readback matched on a disposable service. The service was subsequently removed after CLI validation.

Refs #759. Followed by the CLI adaptation branch `fix/050-759-settings-cli`.


iskakaushik (approved):
LGTM

PR #781, opened 2026-09-08
fix(cli): preserve service settings types and integer precision

### LINKED ISSUE #759: Service settings updates send a string where the live API requires an object

sdairs:
`cloud service settings set` sends a settings object instead of a JSON-encoded string. It supports repeated JSON-valued assignments and file/stdin maps; malformed documents, request wrappers and duplicate settings fail before HTTP.

Settings assignments, files and stdin preserve exact signed and unsigned JSON integers through `u64::MAX`. Floating representations caused by overflow, decimals or exponents are rejected before HTTP; string values remain available when quoted. File read errors include the supplied path and occur before organization discovery. The precision guard is self-contained here and does not depend on descendant #822.

Wire tests cover integer limits and rejected overflow/exponent inputs for assignments, files and stdin, plus missing/unreadable input before organization discovery. API-library prerequisites remain separate.

Refs #832 and #849 (settings numeric input and file diagnostics only).

Closes #759.

Final combined revision `82b374816a917cbac3a68194659beaaa5c7c9087` passed formatting, default CLI Clippy/tests, telemetry-disabled check/all-target Clippy, API/analyzer all-target Clippy/tests, and Python/classifier checks (1,822 CLI, 639 library/doc, and 91 Python tests). Required Cloud checks on the final PR heads remain pending; historical runs are not substituted.


iskakaushik (approved):
LGTM

PR #782, opened 2026-09-08
fix(org): support consistent organization selection

### LINKED ISSUE #686: Org ID is a positional on org get/update but --org-id everywhere else
Found during the #678 docs audit (UX consistency, needs a decision).

`cloud org get` and `cloud org update` take a required positional `<ORG_ID>`. `org prometheus`, `org usage` and every member/invitation/key/activity/backup/service command take an optional, auto-detected `--org-id`. `org prometheus` and `org usage` additionally accept a hidden legacy positional. Help cannot hide the inconsistency; consider accepting `--org-id` (auto-detected) on `org get`/`org update` too, keeping the positional as a hidden alias.

sdairs:
Organization get/update previously required a positional ID while sibling commands accepted optional `--org-id`. Both now accept `--org-id` and use the existing single-organization autodetection when omitted. The legacy positional remains accepted but hidden; providing both selectors fails with usage exit 2 before HTTP, even if the IDs match.

README examples and migration guidance use the flag form. Update request validation still happens before organization discovery.

Validation: organization unit tests (32 passed); three subprocess/wiremock tests covering both explicit selectors, successful autodetection, zero/multiple organizations, matching/differing selector conflicts, authentication, and exact update bodies; `cargo fmt --all`. Full stacked checks are run before publishing.

Closes #686.

### DIFF
--- a/README.md
+++ b/README.md
@@ -696,12 +696,12 @@ Reading a service, Postgres service or organization â€” or deleting a service â€
 
 ```bash
 clickhousectl cloud org list              # List organizations
-clickhousectl cloud org get <org-id>      # Get organization details
+clickhousectl cloud org get --org-id <org-id>      # Get organization details
 clickhousectl cloud org quota list --org-id <org-id>
 clickhousectl cloud org quota get services-per-organization --org-id <org-id>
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #783, opened 2026-09-08
fix(postgres): explain how to stop a running instance before removal

### LINKED ISSUE #680: local postgres remove on a running instance returns the wrong error

`local postgres remove` on a running instance returns `ServerAlreadyRunning` ("Server 'x' is already running", `crates/clickhousectl/src/local/postgres.rs` ~line 1009). The ClickHouse path returns `ServerRunningCannotRemove` with a repair command (`error.rs` ~line 619). The Postgres path should use the same variant so the message says what to do.

Related: `local postgres start` prints its "resuming with stored settings" note to stderr and suppresses it under `--json`, so a JSON caller that passes `--port/--user/--password/--database/-e` on a resume silently gets the stored values. Consider surfacing the resume in the JSON payload.

sdairs:
`local postgres remove` now reports the cannot-remove error for a running instance and gives `clickhousectl local postgres stop <name> --version <major>` as the recovery command. Human and JSON messages agree; JSON retains `server_running`, exit status remains 1, and ClickHouse removal errors keep their existing output.

Regression coverage exercises default and named Postgres instances in human, explicit JSON, and detected-agent modes through a fake Docker API, checking that refusal leaves metadata and data intact and sends no Docker mutations. README documents the recovery step. The separate suggestion about resume JSON output is outside this fix.

Validation: `cargo fmt --all`; 53 local-output unit tests; Postgres readiness, ClickHouse stopped/state-machine, and structured local-error subprocess suites (serial execution).

Fixes #680.


iskakaushik (approved):
LGTM

PR #784, opened 2026-09-08
test(cloud): bound generated integration resource names

### LINKED ISSUE #676: Manual cloud-integration dispatch breaches the 50-character service name cap; consider a client-side guard
## Summary

A `workflow_dispatch` run of `Cloud Integration` (run 33677928044, 2026-09-02, scope `all`, on the stack tip `1c577b3`) failed its service suite at the rename step:

```
FAIL  [BLOCKING] rename service
  error: API error (status 400): BAD_REQUEST: request body.name: 'clickhousectl-it-manual-33677928044-1c577b3-updated'
```

The Cloud API caps service names at 50 characters (`ServicePostRequest.name` / `ServicePatchRequest.name`, `maxLength: 50`, "Alphanumerical string with whitespaces up to 50 characters" in the vendored spec). The harness builds the rename target as `clickhousectl-it-<label>-<sha7>-updated`, and the workflow sets the manual label to `manual-<github.run_id>`. GitHub run IDs are now 11 digits, so the name is 51 characters. The create passed (43 characters); only the `-updated` rename crossed the limit.

The nightly (`nightly`) and PR (`pr-<n>`) labels are short, so those runs never hit it. Rerunning the same suite from the same commit with `CLICKHOUSE_CLOUD_TEST_RUN_LABEL=lt2` passed all 55 steps, confirming this is a harness artefact, not a product defect. Every other suite in that run passed.

## Where

- `.github/workflows/cloud-integration.yml`, "Resolve test run label": `manual-${{ github.run_id }}`.
- `crates/clickhouse-cloud-api/tests/common/support.rs`: `service_name()` / `updated_service_name()` (and the `pg`, `pgrr`, `cp-pg`, `cp-e2e` variants, which sit at 47 to 49 characters under the manual label and are one digit away from the same failure).

## Definite fix

- Shorten the manual label, for example `m<github.run_number>` (run numbers are 3 to 4 digits), and/or
- Clamp every harness-generated name to the API's cap so no label can push it over. A unit test on the name builders should pin the bound.

## Open question: should the client guard it too?

The cap is the API's, but a user who types a 51-character `--name` today gets the same opaque `BAD_REQUEST: request body.name: '...'` back after a network round trip. Options, from least to most invasive:

1. **CLI pre-flight only**: `cloud service create --name` / `update --name` (and the Postgres equivalents) reject names over the cap as a usage error (exit 2) with a message naming the limit, before any request is made. Consistent with the existing exit-2-before-network pattern for invalid ClickPipe inputs.
2. **Library-level typed constraint**: encode `maxLength` on the request field. This is a bigger step: the analyzer does not currently inventory string length constraints (11 `maxLength` occurrences in the spec, none checked), so a library constraint would need analyzer support to stay honest against drift, and a stale limit would reject names the server accepts.

Recommendation: do the definite fix now; do option 1 in the CLI, sourcing the number from the spec constant rather than hard-coding it in prose, and leave option 2 unless the analyzer gains constraint checking.

## Acceptance criteria

- `workflow_dispatch` runs of `Cloud Integration` with scope `all` pass the service suite.
- Harness name builders can never emit more than 50 characters regardless of label; pinned by a unit test.
- Decision recorded on the client-side guard; if adopted, over-long `--name` values exit 2 before any request, with a clap parse test and a wiremock test asserting no request is sent.

Found while running the full cloud suite over the top of stack #640 (PR #675).

sdairs:
Manual Cloud Integration dispatches could create a service successfully and then fail its rename because appending `-updated` exceeded the API’s 50-character name limit. Route all existing harness resource-name builders through a bounded helper that reserves the suffix and retains a deterministic 64-bit fingerprint of the full run ID when shortening. Short names remain unchanged, long labels with identical prefixes stay distinct, and full run IDs remain in tags.

Fixes #676.

Client-side guard decision: leave CLI name validation unchanged and keep the API authoritative. Broader request-constraint enforcement belongs in separate work; this fix adds no library constraint or workflow-label change.

Validation: direct builder tests cover the failing manual label, very long and Unicode labels, distinct resource purposes and run tails/SHAs, determinism, and the exact 50-character boundary. `cargo fmt --all`, library/analyzer all-target clippy with warnings denied, library/analyzer tests, and all 91 Python tests pass. The library tests needed localhost mock-server access outside the sandbox; Python temporary-repo tests ran with commit signing disabled in their environment. Full manual Cloud Integration (`all`) acceptance will run on the final stack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2785,6 +2785,9 @@ Cloud API integration is tested against a real ClickHouse Cloud workspace via th
 - [`tests/integration_org_test.rs`](crates/clickhouse-cloud-api/tests/integration_org_test.rs) — org-scoped endpoints (members, invitations, roles, activity, prometheus, private endpoint config)
 - [`tests/clickpipes/`](crates/clickhouse-cloud-api/tests/clickpipes/) — ClickPipes E2E; only Postgres CDC runs in CI
 
+Generated service and replica names stay within 50 characters, including rename suffixes.
+Long run labels are shortened with a hash of the full run ID; resource tags retain the full ID.
+
 Required environment variables:
 
 ```bash
--- a/crates/clickhouse-cloud-api/tests/common/support.rs
+++ b/crates/clickhouse-cloud-api/tests/common/support.rs
@@ -105,12 +105,35 @@ impl TestContext {
         })
     }
 
+    /// Keep generated resource names within the Cloud service name limit. Hash
+    /// the full run ID before shortening, so a shared label prefix does not
+    /// collapse distinct runs. Reserve the suffix before trimming the label.
+    fn resource_name(&self, prefix: &str, suffix: &str) -> String {
+        const MAX_NAME_CHARS: usize = 50;
+        let name = format!("{prefix}{}{suffix}", self.run_id);
+        if name.chars().count() <= MAX_NAME_CHARS {
+            return name;
+        }
+        // Fixed FNV-1a keeps names reproducible across processes and Rust
+        // versions; this fingerprint is for collision avoidance, not security.
+        let hash = self
+            .run_id
+            .bytes()
+            .fold(0xcbf29ce484222325_u64, |hash, byte| {
+                (hash ^ u64::from(byte)).wrapping_mul(0x100000001b3)
+            });
+        let fingerprint = format!("{hash:016x}");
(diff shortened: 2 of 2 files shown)


iskakaushik (approved):
LGTM

PR #785, opened 2026-09-08
fix(local): require separator before native client arguments

### LINKED ISSUE #253: fail on unknown flags?
`chctl local server start -v 26.6.1.811 --config default`

^ this command work but ignores --config  because it isnt a valid flag. Makes for confusing behaviour because the user expect the config to respected. Should we fail on unknown flags?

sdairs:
Require `--` before native arguments to both `local client` and `local postgres client`. Previously an unknown option could start passthrough and consume later wrapper selectors such as `--name` or `--host`; it now produces usage exit 2 before resolving a connection or starting a child. Arguments after the separator retain their order and literal values, including flags that share wrapper names.

README documents the compatibility change and both help screens expose the separator. Coverage includes unknown native arguments before and after wrapper selectors, and real-binary/fake-child checks for preserved empty arguments, spaces, newlines, and native flags.

Validation:
- `cargo fmt --all`
- Local clap tests: 45 passed.
- Telemetry unit tests: 64 passed.
- Local client selector subprocess tests: 15 passed.
- Postgres client input tests: 6 passed; 1 real-Docker test ignored. Mock Unix-socket tests passed with sandbox escalation.
- Both client help screens inspected; `git diff --check` clean.

Closes #253. Implements the client boundary scope accepted in #757.

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,6 +407,8 @@ clickhousectl local client --host remote-host --version 26.8.1.1760  # Use an in
 clickhousectl local client -- --format Pretty        # Extra clickhouse-client args after --
 ```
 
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #786, opened 2026-09-08
fix(api): preserve numeric service setting values

### LINKED ISSUE #779: Service setting get cannot deserialize numeric values
After a supported numeric service-setting update, `cloud service settings get` fails to deserialize the successful API response.

Observed during live validation of #759 on a disposable service:

```sh
clickhousectl cloud service settings set "$SERVICE_ID" --setting 'max_query_size=262146' --json
clickhousectl cloud service settings get "$SERVICE_ID" max_query_size --json
```

The update succeeds. The read exits 1 with `JSON error: invalid type: integer 262146, expected a string`. A direct GET of `/v1/organizations/{org}/services/{service}/clickhouseSettings/max_query_size` returns HTTP 200 and `{"result":{"name":"max_query_size","value":262146},"status":200}` (request identifier omitted).

`ServiceClickhouseSetting.value` in `crates/clickhouse-cloud-api/src/models/services.rs` is `Option<String>`. The current public OpenAPI also declares string. The list response reuses the same model; list failure has not independently been reproduced.

Acceptance:
- Preserve observed numeric values and existing string values without coercion or truncation; tolerate missing/null response fields per library policy.
- Keep the live/schema divergence explicit and narrowly scoped; do not hand-edit OpenAPI.
- Add model and client coverage for numeric/string values, plus CLI JSON/human regression for get and representative list fixtures.
- Verify a real numeric setting update and read-back on an owned disposable service, then clean it up.
- Keep library changes separate from CLI tests/docs, and update README for user-visible output types.

Related: #759, release tracker #757. This is separate from the settings PATCH object-shape fix.

sdairs:
A successful numeric ClickHouse setting update could make the next GET fail with “expected a string.” Change `ServiceClickhouseSetting.value` to `Option<serde_json::Value>` so GET and list preserve the API’s JSON values, including numbers and existing strings. Missing/null fields remain absent. The schema divergence is documented at the field and in README; the OpenAPI snapshot is unchanged.

Migrate the library’s settings lifecycle test and cleanup registry to retain the original JSON type for restoration. Model tests cover integer precision, floats, strings, booleans, future objects, missing/null fields, and list round trips; client tests cover GET and list over authenticated HTTP fixtures.

Validation: focused model/client tests, vendored spec coverage, and library/analyzer all-target clippy pass. Live validation on a new owned service updated `max_query_size` to 262146 and read it back as a number using both GET and list, alongside string `compatibility` = "26.2". Its configurable schema advertises integer/string settings; boolean/object test cases cover forward compatibility, not additional claims about the live schema.

Refs #779. CLI rendering and regression tests are in the stacked CLI follow-up.

Disposable service cleanup completed successfully; subsequent GET returned NOT_FOUND.


iskakaushik (approved):
LGTM

PR #787, opened 2026-09-08
fix(cli): render typed service setting values

### LINKED ISSUE #779: Service setting get cannot deserialize numeric values

sdairs:
Keep human service-settings list output readable after the response model accepts JSON values: strings display without JSON quotes, while numbers and booleans retain their normal representation. GET and list JSON output preserve the value types returned by the API.

Add subprocess regressions for numeric/string/boolean GET in human and JSON modes, plus mixed-type lists with missing/null fields. Document the output behavior in README.

Validation: all 9 service-settings subprocess tests pass. Default CLI clippy, no-default-features check, and no-default-features all-target clippy pass. Live validation on an owned disposable service confirmed numeric `max_query_size` = 262146 and string `compatibility` = "26.2" in GET/list JSON and correct human numeric GET/mixed list output.

Closes #779. Depends on the stacked API model fix.

Disposable service cleanup completed successfully; subsequent GET returned NOT_FOUND.


iskakaushik (approved):
LGTM

PR #789, opened 2026-09-08
test(local): define the native client output contract

### LINKED ISSUE #468: Define the JSON output contract for local client
## Problem

`local` advertises `--json`, and coding-agent detection automatically enables JSON mode, but `local client` drops the resolved JSON boolean and forwards identical native argv.

On HEAD `93b3dd7`, an isolated fake binary receives the same arguments for all three forms:

```console
clickhousectl local client --host remote --query Q
clickhousectl local --json client --host remote --query Q
clickhousectl local client --json --host remote --query Q
```

The README limits `--json` to commands that return structured data and says protocol-oriented commands retain natural output, but it does not name `local client`. The command help also does not explain that explicit JSON and automatic agent JSON do not select a native output format.

## Expected behavior

Define `local client` as an explicit protocol-output exception, or define JSON behavior for non-interactive queries without overriding SQL `FORMAT` clauses or user-supplied native format flags. Interactive mode must not imply JSON it cannot guarantee.

## Acceptance criteria

- Help and README state exactly what `--json` and coding-agent detection do for interactive and one-shot local-client modes.
- If JSON is supported, format precedence is defined for SQL `FORMAT`, `--format`, `--output-format`, queries files, and multiple queries.
- If natural output is retained, explicit `--json` is rejected or clearly documented rather than silently suggesting a transformation.
- Fake-child argv and real-client output tests cover human, explicit JSON, detected-agent, interactive, query, query-file, and user-format cases.

sdairs:
`local client` advertised JSON mode without explaining that SQL output remains native. Document the accepted native-output contract in both local client help screens and the README: interactive sessions, queries and query files retain their client's output under `--json` and coding-agent detection. Show native format options after the required `--` separator and leave format precedence to ClickHouse.

Add a fake-child argv matrix for ClickHouse and psql across human, explicit JSON and detected-agent modes, and an opt-in real ClickHouse test covering stdin, queries, multi-statement query files, CSV, JSONEachRow, SQL FORMAT and an actual interactive PTY. The conflicting-format case compares output with the selected native binary directly. Both path classifiers include the new test.

Validation:
- `cargo fmt --all` and focused test Clippy passed.
- Both output-contract tests passed, including real ClickHouse 26.8.1.1760 on a disposable local server. All three wrapper modes reached and exited actual interactive PTY sessions successfully. The disposable server was stopped afterward.
- 45 local clap tests and 91 Python tests passed (Git signing disabled only for temporary test commits).
- Native Postgres output was not exercised against a real server; psql coverage is the isolated fake-child argv matrix. The real ClickHouse test is ignored by default and requires explicit `CHCTL_TEST_NATIVE_BINARY` and `CHCTL_TEST_NATIVE_PORT` pointing at a disposable server. Full stack checks are recorded by the integration run.

Closes #468.


iskakaushik (approved):
LGTM

PR #815, opened 2026-09-09
fix(local): generate fresh names for unnamed Postgres starts

### LINKED ISSUE #801: local postgres: generate a new name when the unnamed default is already running
Found in the 2026-09-08 exploratory QA of the combined PR stack at [`4c17bc55`](https://github.com/ClickHouse/clickhousectl/commit/4c17bc55d89108a0a73579b0655baf8189d97207), top PR #793. Tested the installed release build (package version still 0.4.2). This report records reproduced behavior; no implementation is included.

Severity: medium. Pre-existing in v0.4.2, confirmed by baseline source inspection.

Start a default local Postgres instance in a disposable project, then run another start without a name, using a different free port:

```sh
clickhousectl local --json postgres start --password '<test-password>'
clickhousectl local --json postgres start --password '<test-password>' --port 55490
```

Actual: second command exits 1 with `server_running`, `Server 'default' is already running`.
Expected: generate a new name as documented in start help. Reproduced with automatic and explicit free port selection.

Cause: `default_pg_name_locked` detects the running Postgres default but delegates to the ClickHouse-oriented resolver, whose metadata lookup does not see that Postgres default. See [crates/clickhousectl/src/local/postgres.rs:527](https://github.com/ClickHouse/clickhousectl/blob/4c17bc55d89108a0a73579b0655baf8189d97207/crates/clickhousectl/src/local/postgres.rs#L527) and [crates/clickhousectl/src/local/server.rs:745](https://github.com/ClickHouse/clickhousectl/blob/4c17bc55d89108a0a73579b0655baf8189d97207/crates/clickhousectl/src/local/server.rs#L745).

Suggested fix: generate a collision-free Postgres name under the metadata lock, or make engine-specific lookup explicit. Add a real/default-name selection regression test that proves the second instance gets a distinct name.

## Implementation and validation notes

Reproduce in a disposable project with Docker available. The first command starts the instance named `default`; the second should select a distinct name. Stop/remove only test-owned instances after checking the result. Preserve explicit-name semantics, stopped-instance resume behavior, automatic port selection, and same-name multi-version isolation.

sdairs:
When the default Postgres instance is running, another unnamed start now calls the name generator directly instead of consulting ClickHouse default state. Generated names avoid existing metadata across engines; stopped default instances retain resume behavior.

Regression tests cover absent/stopped/running defaults and metadata collisions. Docker was unavailable, so live container validation was not run.

Validation: focused regressions plus the required formatting, default Clippy/tests, telemetry-disabled check/all-target Clippy, and Python classifier tests on the combined QA stack tip.

Closes #801.

Stack: appended to existing GitHub PR stack #769; leave merging and releasing to Al.

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,6 +578,8 @@ Native `psql` arguments require `--`, with all wrapper selectors before it: `loc
 
 `local postgres client` also retains native psql output for interactive sessions, `--query` and `--queries-file`, regardless of `--json` or coding-agent detection. Put psql output options after `--`, for example `local postgres client --query "SELECT 1" -- --csv`.
 
+When `local postgres start` is run without `--name`, the first instance is named `default`. If that instance is already running, each subsequent unnamed start gets a fresh generated name instead of resuming or colliding with existing Postgres state.
+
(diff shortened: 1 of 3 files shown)


iskakaushik (approved):
LGTM

PR #870, opened 2026-09-12
fix(postgres): preserve tags on name and size updates

### LINKED ISSUE #827: `cloud postgres update` without a tag flag clears all tags (API treats an absent `tags` on PATCH as replace-with-empty)
## Validated scope and current-PR plan — 2026-09-11

This section records the QA review and the current decision to finalize existing PRs without opening new PRs. It supersedes the proposed routing/fix suggestions in the original report below; the original observations are retained. Central plan: #757.

### Disposition

Deferred from this existing-PR finalization pass. Keep the issue open; no new PR is requested now. This is not a resolution, a severity downgrade, or a waiver of release acceptance.

### Validation and corrections

Confirmed outgoing rename PATCH omits tags and does no preliminary GET; the destructive backend effect remains QA evidence. [postgres.rs:1648](https://github.com/ClickHouse/clickhousectl/blob/d2c37f814fff53fe68d636e34103e2d0550bf84d/crates/clickhousectl/src/cloud/postgres.rs#L1648) matches the reported mechanism. Preserve a complete fetched tag snapshot, fail if tags are absent, and account for concurrent updates. Keep explicit clear semantics. No currently open PR owns tag-preserving PATCH behavior.

---

## Original QA report

Found in the 2026-09-11 QA sweep of the combined PR stack at [`d2c37f81`](https://github.com/ClickHouse/clickhousectl/commit/d2c37f814fff53fe68d636e34103e2d0550bf84d), top PR #823, installed release build (package version still 0.4.2). Reproduced behaviour only; no implementation included.

Severity: high; silent metadata loss on every rename or resize.

```sh
clickhousectl cloud postgres update "$PG" --add-tag env=qa --add-tag owner=me | jq -c .tags
# [{"key":"env","value":"qa"},{"key":"owner","value":"me"}]
clickhousectl cloud postgres update "$PG" --size c6gd.large | jq -c .tags
# []
clickhousectl cloud postgres get "$PG" | jq -c .tags
# []
```

README: "`--clear-tags` replaces the tag list with an empty list … omitting all three leaves tags unchanged." That is false against the live API.

The CLI request is correct per the spec — `build_postgres_update_request` passes `tags: None` and `PostgresServicePatchRequest.tags` is `skip_serializing_if = "Option::is_none"`, so no `tags` key is sent. The API treats the absent key as "clear". `cloud service update` did not show this behaviour.

Suggested CLI fix, independent of the backend fix: the handler already GETs the service when `--add-tag`/`--remove-tag` is present; always GET and echo the current tags back when no tag flag is supplied. Correct the README sentence either way. A separate backend issue is needed for the PATCH semantics (to be raised by the maintainers alongside #835).

Acceptance: wiremock test that a `--name`/`--size`-only update sends the current tags; `postgres get` after such an update shows tags unchanged.

sdairs:
Postgres name, size and HA-only updates previously omitted `tags`, which the deployed API treats as clearing the tag list. Every update except explicit `--clear-tags` now fetches and preserves the complete current tag snapshot, applying requested tag additions/removals to it.

Missing/null tag lists, incomplete tag identities and failed reads prevent PATCH. Key-only and empty-string-valued tags retain their distinct representations; explicit clearing still needs no GET. README documents that the API replaces the snapshot, so concurrent tag changes can still race with an update.

Closes #827.

Validation: real-binary wiremock coverage for name/size/HA changes against empty/populated tags, missing/null response fields, failed reads, additions/removals and explicit clearing. Focused merge tests passed. Combined stack validation passed formatting, default CLI Clippy and 1,846 passing CLI tests (four existing ignored), telemetry-disabled check/all-target Clippy, and all 91 Python/classifier tests. The deployed backend effect was established in the issue; this change was verified locally against mocks, without modifying a live service.

Appended in stack #769. No existing stack commits were rewritten.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1210,7 +1210,9 @@ clickhousectl cloud postgres update <pg-id> \
 clickhousectl cloud postgres update <pg-id> --clear-tags
 
 # --clear-tags replaces the tag list with an empty list and conflicts with
-# --add-tag and --remove-tag; omitting all three leaves tags unchanged.
+# --add-tag and --remove-tag; omitting all three preserves the fetched tags.
+# Updates read the complete tag list first (except --clear-tags) and fail if it
+# cannot be read safely. Avoid concurrent tag changes: PATCH replaces that snapshot.
 
 # Delete (works from any state, including running; no stop needed first)
 clickhousectl cloud postgres delete <pg-id>
--- a/crates/clickhousectl/src/cloud/postgres.rs
+++ b/crates/clickhousectl/src/cloud/postgres.rs
@@ -1234,7 +1234,7 @@ fn merge_tags(
     merged
 }
 
-/// Merges `--add-tag`/`--remove-tag` against the tag list a GET returned.
+/// Preserves or merges the tag list a GET returned before every update.
 ///
 /// An omitted `tags` in the response is indistinguishable from a field the API
 /// dropped, so a read-modify-write must not proceed on it: `tags` is replaced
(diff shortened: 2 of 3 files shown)


iskakaushik (approved):
LGTM

PR #898, opened 2026-09-15
feat(local): standardize positional instance names

### LINKED ISSUE #889: Standardize local instance selectors on positional names with hidden --name compatibility
## Problem

The same local instance is addressed differently across commands:

| Family | Start | Stop/remove | Client | Dotenv |
| --- | --- | --- | --- | --- |
| ClickHouse | positional or `--name` | positional or hidden `--name` | `--name` only | `--name` only |
| Postgres | `--name` only | positional only | `--name` only | `--name` only |

For example, `local postgres start dev` fails but `local postgres stop dev` succeeds; the reverse is true for `--name dev`.

Part of the v0.5.0 consistency work in #757.

## Target behavior

Local instances have names rather than alternative resource IDs. Make positional `[NAME]` the advertised form everywhere an instance is selected or started:

```text
local server start [NAME]
local server stop [NAME]
local server remove [NAME]
local server dotenv [NAME]
local client [NAME] [OPTIONS] [-- <ARGS>...]
local postgres start [NAME]
local postgres stop [NAME]
local postgres remove [NAME]
local postgres dotenv [NAME]
local postgres client [NAME] [OPTIONS] [-- <ARGS>...]
```

Keep names optional. This is a syntax standardization; retain existing bare-command defaults, automatic start naming, stop selection, remove safeguards, and Postgres major-version disambiguation.

Support a hidden `--name NAME` compatibility form on all these commands. Preserve existing client `-n` compatibility as hidden too. Reject positional plus flag input, even when the strings match.

## Acceptance

- [ ] All ten commands parse positional names and hidden `--name` consistently; existing short-name aliases remain accepted but unadvertised.
- [ ] No-name behavior is preserved, including generated start names, ClickHouse stop auto-selection, and conservative remove behavior. Do not make bare starts fail or broaden deletion selection.
- [ ] Both clients preserve direct host/port conflicts with managed names and their existing version rules.
- [ ] `--` remains mandatory for native arguments. Tokens after it remain opaque native-client/server arguments and cannot be consumed as the wrapper name or flags.
- [ ] Remove/update the ClickHouse client's custom usage override so positional NAME is actually shown. Normalize all lifecycle/client/dotenv help and README examples.
- [ ] Preserve name validation and prove positional/compatibility forms select the same instance in focused local subprocess tests.
- [ ] Cover conflicts, omitted names, explicit `default`, custom names, direct mode, and passthrough boundaries in clap tests.
- [ ] Run all required CLI gates and affected local integration tests; update classifiers if adding files.

Source: `crates/clickhousectl/src/local/cli.rs`, `src/local/mod.rs`, `src/local/postgres.rs`, and local selector/compatibility subprocess tests.

sdairs:
All ten local instance commands now accept optional positional `[NAME]`: ClickHouse and Postgres start, stop, remove, client, and dotenv. Hidden `--name` compatibility remains available everywhere, including existing client `-n` aliases. Mixing positional and flag selectors is a usage error even when their values match.

Dispatch preserves omitted-name defaults, generated start names, conservative removal, Postgres major-version selection, direct connection constraints, and opaque native arguments after `--`. README examples and connection hints use positional names.

Validation: 272 local unit tests and 38 focused subprocess tests passed (one existing real-Docker test ignored); telemetry-disabled all-target clippy, formatting, and diff checks passed. Full combined-stack CLI gates will be recorded after integration.

Closes #889.
Part of #757.


iskakaushik (approved):
LGTM

PR #875, opened 2026-09-14
fix(clickpipes): inherit schema discovery organization flag

sdairs:
`cloud clickpipe schema-discover <service-id> <source> --org-id <org-id>` previously failed at argument parsing because the organization selector was not inherited by source subcommands. All four sources now accept it before the service ID, between the service ID and source, or after source options, and expose it in their help.

Adds a parsing matrix for all sources and supported positions, structural flag-visibility coverage, and a README note. The cosmetic ordering of shared flags across unrelated cloud commands remains outside this focused fix.

Addresses the functional defect in #851; shared-flag ordering remains open.

Validation: formatting, focused schema-discovery tests (including the existing wiremock case), command-tree/shared-help structural checks, and default CLI clippy passed.

Combined stack `8ecf0d1eab31f15ae1d0702a1e83346000fde95a` passed formatting, default CLI clippy, all CLI tests (1,873 passed; five ignored), telemetry-disabled build and all-target clippy, 91 Python/classifier tests, and diff checks. Per-PR Cloud integration decisions remain separate.

PR #899, opened 2026-09-15
feat(cloud): make organization selection cloud-wide

### LINKED ISSUE #851: Make --org-id a cloud-wide selector and remove legacy positional org IDs
## 0.5.0 scope and current recommendation — 2026-09-15

Make `--org-id` a cloud-wide option, accepted anywhere after `cloud`, and remove the four legacy positional organization-ID forms. This is part of #757. A breaking change to retire those positionals is explicitly accepted for 0.5.0.

This section defines the current implementation scope and supersedes the earlier limited routing/deferral notes below. The original report is preserved. Shared help ordering now has its own owner: #887. ClickPipe source/service positional reordering is #888. Optional organization-name lookup and the broader resource-selector policy are tracked in #891.

### Current behavior

There are 148 `--org-id` declarations. Most are leaf-local; only the UDF, Query API endpoint, and org Prometheus declarations are global within their groups. ClickPipe discovery's selector belongs to its parent, so it is rejected after the source subcommand.

These should all select exactly the same organization:

```text
cloud --org-id ORG service get SERVICE
cloud service --org-id ORG get SERVICE
cloud service get --org-id ORG SERVICE
cloud service get SERVICE --org-id ORG
cloud clickpipe schema-discover kafka SERVICE --org-id ORG <source flags>
```

The last example uses the canonical source-first discovery grammar from #888. The global selector must also work throughout UDF, Query API, ClickStack, Postgres, nested service operations, and org subcommands.

### Positional organization IDs to remove

- `cloud org get ORG`
- `cloud org update ORG ...`
- `cloud org usage ORG ...`
- `cloud org prometheus ORG [discovery]`

Use `--org-id ORG` instead, including in org commands. Organizations are shared scope, so this is a deliberate exception to positional target IDs. Do not leave hidden positional compatibility behind for these four forms; update all old examples, including the org parent help's BYOC lookup hint.

### Implementation

Clap already supports this through `#[arg(long, global = true)]`, as used by cloud credentials, JSON, and debug options. Put the organization selector once on `CloudArgs` and thread that shared value/context through dispatch and the handlers, removing the repeated domain fields rather than maintaining competing sources of truth.

The parser feature is straightforward; migration touches many declarations, dispatch arms, helper/options structs, tests, and examples. No Cloud API change is required for ID selection.

Resolve the org only when an operation needs org scope. Preserve explicit selection, single-organization auto-detection, errors for zero/multiple organizations, and existing auth/write checks. Do not make auth/help/org-list operations perform unnecessary organization lookup merely because the option is global.

"Anywhere" means anywhere in the cloud command hierarchy, after the `cloud` token, just like existing cloud-wide flags. It does not introduce a root-level option before `cloud`.

### Acceptance

- [ ] One canonical cloud-wide declaration, visible consistently wherever inherited and consumed through the shared context.
- [ ] Parsing tests cover before/between/after subcommands and before/after positional IDs at shallow and deep levels, asserting the same resolved CloudArgs selector.
- [ ] Cover all four discovery sources and descendants of existing global-org groups; no accidental field shadowing.
- [ ] Subprocess/mock tests verify the supplied org reaches request paths regardless of option placement and that missing/multiple organizations retain current behavior.
- [ ] Former positional org forms are rejected as usage errors (exit 2); README, help, hints, tests, and release migration notes use the flag.
- [ ] Define/test repeated-option behavior consistently with other global flags. If #891 adds `--org-name`, reject mixed org-ID/org-name selectors even when supplied at different depths.
- [ ] Preserve structured failure metadata and argument-name-only telemetry.
- [ ] Follow #887's display-order policy and the repository's required CLI checks, including telemetry-disabled builds/lints and classifier updates if files are added.

---

## Validated scope and current-PR plan — 2026-09-11

This section records the QA review and the current decision to finalize existing PRs without opening new PRs. It supersedes the proposed routing/fix suggestions in the original report below; the original observations are retained. Central plan: #757.

### Disposition

Deferred from this existing-PR finalization pass. Keep the issue open; no new PR is requested now. This is not a resolution, a severity downgrade, or a waiver of release acceptance.

### Validation and corrections

Confirmed trailing --org-id is rejected after schema-discover kafka. #782 only changes org get/update selectors and backward-compatible positionals; it does not touch ClickPipe argument inheritance. Same theme is not the same defect. A new focused PR should cover every schema-discover leaf and allowed flag positions. Ordering shared help flags is cosmetic and can share general help cleanup rather than block that functional fix.

---

## Original QA report

Found in the 2026-09-11 QA sweep of the combined PR stack at [`d2c37f81`](https://github.com/ClickHouse/clickhousectl/commit/d2c37f814fff53fe68d636e34103e2d0550bf84d), top PR #823, installed release build (package version still 0.4.2). Reproduced behaviour only; no implementation included.

Severity: low-medium.

Order of the five shared flags across every screen that has them:

| order | screens |
|---|---|
| api-key, **org-id**, api-secret, json, debug | 94 |
| api-key, api-secret, **org-id**, json, debug | 25 |
| api-key, api-secret, json, **org-id**, debug | 13 |
| api-key, api-secret, json, debug, **org-id** (after 20+ command flags) | 29 — all 8 `clickpipe create <source>`, `service create` |

Almost certainly where `#[command(flatten)]` lands in each struct's field order. Pick one canonical position (e.g. shared flags first, fixed order).

Functional variant: `cloud clickpipe schema-discover kafka --help` (and kinesis/object-storage/pubsub) shows no `--org-id`; it exists only on the parent `schema-discover` struct, so it must precede `<SERVICE_ID>`:

```sh
clickhousectl cloud clickpipe schema-discover "$SVC" kafka --org-id X --brokers b --topics t --format JSONEachRow
# error: unexpected argument '--org-id' found     exit 2
clickhousectl cloud clickpipe schema-discover --org-id X "$SVC" kafka …        # reaches the API
```

Every other multi-org command, including the sibling `clickpipe create <source>`, accepts `--org-id` after the subcommand.

## Organization positional trade-off revisited — 2026-09-15

Cloud-wide `--org-id` is agreed. Removing positional org IDs remains the current recommendation, but the owner has no hard preference between removing them and retaining optional positional compatibility on org-level commands.

Both designs are technically reasonable. Clap already supports optional positionals; retaining both is not inherently difficult. Its ongoing cost is an additional representation of the same scope, explicit conflict checks against global `--org-id` (and future `--org-name`) at different command depths, plus compatibility/help/tests. The argument for removal is one organization-scope mechanism throughout the cloud tree. The argument for retention is consistency with positional entity IDs on other resource commands and existing callers.

A resolved organization is needed for org-scoped operations, but callers do not always have to type `--org-id`: single-organization auto-detection remains supported, and auth/org-list operations do not need a selected org. That distinction should not be used as an incorrect justification for removal.

The acceptance checklist currently describes removal. If implementation review chooses optional positional compatibility instead, update that checklist explicitly and reject mixed selector forms rather than silently giving one precedence. No broader syntax change is implied by this trade-off note.

sdairs:
`--org-id` now selects organization scope anywhere after `cloud`, including after nested subcommands and resource IDs. CloudArgs declares it once; handlers use a lazy, cached CloudClient scope instead of maintaining competing per-command fields.

The four legacy positional org forms (`get`, `update`, `usage`, `prometheus`) now fail as usage errors. Single-organization detection, zero/multiple-organization errors, authentication/write checks, and failure metadata are preserved. Auth, help, and org-list commands do not gain organization lookup. README examples use the current organization selector and document repeated-option behavior.

Validation: 1,228 unit tests, 461 Cloud subprocess tests, and 52 telemetry tests passed (three existing ignored); both required clippy configurations, telemetry-disabled build, formatting, and diff checks passed. Root reviewed lazy resolution and dispatch migration; full combined-stack gates will follow integration.

Closes #851.
Part of #757. Extends the earlier limited discovery fix in #875.


iskakaushik (approved):
LGTM

PR #902, opened 2026-09-15
feat(clickpipes): put discovery source before service ID

### LINKED ISSUE #888: Make ClickPipe discovery use source-before-service argument order
## Problem

Two adjacent ClickPipe workflows require opposite positional order:

```text
cloud clickpipe create kafka <SERVICE_ID> ...
cloud clickpipe schema-discover <SERVICE_ID> kafka ...
```

Discovery is the only visible command group with a positional resource identifier before its source subcommand. Swapping either current form fails to parse. This affects Kafka, Kinesis, object storage, and Pub/Sub.

Part of the v0.5.0 consistency work in #757. A breaking change is explicitly accepted for this command; do not retain the old order through guessing or a compatibility parser.

## Target behavior

Finish the command path before resource arguments, matching creation:

```text
cloud clickpipe schema-discover kafka <SERVICE_ID> --brokers ... --topics ... --format ...
cloud clickpipe schema-discover kinesis <SERVICE_ID> ...
cloud clickpipe schema-discover object-storage <SERVICE_ID> ...
cloud clickpipe schema-discover pubsub <SERVICE_ID> ...
```

The old `schema-discover <SERVICE_ID> <source>` order must produce a usage error. `--org-id` is handled by the cloud-wide selector in #851 and must work before or after the source/ID; do not recreate a non-global parent selector here.

## Implementation and acceptance

- [ ] Move the service selector from the discovery parent into each source command's argument wrapper, while preserving shared source-field types and request builders.
- [ ] Keep `create <source> <SERVICE_ID>` unchanged.
- [ ] Cover all four source commands, flags around the positional ID, the old rejected order, and organization flags at multiple command depths.
- [ ] Verify handlers send the same service/organization IDs and source request bodies after the argument-structure change.
- [ ] Update README, parent/leaf help, error hints, and affected test/harness examples; describe the breaking change for 0.5.0.
- [ ] Run required CLI and relevant subprocess/request-shape checks; update classifiers if files are added or renamed.

Source: `crates/clickhousectl/src/cloud/clickpipes.rs`, especially `ClickPipeCommands::SchemaDiscover`, `ClickPipeSchemaDiscoverCommands`, and discovery dispatch.

sdairs:
ClickPipe schema discovery now uses `schema-discover <source> <SERVICE_ID>` for Kafka, Kinesis, object storage, and Pub/Sub, matching the creation workflow. Each source wrapper owns its service ID and reuses the existing source fields and request builders.

The former service-before-source order is a usage error. Global organization selection works before or after the source and ID. README/help and affected examples use the source-first syntax; destination-running guidance stays on the object-storage leaf.

Validation: 177 ClickPipe unit tests, 67 CLI tests, and eight discovery request tests passed. Tests assert exact organization/service paths, unchanged source bodies, multiple selector positions, and rejection before HTTP. Formatting and diff checks pass; combined-stack gates follow integration.

Closes #888.
Part of #757.


iskakaushik (approved):
LGTM

Issue #979, opened 2026-09-18
Expose remaining CLI features from merged OpenAPI drift remediations


sdairs:
## Implementation and Development relationship — 2026-09-21

**Fixing in v0.5.0. Development is linked to PR #989**, the final PR of this four-part sequence in native stack #982:

| Feature | Implementation |
| --- | --- |
| Snapshot list/get and configuration get/update | #986 |
| Kinesis Protobuf on create and discovery | #987 |
| MongoDB create-time initial-load parallelism | #988 |
| Start supported ClickPipes paused | #989 — closes this issue after its three prerequisites merge |

Only #989 is a closing Development relationship. #986–#988 are contributing PRs and remain explicitly referenced here; none should independently close this umbrella before all four features land. Preserve the stack order.

All four groups are implemented, current-head checks pass, and README/help plus parsing, classification, builder and subprocess wire tests are present. A fresh source/test audit found no additional implementation gap. No additional PR is needed.

- [x] All four focused implementation PRs exist and are validated as recorded on their current heads.
- [x] The native Development field links the final closing PR, #989.
- [ ] Merge #986 → #987 → #988 → #989; #989 closes this implementation issue.

Final combined release-candidate validation is owned by #757 and remains required after merge. Existing fixture limits remain: live Kinesis/MongoDB ingestion and successful nonempty snapshot-detail retrieval were not established by the recorded fixtures. #990 separately owns pagination/BYOC CLI exposure.

<details>
<summary>Earlier scope and evidence (preserved)</summary>

## Audit result — 2026-09-18

The historical OpenAPI remediations are substantially exposed in the CLI. Four feature groups remain: service snapshots/configuration, Kinesis Protobuf input, MongoDB initial-load parallelism, and creating ClickPipes paused. This issue owns the remaining CLI exposure; it does not reopen completed library drift fixes.

[...]

Issue #939, opened 2026-09-15
Track remaining issue work, handoffs, and deferrals


sdairs:
## Remaining-issue action plan — 2026-09-18

**This is the current plan and supersedes all execution instructions in the historical record below.** The 58-PR stack is merged. Al has now authorized implementation of the five issues in the first work table. Ten reviewed draft PRs are published in native GitHub stack **#982**, through #992. All seven issue assignments are implemented and the combined local gates pass. Final CI and post-merge acceptance remain explicit below. Each issue has a fresh, finite subagent assignment and orchestrator review. The ordinary corrected drift workflow generated API-library issue #985; intentional CLI exposure remains separate in #990. The separately authorized Homebrew PR refresh is complete and remains on hold.

The original planning inventory of **31 open issues** has one primary home per issue here (the execution follow-ups #985 and #990 are recorded below): **5** for Al and an agent, **4** for new engineers, **8** waiting on upstream, **1** Homebrew signing hold, **8** parked, **3** outside this plan, and **2** trackers. #935 is both a newcomer task and blocked; it is counted once under the newcomer handoff.

### What Al needs to do

1. **Review the implementation stack once validation finishes.** Work is underway under the authorization in this session. Live Cloud integration is queued for the affected exact PR heads; merge and release remain separate decisions. After merging the detector and remediation, rerun the normal drift workflow on corrected `main`.
2. **Assign the newcomer work later:** #845, #926, #928 and #935 all have `good first issue`. They are available for the new CLI engineers; none has been assigned by this planning pass.
3. **Coordinate upstream feedback:** share #923's stopped-service context suggestion with the API team; add upstream owners/ticket links for the already-raised backend group when available. The current agent should wait for those contracts/fixes, not invent substitutes.
4. **Revisit release scope in #757 before selecting the release candidate.** In particular, explicitly decide the release treatment of deferred #234/#250/#254. Existing milestones are preserved; #928 remains in v0.5.0. Handoff or deferral here does not silently remove a release requirement.

The macOS signing key is a separate future dependency for Homebrew, not a prerequisite for progressing the five ready issues.

### Authorized implementation queue

| Issue | Agreed work for the agent | Dependency / completion boundary |
| --- | --- | --- |
[...]

PR #986, opened 2026-09-18
Expose service snapshots and snapshot configuration

sdairs:
Add `cloud service snapshot list/get/config get/update` using the existing snapshot APIs, separately from backups. Service selectors, JSON/detail output and read/write authentication follow existing conventions. Configuration updates preserve omitted values, use minutes and require cadence fields together.

Part of #979 and #939 (1 of 4 feature PRs).

Validation: parsing/read-write classification, minimal/maximal request builders, empty/one-sided update rejection, exact endpoint and PATCH wire tests, and human/JSON output tests pass. Full combined CLI formatting, default Clippy/tests, no-default check and no-default all-target Clippy passed. Current-head CI is green.

Live Cloud validation: snapshot list/config reads passed on a disposable service. Enabling snapshots (30/1440 minutes), changing cadence (60/2880) while preserving enabled state, and disabling while preserving cadence all matched readback. The service was deleted and confirmed absent. Its snapshot list was empty, so successful snapshot-detail retrieval remains covered by mocks rather than a live result.


mdevils (approved):

iskakaushik (approved):
LGTM

PR #987, opened 2026-09-18
Support Kinesis Protobuf schema input

sdairs:
Kinesis create and schema discovery now accept Protobuf with `--protobuf-schema-file PATH|-`. The shared reader base64-encodes raw source or descriptor bytes and rejects missing, empty, oversized or format-incompatible input. Existing Kinesis authentication inference is preserved.

Part of #979 and #939 (2 of 4 feature PRs).

Validation: eight Kinesis unit tests and seven wiremock tests pass, including exact file/stdin forwarding on both routes and the authentication regressions. Full combined CLI formatting, default Clippy/tests, no-default check and no-default all-target Clippy passed; current-head CI is green. Live Kinesis ingestion requires an external stream fixture and is not claimed by these tests.


mdevils (approved):

iskakaushik (approved):
LGTM

Issue #990, opened 2026-09-18
Add explicit API-key pagination across the library and CLI


sdairs:
Add explicit API-key pagination across the Rust library and CLI in #991.

The agreed interface is:
- `cloud key list` fetches one page by default.
- `--limit` sets the number of keys per request, from 1–250; omitting it uses the server default of 250.
- `--cursor` resumes from the opaque `nextCursor` returned by a previous page.
- `--all` fetches every page; it permits `--limit` for request size and conflicts with `--cursor`.
- Single-page JSON exposes `result` and available pagination metadata. `--all --json` returns the combined array.
- Name-based get, update, and delete always search all pages and reject cross-page ambiguity.
- All-page operations detect cursor cycles and fail without partial output on later-page errors.

Acceptance includes library request/model coverage, Clap validation, exact subprocess request/output/error coverage, live-test inventory helpers that consume every page, both READMEs, and the repository's required lint/test gates.

The BYOC portion of the earlier mixed scope is now tracked separately in #996 and implemented in #992. The feature split supersedes the earlier always-all default and library/CLI PR split. Generated library drift is recorded in #985. Both feature PRs must merge before the normal corrected-main drift rerun required by #757.

Keep open until #991 merges. Current validation evidence is maintained in that PR.

<details>
<summary>Earlier mixed scope and evidence</summary>

## Implementation status — 2026-09-21

Implemented in #992, the tip of native stack #982 at `18291488`. The current-head checks now pass; the earlier pending-CI note is superseded. Source/test review confirms complete API-key cursor iteration for listing/name lookup and optional-region BYOC profile selection. README and subprocess coverage are present.

Keep open until #992 merges. Multi-page key responses and BYOC-only profiles remain mock-covered where live fixtures were unavailable; the recorded read-only smoke checks cover ordinary key listing/name lookup and region-only profiles. Central plan: #939.

<details>
<summary>Earlier scope and evidence (preserved)</summary>

Follow-up required by #978 and release tracker #939, after generated library remediation #985. The detector workflow has now exposed the contracts; keep this CLI behavior separate from the library PR.

Implement two intentional CLI changes:

- `cloud key list` and API-key name lookup must consume every page by forwarding the opaque `nextCursor` until it is absent/null. Preserve existing JSON array and human output, error/auth handling and lookup ambiguity behavior. Detect repeated cursors to avoid an unbounded request loop. Do not expose a silently truncated first page.
[...]

Issue #978, opened 2026-09-18
Detect operation-parameter and pagination-envelope OpenAPI drift


sdairs:
## Implementation and closure ownership — 2026-09-21

**Fixing in v0.5.0. Development is linked to PR #984**, which implements operation-parameter and successful-response-envelope drift detection. Current-head checks pass, including the [live Cloud run](https://github.com/ClickHouse/clickhousectl/actions/runs/35347942152).

The issue's closure boundary is now the detector implementation merging through #984. **The required normal drift-workflow rerun on corrected `main` remains mandatory and is explicitly owned by the open v0.5.0 release-acceptance tracker #757.** This separates a PR-closeable implementation issue from a check that can only run after merge; no acceptance requirement is waived. #757 must remain open until that run is recorded and the result inspected after the detector and remediation merge.

The corrected-branch generator already produced library issue #985. Library remediation is #991; intentional CLI exposure is separately tracked by #990 / PR #992. #985 was automatically closed on 2026-09-19 by a later report, which does not establish that #991 merged or that corrected-main acceptance passed. The remediated stack dry run found zero actionable findings.

No additional detector code gap was found. The historical completion language below is superseded only in ownership of the post-merge acceptance gate, which remains in #757.

<details>
<summary>Earlier scope and evidence (preserved)</summary>

## Problem

The drift workflow reports no actionable drift for API-key pagination and service-profile query changes that the library does not support. This prevents the normal automated-issue remediation workflow from starting.

Observed on `main` at `a65a9ca6fa49eeac328b20634230f3f11399f5f1`, 2026-09-18. This is a checker bug, not a generated `openapi-drift` report. No remediation was implemented during this audit.

## Workflow evidence

- [Scheduled run, 08:12 UTC](https://github.com/ClickHouse/clickhousectl/actions/runs/35323236572): success, `Actionable drift: 0`, no issue opened.
- [Manual run, 11:00 UTC](https://github.com/ClickHouse/clickhousectl/actions/runs/35337480920): success, `Actionable drift: 0`, no issue opened.
- The canonical Rust analyzer was also run against the locally captured live document containing the changes below and the same main revision. It returned zero findings and three acknowledged legacy enum constraints. This independently establishes a detection gap; the Actions runs do not retain their fetched documents, so their exact fetched bytes cannot be reconstructed from logs alone.

### API-key pagination

[...]

PR #991, opened 2026-09-18
fix(api): support key pagination and optional-region BYOC lookup

### LINKED ISSUE #985: OpenAPI drift: 9 gaps between live spec and library
The live ClickHouse Cloud OpenAPI spec has drifted from the Rust API library.
The comparison was produced by the shared `syn`-based analyzer.

- **Live spec:** `https://api.clickhouse.cloud/v1`
- **Client:** `crates/clickhouse-cloud-api/src/client.rs`
- **Models:** `crates/clickhouse-cloud-api/src/models.rs`
- **Analyzer:** `crates/clickhouse-openapi-analyzer`

## Summary

| Change | Count |
|--------|-------|
| Missing client methods | 0 |
| Extra client methods | 0 |
| Missing operation parameters | 2 |
| Operation parameter mismatches | 1 |
| Unsupported success response schemas | 0 |
| Missing model types | 0 |
| Missing schema definitions | 0 |
| Missing struct fields | 3 |
| Extra struct fields | 0 |
| Missing enum values | 0 |
| Extra enum values | 0 |
| Enum VALUES const mismatches | 0 |
| Additional properties mismatches | 0 |
| Field optionality mismatches | 0 |
| Beta status changes | 0 |
| Deprecated-field changes | 0 |
| Stale snapshot changes | 3 |
| Stale exemptions | 0 |
| New unsupported enum constraints | 0 |
| Changed acknowledged enum constraints | 0 |
| Acknowledged unsupported enum constraints | 3 |

## Missing Operation Parameters

- `/paths/~1v1~1organizations~1{organizationId}~1keys/get/parameters/2` — openapi_key_get_list has no argument for query parameter cursor
- `/paths/~1v1~1organizations~1{organizationId}~1keys/get/parameters/1` — openapi_key_get_list has no argument for query parameter limit

## Operation Parameter Mismatches

- `/paths/~1v1~1organizations~1{organizationId}~1serviceProfiles/get/parameters/1` — service_profiles_list query parameter region_id has incompatible requiredness; Rust argument &str

## New Operation Parameters Missing From Snapshot

- `/paths/~1v1~1organizations~1{organizationId}~1keys/get/parameters/2` — openapi_key_get_list query parameter cursor differs from the snapshot
  - Live contract: `{"in":"query","name":"cursor","required":false,"schema":{"type":"string"}}`
- `/paths/~1v1~1organizations~1{organizationId}~1keys/get/parameters/1` — openapi_key_get_list query parameter limit differs from the snapshot
  - Live contract: `{"in":"query","name":"limit","required":false,"schema":{"default":250,"maximum":250,"minimum":1,"type":"integer"}}`

## Changed Operation Parameter Contracts

- `/paths/~1v1~1organizations~1{organizationId}~1serviceProfiles/get/parameters/1` — service_profiles_list query parameter region_id differs from the snapshot
  - Snapshot contract: `{"in":"query","name":"region_id","required":true,"schema":{"type":"string"}}`
  - Live contract: `{"in":"query","name":"region_id","required":false,"schema":{"type":"string"}}`
  - Snapshot location: `/paths/~1v1~1organizations~1{organizationId}~1serviceProfiles/get/parameters/1`

## Missing Struct Fields

- `/paths/~1v1~1organizations~1{organizationId}~1keys/get/responses/200/content/application~1json/schema/properties/limit` — ApiResponse.limit is missing from the openapi_key_get_list success envelope
- `/paths/~1v1~1organizations~1{organizationId}~1keys/get/responses/200/content/application~1json/schema/properties/nextCursor` — ApiResponse.nextCursor is missing from the openapi_key_get_list success envelope
- `/paths/~1v1~1organizations~1{organizationId}~1keys/get/responses/200/content/application~1json/schema/properties/totalCount` — ApiResponse.totalCount is missing from the openapi_key_get_list success envelope

## Acknowledged Unsupported Enum Constraints

These locations are inventoried but cannot yet be compared to a typed Rust value enum.
Acknowledgments cover the snapshot's enum values; changed values and new or stale locations count as drift.

- `/components/schemas/ApiKey/properties/roles/items` (`models.rs::ApiKey::roles`) — Rust type String is not an enum
- `/components/schemas/ApiKeyPatchRequest/properties/roles/items` (`models.rs::ApiKeyPatchRequest::roles`) — Rust type String is not an enum
- `/components/schemas/ApiKeyPostRequest/properties/roles/items` (`models.rs::ApiKeyPostRequest::roles`) — Rust type String is not an enum

## Implementation Guide

1. Replace `crates/clickhouse-cloud-api/clickhouse_cloud_openapi.json` with this same live document; do not hand-edit it.
2. Follow each finding's `spec_pointer` and `rust_item` to update `client.rs`, `models.rs`, or `meta.rs`.
3. Regenerate beta/deprecation metadata when applicable and add focused model/client tests.
4. Edit `crates/clickhouse-openapi-analyzer/src/config.rs` only for a deliberate, documented divergence. New unsupported acknowledgements require a tracking issue.
5. Run the analyzer and Cloud API tests, Clippy, Python renderer tests, and this dry run again; see `AGENTS.md` for the exact commands.

sdairs:
The corrected OpenAPI drift workflow reported missing API-key pagination and an overly strict region requirement for BYOC service-profile discovery. Add optional limit/cursor query arguments and response pagination metadata, and make the profile region query optional. Refresh the complete captured OpenAPI snapshot from the same live contract.

The CLI changes here only adapt existing calls to the new library signatures; intentional CLI pagination and BYOC-only lookup are tracked separately in #990. Historical analyzer fixtures now embed the pre-remediation Rust surface so the original nine findings and six snapshot-independent findings remain stable regression evidence.

Fixes #985. Follow-up to #978; part of release tracker #939 and native GitHub stack #982.

Validation:
- API/analyzer: 696 tests passed, 26 live/provisioning tests ignored; all-target Clippy clean.
- CLI: 2,002 tests passed, 5 ignored; default Clippy, no-default check and no-default all-target Clippy clean.
- Python: 95 tests passed; formatting clean.
- Corrected analyzer against the live OpenAPI: zero actionable findings. The normal generator must be rerun on corrected main after merge; this branch check does not satisfy that later boundary.
- One existing local Postgres mock-socket timeout passed both an isolated retry and the complete CLI rerun.

Current-head formatting, CLI CI and API-library CI are green. Cloud integration for this exact PR head is requested separately and remains queued until the shared environment is available.


crates/clickhouse-cloud-api/src/client/services.rs
```diff
@@ -4,15 +4,22 @@ use crate::models::*;
...
         byoc_id: Option<&str>,
     ) -> Result<ApiResponse<Vec<ServiceProfile>>, Error> {
         let path = format!("/v1/organizations/{organization_id}/serviceProfiles");
         let req = self.request(reqwest::Method::GET, &path);
-        let req = req.query(&[("region_id", region_id)]);
+        let req = if let Some(region_id) = region_id {
+            req.query(&[("region_id", region_id)])
+        } else {
+            req
+        };
```
balanza: **issue:** ‏these changes are unrelated

crates/clickhousectl/src/cloud/services.rs
```diff
@@ -4296,7 +4296,7 @@ impl CloudClient {
     ) -> crate::cloud::client::Result<Vec<ServiceProfile>> {
         let response = self
             .api()
-            .service_profiles_list(org_id, region_id, byoc_id)
+            .service_profiles_list(org_id, Some(region_id), byoc_id)
```
balanza: **issue:** ‏These changes are unrelated

balanza (changes requested):
I see no need to introduce these model variations (limit/cursor) if, in the end, the user will not have the possibility to walk through the next pages (if any).
Nevertheless, the underlying API already handles defaults correctly, thus no edits are required if we are going to render only the first page of the result set.

I'd defer the implementation to when we will also introduce appropriate command line UX and interface.

(PS: please review the changes, I think some changes should not be here)

sdairs:
Thanks @balanza. I agree with the concern in [your review](https://github.com/ClickHouse/clickhousectl/pull/991#pullrequestreview-5264722025): shipping an all-pages default and changing it later could break users' scripts.

We've split the work by feature:
- #991 now contains pagination across both the Rust library and CLI.
- #992 contains BYOC profile lookup across both the library and CLI.

We've also implemented the full pagination interface now. `cloud key list` returns **one page by default** (250 keys unless `--limit` is supplied). `--limit` accepts 1–250, `--cursor` requests the next page, and **`--all` explicitly fetches every page**. This is the intended default going forward, so we don't need to ship automatic traversal as a temporary default and reverse it later.

Single-page JSON exposes `result` and pagination metadata, including `nextCursor`; `--all --json` returns the combined array. Name-based get/update/delete still search all pages so later matches and duplicate names are handled correctly.

Both PR descriptions and the documentation are updated, and the required local test and lint suites pass.

crates/clickhousectl/src/cloud/api_keys.rs
```diff
@@ -9,16 +9,34 @@ use clap::{Subcommand, builder::PossibleValuesParser};
...
 use tabled::{Table, Tabled, settings::Style};
 
 const API_KEY_STATES: &[&str] = &["enabled", "disabled"];
 
 #[derive(Subcommand)]
 pub enum KeyCommands {
     /// List API keys
-    List,
+    #[command(after_help = "\
+CONTEXT FOR AGENTS:
```
balanza, line 23: **suggestion:** ‏The context is also for humans - agents read the helper anyway


```suggestion
    #[command(after_help = "\=
```

balanza (approved):
LGTM

Let's be aware that this is changing the shape of the output.

crates/clickhousectl/src/cloud/api_keys.rs
mdevils: I agree. I think we shouldn't explicitly target common information to agents.

crates/clickhousectl/src/cloud/api_keys.rs
```diff
@@ -355,45 +375,101 @@ fn build_api_key_update_request(options: &KeyUpdateOptions) -> CloudResult<ApiKe
...
+        print_key_table(page.result, all);
+        if let Some(cursor) = page.next_cursor {
+            if cursor.chars().any(char::is_control) {
+                println!(
+                    "More API keys are available. Use --json to read nextCursor and pass it with --cursor, or use --all."
+                );
+            } else {
+                // Single quotes preserve shell metacharacters, including embedded quotes.
+                // Equals signs keep tokens beginning with '-' attached to their flags.
+                let cursor = cursor.replace('\'', "'\"'\"'");
```
mdevils: I think it makes sense to extract escaping to a function. Or better to use some ready-made escaping. Maybe ome of the already used libraries provide it.

mdevils (changes requested):

crates/clickhousectl/src/cloud/api_keys.rs
sdairs: This is present in every help page and is part of the CLI design for agents; I'm not opposed to going around and removing `CONTEXT FOR AGENTS` - I did it early on, and its probably redundant - but we'd need to go remove it all help pages, not just here.

crates/clickhousectl/src/cloud/api_keys.rs
sdairs: Replaced the custom escaping with shlex::try_quote in a helper

mdevils (approved):

iskakaushik (approved):
LGTM

